PR #20, opened 2023-08-04
Remove deprecated functions, structs and fields from the client and API package

stgraber:
Closes #12

PR #21, opened 2023-08-04
Remove /1.0/containers and /1.0/virtual-machines aliases

### LINKED ISSUE #13: Remove backward compatibility containers/virtual-machines API
As this is effectively a new project, the `/1.0/containers` and `/1.0/virtual-machines` API routes can go away as they were only there for backward compatibility reason with clients that don't use `/1.0/instances`/

stgraber:
With Incus we get a nice clean slate, so remove the compatibility APIs.

Closes #13


stgraber:
Blocked on #20 as lxc-to-incus needs the update to the new client functions.

PR #25, opened 2023-08-04
Remove shiftfs

### LINKED ISSUE #6: Remove shiftfs support
Shiftfs is an Ubuntu-specific technology which is very actively being deprecated in favor of the mainline VFS idmap feature.
There's not much reason to keep the double logic in a new fork, so let's remove it.

stgraber:
Closes #6


brauner:
@sforshee and I are going to savor this moment.

PR #26, opened 2023-08-05
Added publishing to Github pages to for better docs experience

MrDaGree:
I like being able to click links and Im sure others are too. I figured a good starting point was to ensure Pages is enabled and then do something like this.

A few things I am not sure of and would appreciate some feedback on:

1. How much do we want to replace links/references to LXD *now*? Shall we wait until the first release of Incus?
1. For things like forums/discussion topics, how we do want to link stuff like that in the docs?

Example of the [action](https://github.com/MrDaGree/incus/actions/runs/5768247372) in my fork as well as a [preview](https://mrdagree.github.io/incus/)

Checks for Documentation will fail due to [pages not being created yet](https://github.com/cyphar/incus/actions/runs/5768416085/job/15639375367?pr=26#step:7:393)

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -1,0 +1,66 @@
+name: Pages
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - docs/**
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  documentation:
+    name: Documentation Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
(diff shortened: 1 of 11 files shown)


stgraber:
We've been having some discussions with @cyphar about how to handle that in the future, we're hoping to be able to provide some more clarity and direction around that early next week.

stgraber:
Right, so what I was referencing above was the whole move of Incus over to the LXC project.
With that done, documentation will be published at https://linuxcontainers.org/incus/doc/main/ similarly to what was done in the LXD times. Same goes for the forum, we'll use https://discuss.linuxcontainers.org.

Closing the PR as we won't be using Github Pages for this, but simply rely on our existing logic to publish the GH actions artifact on the website.

(I expect most of that tooling to be put back in place over the next couple of weeks)

PR #36, opened 2023-08-10
Rename `inc` to `incus`

stgraber:
Follow-up from my previous PR, this time renaming the client tool.


stgraber:
Tests are still running on my system so this may well be broken but figured I'd push it up before going to bed so if it's all good we end up with a mergable branch.

PR #38, opened 2023-08-10
Switch from Dqlite to CowSQL

freeekanayaka:
This changes the code to point to [cowsql](https://github.com/cowsql/cowsql) instead of dqlite.

It's the minimal necessary change to make things work. Follow-up work might include:

- Updating references of dqlite in the documentation under `doc/`
- Updating variable names across the code base


stgraber (approved):
Nice!

grawlinson:
Hi! I co-maintain the `lxd` package for Arch Linux, and I'm understandably quite curious how much of a hard fork incus will be.

For example; I see that `cowsql/raft` has also been created, are there also plans to switch from `canonical/raft` to this fork?

I'm following the changes quite closely, from both a distribution packager's and end-user's point of view.

freeekanayaka:
Hi!

> Hi! I co-maintain the `lxd` package for Arch Linux, and I'm understandably quite curious how much of a hard fork incus will be.

I'll let @stgraber speak on that.

> 
> For example; I see that `cowsql/raft` has also been created, are there also plans to switch from `canonical/raft` to this fork?

The switch already happened, cowsql/cowsql uses cowsql/raft. And users of incus should build/distribute incus against cowsql/* libraries, since that's what incus upstream depends on and test against.

> I'm following the changes quite closely, from both a distribution packager's and end-user's point of view.

Nice, please keep letting us know of any doubts or concerns.

freeekanayaka:
> Hi! I co-maintain the `lxd` package for Arch Linux, and I'm understandably quite curious how much of a hard fork incus will be.

Note that as far as github.com/cowsql/* is concerned, the fork is currently light but it will likely slowly divert significantly.

stgraber:
> Hi! I co-maintain the `lxd` package for Arch Linux, and I'm understandably quite curious how much of a hard fork incus will be.

Hey there, the goal on the Incus side is to support seamless data migration from LXD 5.0.x and 5.16 or earlier.
It's quite possible that we'll be able to extend that to some later LXD versions so long as the DB schema doesn't diverge.

Other than that, we will be keeping track of changes on the LXD side and import those that make sense into Incus, but not all of them will and Incus is likely to grow new features which will not be included in LXD.

So far as there is no CLA enforcement on LXD, both projects are free to import from each other as they see fit.
Should Canonical introduce a CLA on LXD, then the contributions made to Incus will not be suitable for inclusion into LXD while the other way around (taking LXD changes into Incus) will remain perfectly fine.

PR #39, opened 2023-08-11
shared/termios: Use golang.org/x/term instead of golang.org/x/crypto/ssh/terminal

ucpr:
## Summary

Since `golang.org/x/crypto/ssh/terminal` is deprecated, modify it to use `golang.org/x/term`.

## Links

- https://pkg.go.dev/golang.org/x/crypto/ssh/terminal
- https://pkg.go.dev/golang.org/x/term


stgraber (approved):
Looks good, thanks!

PR #46, opened 2023-08-17
Make `snapshot` a proper subcommand

### LINKED ISSUE #9: Introduce a proper `inc snapshot` subcommand
One bit of inconsistency that's always bothered me with LXD is that `lxc snapshot` and `lxc restore` are top level commands of their own. This makes it very awkward to allow for snapshot listing, renaming, ...

We've been wanting to fix this for years, but couldn't without breaking backward compatibility. As the fork doesn't pretend to be backward compatible at the CLI level, it's a good opportunity to fix that one.

monstermunchkin:
Closes #9


stgraber (approved):

stgraber:
Thanks!

PR #69, opened 2023-09-07
cmd/lxd-to-incus: Uncompress snapshot files in target database

### LINKED ISSUE #66: Disable raft snapshot compression
I've run a few tests and benchmarks around cowsql database raft snapshot compression using lz4, and the conclusion seems to be that they are not worth, since they save relatively little space at the price of noticeable CPU overhead.

The database of an incus deployment with 1000 instances and associated metadata (profiles, networks, snapshots, images, volumes, etc) has an uncompressed size of ~4M and a compressed size of ~3M, so lz4 doesn't really move the needle there. A deployment of 50k instances has a database size of ~181M, and a compressed size of ~150M: the storage/memory gain is modest, while the CPU overhead of running compression can be noticeable when hitting the API when the snapshot is in progress.

A much more effective measure seems to be to regularly `VACUUM` the database, which can reduce the size of a database dramatically (even 10x for databases that have been running for a long time and have a lot of fragmentation). I'm planning to run VACUUM transparently in cowsql.

If there's consensus around this, I'd like to remove compression from both incus and cowsql and figure the best option for upgrades (it'd basically just a matter of uncompressing the existing snapshots at upgrade time and deciding where to put that logic).

This should also simplify the on-disk format and make it easier to improve the command line tooling that we have around it.

freeekanayaka:
Fixes #66

### DIFF
--- a/cmd/lxd-to-incus/db.go
+++ b/cmd/lxd-to-incus/db.go
@@ -1,0 +1,105 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/pierrec/lz4/v4"
+
+	"github.com/lxc/incus/shared"
+)
+
+// Uncompress the raft snapshot files in the given database directory.
+//
+// A backup will be created and kept around in case of errors.
+func migrateDatabase(dir string) error {
+	global := filepath.Join(dir, "global")
+
+	err := shared.DirCopy(global, global+".bak")
+	if err != nil {
+		return fmt.Errorf("Failed to backup database directory %q: %w", global, err)
+	}
+
+	files, err := ioutil.ReadDir(global)
+	if err != nil {
+		return fmt.Errorf("Failed to list database directory %q: %w", global, err)
+	}
+
+	for _, file := range files {
+		var timestamp uint64
+		var first uint64
+		var last uint64
+
+		if !file.Mode().IsRegular() {
+			continue
+		}
+
+		n, err := fmt.Sscanf(file.Name(), "snapshot-%d-%d-%d\n", &timestamp, &first, &last)
+		if err != nil || n != 3 {
+			continue
+		}
+
+		filename := filepath.Join(global, file.Name())
+		err = lz4Uncompress(filename)
+		if err != nil {
+			return fmt.Errorf("Failed to uncompress snapshot %q: %w", filename, err)
+		}
+	}
+
+	return nil
+}
+
+// Uncompress the given file, preserving its mode and ownership.
+func lz4Uncompress(zfilename string) error {
+	zr := lz4.NewReader(nil)
+
+	zfile, err := os.Open(zfilename)
+	if err != nil {
+		return fmt.Errorf("Failed to open file %q: %w", zfilename, err)
+	}
+
+	zinfo, err := zfile.Stat()
+	if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/lxd-to-incus/db.go
```diff
@@ -0,0 +1,102 @@
+package main
+
+import (
+	"fmt"
+	"github.com/lxc/incus/shared"
+	"github.com/pierrec/lz4/v4"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
```
stgraber, line 10: We usually group those like:
```
"fmt"
"io"
"io/ioutil"
"os"
"path/filepath"

"github.com/pierrec/lz4/v4"

"github.com/lxc/incus/shared"
```

Basically: stdlib, external, local

cmd/lxd-to-incus/db.go
```diff
@@ -0,0 +1,102 @@
...
+		if err != nil {
+			return fmt.Errorf("Failed to uncompress snapshot %q: %w", filename, err)
+		}
+	}
+
+	return nil
+}
+
+// Uncompress the given file, preserving its mode and ownership.
+func uncompress(zfilename string) error {
```
stgraber, line 54: Maybe `lz4Uncompress` so it's clear that it does just lz4?

cmd/lxd-to-incus/db.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return fmt.Errorf("Failed to open file %q: %w", zfilename, err)
+	}
+
+	zinfo, err := zfile.Stat()
+	if err != nil {
+		return fmt.Errorf("Fialed to get file info for %q: %w", zfilename, err)
+	}
+
+	mode := zinfo.Mode() // use the same mode for the output file
```
stgraber, line 67: Tiny nit but we usually prefer the comments to be on their own line :)

cmd/lxd-to-incus/main.go
```diff
@@ -395,6 +395,11 @@ Instances will come back online once the migration is complete.
 		return fmt.Errorf("Failed to move %q to %q: %w", sourcePaths.Daemon, targetPaths.Daemon, err)
 	}
 
+	err = migrateDatabase(filepath.Join(targetPaths.Daemon, "database"))
```
stgraber, line 398: I'd add another status update here for the user.
```
fmt.Println("=> Migrating database")
```

stgraber (changes requested):

stgraber:
@freeekanayaka so this will uncompress the database on migration, but I guess we need another change to have Incus not set it up to be compressed in the first place, right?

Otherwise as soon as we start Incus, it will just go back to being compressed again, no?

freeekanayaka:
> @freeekanayaka so this will uncompress the database on migration, but I guess we need another change to have Incus not set it up to be compressed in the first place, right?
> 
> Otherwise as soon as we start Incus, it will just go back to being compressed again, no?

Not really, as soon as this lands I'd disable/remove compressing entirely from cowsql.

cmd/lxd-to-incus/db.go
freeekanayaka: Fixed.

cmd/lxd-to-incus/db.go
freeekanayaka: Fixed.

stgraber (approved):

Issue #10, opened 2023-08-03
Move the daemon commands under `incus admin`


stgraber:
Rather than have users invoke sub-commands on the daemon, lets move all of those over to a new `incus admin` sub-command in the client. All daemon commands have been API based for years, so there's no real reason for them to be under a different binary and having to keep both the daemon and client in the PATH has caused no end of confusion over the years.

With this change in place, the `incusd` binary should be kept out of PATH, only leaving `incus` to be user visible.

cyphar:
This really would make things a lot nicer. Nee-`lxd init` was always a weird command to deal with. Is there a nice way to move the daemon configuration stuff from `inc config` to `inc admin` as well? I always felt it was strange that daemon and workload configuration was done through the same subcommand. Or would the plan just be to move the daemon subcommands?

stgraber:
Hmm, I didn't think of trying to undo the `inc config` for both instance and server settings but that's something that may be worth thinking about.

My current plan was to make `inc admin` restricted to local use only, so no support for remotes or anything at that level, effectively matching what we had with the `lxd XYZ` commands as those may rely on internal unix-only APIs.

Maybe we should have a separate `inc server` sub-command with the usual set of `set/unset/show/edit/info`.
Essentially moving the server config logic over to that and keeping the top-level commands focused on instances.

sharathsivakumar:
@stgraber I would like to work on this. Would it be possible to assign this to me?

stgraber:
For that one, I think I'd recommend:
 - Send an initial PR moving `incusd init` to `incus admin init`
 - Then do `incusd cluster` to `incus admin cluster`
 - Then remove `import` from `incusd` and move `recover` to `incus admin recover`
 - Finally move `activateifneeded`, `shutdown` and `waitready`

That should leave us with nothing other than the boilerplate `help` and `version` commands as well as the hidden internal subcommands (mostly `forkXYZ`).

sharathsivakumar:
Thanks @stgraber. Those recommendations help a lot.

stgraber:
@sharathsivakumar how's that going?

sharathsivakumar:
@stgraber Sorry was sick for the last few weeks. I am back to work this week. Looking to get the first PR out before the end of this week. Hope that's fine.

stgraber:
Sounds good!

sharathsivakumar:
- [ ] `incusd init` to `incus admin init` 
[...]

PR #79, opened 2023-09-12
Move `incusd init` to `incus admin init`

sharathsivakumar:
This change creates a new sub-command `admin` under `incus`. Further it moves `incusd init` to `incus admin init`

Check https://github.com/lxc/incus/issues/10 for more details

@stgraber Please review and let me know if fixes are needed.

TODO: Fix tests for change in command


sharathsivakumar:
@stgraber I have left the init files under incusd as it is. Will remove them once all the tests pass for incus admin init

stgraber:
@sharathsivakumar looking at the test results, you need a quick rebase at least and update after that as some packages have moved from shared/ to internal/ (cmd and i18n at least)

sharathsivakumar:
Noted. The tests longer to complete and I forgot about it. I will rebase and check the issues.

stgraber:
Still failing. Let me do the rebase for you see if that helps.

stgraber:
So I ended up having to do:
 - Rebase on the current main branch
 - Update admin.go for new package paths
 - Applied all recent changes from incusd/main_init* to matching cmd/incus/admin_init*

It now builds fine, so lets see what the tests say.

stgraber:
Some package rework is going to be needed to sort out the client builds (CGO_ENABLED=0) but those may be potentially tricky so I'll take care of those once the rest is happy.

sharathsivakumar:
Thanks a ton. I will proceed with the remaining cmd changes.

stgraber:
Hey there,

So I was thinking about the remaining commands. I don't think we'll actually move them all.
`activateifneeded` requires a lot of internal access and so should stay where it is.

As a summary, I think we'll want the following:
 - incus admin init
 - incus admin cluster
 - incus admin recover
 - incus admin sql
 - incus admin shutdown
 - incus admin waitready

sharathsivakumar:
Cool. Let me start working on these and create PRs. I will create a PR for each one of the sub commands.

sharathsivakumar:
@stgraber Shall I go ahead and delete the init related files under incusd? Maybe once all the test pass, I can do that before the merge?

stgraber:
Yeah, you can add a commit that does that.

The current failures are mostly unrelated so it should be fine to do that now and then deal with anything else still using the old commands.

sharathsivakumar:
@stgraber I started removing the redundant `incud/init.go` file and noticed that  the incusd api uses a function `initDataNodeApply` which was defined in `incusd/init.go`. So I see 2 options to resolve this:

- Define init.go as a shared package and use it in both the api and init commands. Do we have a location for shared internal packages?
- Leave the `incusd/init.go` as is. 

I think the first option is better in the long run. The second one is more of quick fix and does not help a lot.

What are your thoughts on this? Please feel free to suggest alternatives as well.
-

stgraber:
Hmm, it may be possible to turn that function into something suitable for the `client/` package, I'll take a stab at that later today to see if I can unblock you. I'll also sort out the other package mess currently blocking the rest of the tests.

The old cluster join API isn't supported anymore so I think some of the code in those functions can go away, but not quite all of it sadly, so it's going to need a bunch of poking around across the codebase to sort that one out :)

If you have some time before I sort this one out, you may want to look at moving `recover` which hopefully will be a large-ish command but one that shouldn't be quite so tied to incusd internals.

sharathsivakumar:
Cool. I will move on to recover. Feel free to let me know if I can help with the init restructuring. I will also try going through more of the code base and come up with some ideas to untangle it.

stgraber:
I've sorted out the init.go thing, now I'm looking at the internal packages that are causing build failures on windows and macos.

stgraber:
The big package cleanup has now been merged in main, I'm going to rebase this branch now.
I'll also add a commit which updates `test/*` to use `incus admin init` instead of `incusd init`

stgraber:
And pushed, let's see how the tests do :)

stgraber:
Ah, still need to get rid of an idmap import :)

stgraber:
@sharathsivakumar all good!

sharathsivakumar:
Wohoo. Thanks a ton! Will get the remaining things done!

PR #97, opened 2023-09-28
[DRAFT]move `incusd sql` to `incus admin sql` 

sharathsivakumar:
This change moves `incusd sql` to `incus admin sql`

TODO: Fix tests


stgraber:
Superseded by #99

PR #107, opened 2023-10-02
Move `incud recover` to `incus admin recover`

sharathsivakumar:
This change moves `incusd recover` to `incus admin recover`. A new package internal/recover is also introduced  in this change. The structures used by the internal recover api and the recover command code are moved into a common package(internal/recover).

### DIFF
--- a/cmd/incus/admin.go
+++ b/cmd/incus/admin.go
@@ -24,6 +24,10 @@ func (c *cmdAdmin) Command() *cobra.Command {
 	adminInitCmd := cmdAdminInit{global: c.global}
 	cmd.AddCommand(adminInitCmd.Command())
 
+	// recover sub-command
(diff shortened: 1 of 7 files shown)


sharathsivakumar:
@stgraber Could you please take a look at the changes for recover?

stgraber:
@sharathsivakumar sorry, I meant to review and merge that one today but ended up being a bit distracted by the LXD 5.17 import.

My plan at this point is to have someone review my 5.17 import branch while I get some sleep and then in the morning here I'll review your branch, solve any conflict or small changes needed and merge it before continuing with the LXD 5.18 import.

sharathsivakumar:
Thanks a ton!

stgraber (approved):
Rebased things a bit to combine some of the commits and to make the move to internal/recover happen prior to moving the command to cmd/incus.

PR #108, opened 2023-10-02
Import LXD 5.17

stgraber:


stgraber:
I used the new importation tool to process all commits up till LXD 5.17.

A few needed some manual tweaking here and there, but overall the automated remap/rewrite patterns handled 80-90% of cases. With the tricky cases mostly being where Incus and LXD have diverged, specifically around server config keys and authentication.

stgraber:
The DCO check should be ignored as it's just failing flat when submitting more than 100 commits or so...

PR #111, opened 2023-10-03
Import LXD 5.18

stgraber:


stgraber:
This is currently incomplete, I still have another 50 or so commits to include (primarily doc changes).

stgraber:
Assuming the tests behave, that's all of LXD 5.18 processed.

I'll likely be doing a follow-up update for the syslog/OVN bit once I get a bit more time to look at it. I don't think that we really want to stream all log entries, it'd probably make more sense to filter the ACL related ones and change the event type to be about network-acls rather than OVN in general.

Full log aggregation on the system is better handled with a separate LOKI integration that streams the entire system log, having Incus do that job feels odd.

stgraber:
@monstermunchkin since you worked on that one on the LXD front, do you see a problem with restricting the log forwarding to ACL log messages? Did you see other log entries that aren't ACL-related and were worth including?

monstermunchkin:
> @monstermunchkin since you worked on that one on the LXD front, do you see a problem with restricting the log forwarding to ACL log messages? Did you see other log entries that aren't ACL-related and were worth including?

It shouldn't be a problem only including ACL logs. For these logs the module name should be be `acl_log` (see `ovn-appctl vlog/list` for a full list of module names) which can be filtered. Regarding other interesting logs, I don't know which others could be interesting.

PR #124, opened 2023-10-07
shared/cliconfig: Fix main socket detection

stgraber:


stgraber:
Self reviewing and merging as we definitely need this fix in 0.1...

PR #159, opened 2023-10-11
tests: Add incus-user test

### LINKED ISSUE #142: Add test for incus-user
We just had a regression in `incus-user` which really should have been caught prior to release.
This highlighted the fact that this daemon is completely untested and entirely relies on manual testing pre-release.

We should add a basic test to the testsuite which spawns incus-user and attempts to interact with Incus from an unprivileged user to validate its behavior.

stgraber:
Closes #142

### DIFF
--- a/test/suites/incus_user.sh
+++ b/test/suites/incus_user.sh
@@ -1,0 +1,20 @@
+test_incus_user() {
+  ensure_import_testimage
+  ensure_has_localhost_remote "${INCUS_ADDR}"
+
+  incus-user --group nogroup &
+  USER_PID="$!"
+  while :; do
+    [ -S "${INCUS_DIR}/unix.socket.user" ] && break
+    sleep 0.5
+  done
+
+  USER_TEMPDIR="${TEST_DIR}/user"
+  mkdir "${USER_TEMPDIR}"
+  chown nobody:nogroup "${USER_TEMPDIR}"
+
+  cmd=$(unset -f incus; command -v incus)
+  sudo -u nobody -Es -- env INCUS_CONF="${USER_TEMPDIR}" "${cmd}" project list
+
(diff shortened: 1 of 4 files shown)


test/suites/incus_user.sh
```diff
@@ -0,0 +1,17 @@
+test_incus_user() {
+  ensure_import_testimage
+  ensure_has_localhost_remote "${INCUS_ADDR}"
+
+  incus-user --group nogroup &
+  USER_PID="$!"
+  sleep 3
```
hallyn, line 7: Seems fragile - given how slow github runners can sometimes be, it would seem worth checking availability over the socket in a loop, if possible.

hallyn (approved):
Approving anyway, as the comment was just a suggestion.  Leaving it to you to merge though, if you want to skip the suggestion.

test/suites/incus_user.sh
```diff
@@ -0,0 +1,19 @@
+test_incus_user() {
+  ensure_import_testimage
+  ensure_has_localhost_remote "${INCUS_ADDR}"
+
+  incus-user --group nogroup &
+  USER_PID="$!"
+  while :; do
+    [ -S "${INCUS_DIR}/unix.socket.user" ] && break
```
hallyn, line 8: (sorry) well, that will spin pretty hard.  Can you add a sleep 0.5s in there or something?  The idea isn't to get past this as fast as possible, but rather to avoid getting to the later commands before it's ready.

stgraber: oops, yeah, I meant for a 0.5 :)

stgraber: Fixed

hallyn (approved):

PR #169, opened 2023-10-14
lxd-to-incus: Add support for manually built LXD

stgraber:


cmd/lxd-to-incus/main.go
```diff
@@ -94,6 +94,7 @@ func (c *cmdMigrate) Run(app *cobra.Command, args []string) error {
 	if source == nil {
 		return fmt.Errorf("No source server could be found")
 	}
+	fmt.Printf("==> Detected: %s", source.Name())
```
monstermunchkin, line 97: Shouldn't there be a `\n` at the end? Otherwise the output would be:
```
==> Detected: snap package => Looking for target server
```

Or is that intentional?

stgraber: Oops, I'll add the missing line break

stgraber: And fixed!

monstermunchkin:
Static analysis is unhappy.

stgraber:
Oops, got that one fixed now!

monstermunchkin (approved):

freeekanayaka (approved):

PR #185, opened 2023-10-21
shared/cliconfig: Nicer error on missing socket

### LINKED ISSUE #184: Improve error message about failed launch
[The internet said](https://discuss.linuxcontainers.org/t/lxd-error-get-http-unix-socket-1-0-dial-unix-var-snap-lxd-common-lxd-unix-socket-connect-connection-refused/12305/2), that this message essentially says that Incus failed to spin up a server: 

`Error: Get "http://unix.socket/1.0": dial unix /var/lib/incus/unix.socket: connect: no such file or directory`

I that is the cause, do you think we could improve this message, and add some links to the documentation?

Something like: 

The Incus server is not running. Did you try the following? Otherwise, take a look at XYZ.

stgraber:
Closes #184

### DIFF
--- a/shared/cliconfig/remote.go
+++ b/shared/cliconfig/remote.go
@@ -4,7 +4,9 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"encoding/pem"
+	"errors"
 	"fmt"
+	"net"
 	"os"
 	"runtime"
 	"strings"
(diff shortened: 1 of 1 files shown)


iBug:
Not the best way to do this. I got confused by the error message when running `incus` as an unprivileged user that doesn't belong to the `incus-admin` group, but I knew the Incus server was running properly.

It would be much better if this had differentiated between `ENOENT`, `EACCES` and `ECONNREFUSED`. In my case it's `EACCES`, which should not be interpreted as "server not running".

stgraber:
https://github.com/lxc/incus/pull/371

PR #201, opened 2023-10-27
lxd-to-incus: query systemd instead of assuming service file path

adamcstephens:
systemd supports multiple paths, so rely on its detection logic instead of checking a specific path.

### DIFF
--- a/cmd/lxd-to-incus/targets.go
+++ b/cmd/lxd-to-incus/targets.go
@@ -25,7 +25,8 @@ func (s *targetSystemd) Present() bool {
 		return false
 	}
 
-	if !util.PathExists("/lib/systemd/system/incus.service") {
+	_, err := subprocess.RunCommand("systemctl", "list-unit-files", "incus.service")
+	if err != nil {
 		return false
 	}
 


stgraber:
Commit isn't signed-off, please make sure the commit message includes the `Signed-off-by: Firstname Lastname <email>` line.

adamcstephens:
Fixed.

stgraber:
```
 Checking for oneline assign & test...
cmd/lxd-to-incus/targets.go:28:	if _, err := subprocess.RunCommand("systemctl", "list-unit-files", "incus.service"); err != nil {
```

Our code style doesn't allow for one-line define+check syntax, so you'll need to update that to be two lines instead.

adamcstephens:
Fixed. I wish github would allow workflows on a PR that has already been approved to run them. :(

stgraber:
Yeah, we have the setting on Github Actions to require an initial approval for a new contributor which is fine, except that Github apparently requires it for every change in that initial PR...

stgraber (approved):

PR #224, opened 2023-11-07
Add "btrfs.nodatacow" config option

tarruda:
This options is added to btrfs storage pools to control the "nodatacow" attribute of the volume/directory containing VM images. The default value is true for compatibility with previous versions.

As explained in [this post](https://discuss.linuxcontainers.org/t/nodatacow-on-btrfs-storage-breaks-compression/18224), the default "nodatacow" option appears to break compression, so this is likely necessary for anyone using `compress-force`  in `btrfs.mount_options` or in the host Btrfs mount options.

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -325,6 +325,13 @@ This introduces the `btrfs.mount_options` property for Btrfs storage pools.
 
 This key controls what mount options will be used for the Btrfs storage pool.
 
+## `storage_btrfs_nodatacow`
+
+This introduces the `btrfs.nodatacow` property for Btrfs storage pools.
+
+This key controls if `chattr +C` will be invoked for storage volumes on Btrfs
+storage pools. The default value is `true`.
+
 ## `entity_description`
 
 This adds descriptions to entities like containers, snapshots, networks, storage pools and volumes.
--- a/doc/reference/storage_btrfs.md
+++ b/doc/reference/storage_btrfs.md
@@ -62,6 +62,7 @@ The following configuration options are available for storage pools that use the
 Key                             | Type      | Default                    | Description
 :--                             | :---      | :------                    | :----------
 `btrfs.mount_options`           | string    | `user_subvol_rm_allowed`   | Mount options for block devices
+`btrfs.nodatacow`               | bool      | `true`                     | Enable `nodatacow` on the storage volume/directory for VMs
 `size`                          | string    | auto (20% of free disk space, >= 5 GiB and <= 30 GiB) | Size of the storage pool when creating loop-based pools (in bytes, suffixes supported, can be increased to grow storage pool)
 `source`                        | string    | -                          | Path to an existing block device, loop file or Btrfs subvolume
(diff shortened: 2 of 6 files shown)


stgraber:
Wouldn't a better solution be to track down the mount options of the underlying btrfs mount and automatically skip the `+C` part if `compress-force` is present.

tarruda:
> Wouldn't a better solution be to track down the mount options of the underlying btrfs mount and automatically skip the +C part if compress-force is present.

This is not only related to compression, it also affects snapshots and filesystem resiliency. From https://btrfs.readthedocs.io/en/latest/Administration.html:

`Nodatacow implies nodatasum, and disables compression` 

and

 `Updates in-place improve performance for workloads that do frequent overwrites, at the cost of potential partial writes, in case the write is interrupted (system crash, device failure).`

IMO `nodatacow` should not be the default (made the default in this PR for compatibility), because it removes reliability features that are often taken for granted  such as data checksumming.

The only advantage of `nodatacow` (improved performance for overwrite workloads) is probably only visible in synthetic benchmarks.

stgraber:
Hmm, that behavior is something that was recommended by btrfs upstream and which is also present in libvirt and other platforms similarly dealing with files that are expected to be modified in place and where copy-on-write causes extremely fast fragmentation, completely destroying performance on spinning rust and also shortening the lifespan on SSDs.
The lack of checksum is the biggest downside but as those files are raw disk images, that layer is effectively delegated to the guest.

Adding a new config option to turn it off is not terribly likely to be used, 99%+ of our users never go and tweak those, that's why whenever possible, we much prefer to do the right thing automatically.

In this case, you mentioned that you have `compress-force` set and that in such a scenario, the space saving and performance/packing benefits of it may outweigh it.

I'm inclined to agree so I'd suggest we just detect the presence of any of the compress related flags on the parent mount and just skip +C in that case.

tarruda:
I'd like to suggest an alternative: It seems the `datacow`/`nodatacow` options can be specified as btrfs mount options. Would it be enough to check if `datacow` is present in `btrfs.mount_options` and disable the invocation of `chattr +C` in this case? This would allow users to signal the explicit intention of using CoW without adding a new configuration.

Detecting compression seems a bit tricky, AFAICT I would have to probe `/proc/mounts` to find the parent mount of the storage volume (is there an easier way)?

stgraber:
> Detecting compression seems a bit tricky, AFAICT I would have to probe `/proc/mounts` to find the parent mount of the storage volume (is there an easier way)?

I had a chat with @brauner, we have a good way with current kernels (statx to get mnt_id and parsing mountinfo for it) and he's got a kernel patchset to make this be even better likely in Linux 6.7.

stgraber:
> I'd like to suggest an alternative: It seems the `datacow`/`nodatacow` options can be specified as btrfs mount options. Would it be enough to check if `datacow` is present in `btrfs.mount_options` and disable the invocation of `chattr +C` in this case? This would allow users to signal the explicit intention of using CoW without adding a new configuration.

Yeah, that should be easy enough to add as an extra way to force the logic for cases where you want it disabled but also don't have compression enabled.

tarruda:
closed in favor of #225

PR #232, opened 2023-11-21
doc: incus -> incusd in build instructions.

fwaggle:
Following the instructions as-written yields:

```
sudo -E PATH=${PATH} LD_LIBRARY_PATH=${LD_LIBRARY_PATH} $(go env GOPATH)/bin/incus --group sudo
Error: unknown flag: --group
```

This is a trivial change, I do not care about copyright, so if there's something wrong with this commit feel free to just commit it elsewhere. 👍


stgraber (approved):

stgraber:
Thanks!

PR #257, opened 2023-12-01
Some OpenFGA fixes

stgraber:


markylaing:
Hi @stgraber :)

What was the scenario in which you saw OpenFGA requests taking longer than 10 seconds? Was this only during an initial resource synchronisation (https://github.com/lxc/incus/blob/f700f9f40466ecd3ebf0cab3bbd9300c1a3afa7e/internal/server/auth/driver_openfga.go#L860) or was it on write/delete of a single tuple?

Additionally, is this occurring with OpenFGA running an in-memory storage driver?

Thanks

stgraber:
It was during the initial sync.

It's a production OpenFGA server with persistent storage and multiple other servers using it.

PR #268, opened 2023-12-04
doc/installing: init NixOS instructions

adamcstephens:
As requested. :)

I'll also have an alternative installation for the Mac client once I get the incus client portion refactored in nixpkgs.


stgraber:
Thanks!

PR #281, opened 2023-12-07
Allow configuring OVN SSL settings through server configuration

stgraber:


internal/linux/memfd.go
```diff
@@ -0,0 +1,39 @@
...
+func CreateMemfd(content []byte) (*os.File, error) {
+	// Create the memfd.
+	fd, err := unix.MemfdCreate("memfd", unix.MFD_CLOEXEC)
+	if err != nil {
+		return nil, err
+	}
+
+	// Set its size.
+	err = unix.Ftruncate(fd, int64(len(content)))
+	if err != nil {
```
hallyn, line 19: Hm, in tihs case and any case where you return before returning os.Newfile(fd), should you be closing fd?

stgraber: Yep, I definitely should :)

stgraber: Revert logic added which will now close the fd if anything goes wrong in the function.

PR #294, opened 2023-12-11
Add variable certificate store columns

### LINKED ISSUE #250: Tweak certificate store columns
The columns in `incus config trust list` should be tweaked a bit:
 - Add support for configurable columns (like we have in `incus list` and `incus image list`)
 - Hide `common name` by default
 - Hide `issue date` by default
 - Move `name` to be the first column
 - Add a column for `restricted` and include it in the default view
 - Add an optional column for the project list

rhburt:
Closes #250 

I've adapted how variable columns are accomplished in `image list` to work for `config trust list`.

Contributing guidelines said to update the translations alongside changes to any CLI strings as well.

### DIFF
--- a/cmd/incus/config_trust.go
+++ b/cmd/incus/config_trust.go
@@ -372,12 +372,25 @@ func (c *cmdConfigTrustEdit) Run(cmd *cobra.Command, args []string) error {
 }
 
 // List.
+const layout = "Jan 2, 2006 at 3:04pm (MST)"
+
 type cmdConfigTrustList struct {
 	global      *cmdGlobal
 	config      *cmdConfig
 	configTrust *cmdConfigTrust
 
-	flagFormat string
+	flagFormat  string
+	flagColumns string
+}
+
+type certificateColumn struct {
+	Name string
+	Data func(rowData rowData) string
+}
+
+type rowData struct {
+	Cert    api.Certificate
+	TlsCert *x509.Certificate
 }
 
 func (c *cmdConfigTrustList) Command() *cobra.Command {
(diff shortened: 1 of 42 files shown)


stgraber:
```
 Error: cmd/incus/config_trust.go:496:2: S1011: should replace loop with `projects = append(projects, rowData.Cert.Projects...)` (gosimple)
	for _, project := range rowData.Cert.Projects {
	^
```

stgraber:
This is also breaking the `TLS restrictions` test, it likely just needs a simple update to select just the column it requires.

stgraber:
Hey there, you're likely going to want to add the description column now that it's been added with https://github.com/lxc/incus/pull/297

rhburt:
Hey @stgraber 

I've fixed both the `gosimple` error and the TLS restrictions test. I'm still getting a failing test from `suites/basic.sh`, but I've run the test over the main branch with the same result so I don't believe my changes are affecting that test. Please let me know if I'm wrong here.

I've also added in the description column as per your last comment. It appears the `config trust add-certificate` command does not support a description field, so I wasn't able to verify the change past ensuring the column exists in the table.

Thanks!

stgraber:
> I've also added in the description column as per your last comment. It appears the `config trust add-certificate` command does not support a description field, so I wasn't able to verify the change past ensuring the column exists in the table.

`incus config trust edit` should let you set the description field.

rhburt:
> > I've also added in the description column as per your last comment. It appears the `config trust add-certificate` command does not support a description field, so I wasn't able to verify the change past ensuring the column exists in the table.
> 
> `incus config trust edit` should let you set the description field.

Great, thanks. I've confirmed the description is displaying as expected. I should have some time tonight to figure out why the `test_remote_admin` test is failing.

The `suites/basic.sh` failure turned out to be a permission error on my system, so no worries there.

stgraber:
@rhburt I took a quick look. I also used the opportunity to extract the test changes to a separate commit.

Basically the issue here was an invocation of `incus config trust list` in the previous test (remote_usage) which was meant to delete all certificates from the trust store, but didn't anymore due to the syntax change. This was then causing the next test (remote_admin) to run with a dirty state and break.

That kind of breakage can be a bit weird to debug as running the test in isolation wouldn't show the issue.

stgraber (approved):

stgraber:
Oops, sorry I got distracted by all of today's drama, PR looks great now!

Thanks!

PR #297, opened 2023-12-11
Add description field on certificate entries

### LINKED ISSUE #246: Add description field on certificate entries
Certificates are one of the few objects which currently don't have a description.
With web UI and a variety of tools now using restricted certificates, it'd be good to have the ability to add comments.

presztak:
Adds description field on certificate entries and possibility to modify description using API.

Closes #246

### DIFF
--- a/internal/server/db/cluster/update.go
+++ b/internal/server/db/cluster/update.go
@@ -106,6 +106,20 @@ var updates = map[int]schema.Update{
 	67: updateFromV66,
 	68: updateFromV67,
 	69: updateFromV68,
+	70: updateFromV69,
+}
+
+// updateFromV69 adds description column to certificate.
+func updateFromV69(ctx context.Context, tx *sql.Tx) error {
+	q := `
+	ALTER TABLE certificates ADD COLUMN description TEXT NOT NULL DEFAULT "";
+	`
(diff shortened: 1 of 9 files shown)


internal/server/db/cluster/update.go
```diff
@@ -106,6 +106,20 @@ var updates = map[int]schema.Update{
 	67: updateFromV66,
 	68: updateFromV67,
 	69: updateFromV68,
+	70: updateFromV69,
+}
+
+// updateFromV69 adds description column to certificate.
+func updateFromV69(ctx context.Context, tx *sql.Tx) error {
+	q := `
+	ALTER TABLE certificates ADD COLUMN description TEXT NOT NULL;
```
stgraber, line 115: This is going to fail to run on any Incus system that has a DB entry in the certificates table.

This is going to need a `DEFAULT ""` to avoid this problem.

stgraber:
Looks good, just that small schema issue and it should be good to go.

internal/server/db/cluster/update.go
presztak: ahh, that's right. Fixed.

stgraber (approved):

stgraber:
Thanks @presztak!

PR #299, opened 2023-12-13
Make sure we don't import LXD AGPL

stgraber:


hallyn (approved):

hallyn:
:(

Makefile
```diff
@@ -99,6 +99,7 @@ endif
 	$(GO) get toolchain@none
 
 	cd cmd/lxd-to-incus && $(GO) get -t -v -d -u ./...
+	cd cmd/lxd-to-incus && $(GO) get github.com/canonical/lxd@lxd-5.19
```
tych0, line 102: This will work, but go native people will probably look for a 'replace' directive in go.mod

stgraber: Yeah, I just didn't want to risk `go mod tidy` or something messing it up down the line ;)

We already use the `go get` approach to force some version downgrades on other packages ahead of running `go mod tidy` with a specific target Go version (1.20 currently).

The other issue is that the LXD tags are not properly recognized by Go so you can't really put that `@lxd-5.19` syntax in `go.mod` itself, you need the longer expanded reference which makes it pretty hard to ensure you're looking at a commit from the Apache 2 days.

PR #303, opened 2023-12-13
Instance migration related fixes

stgraber:
Those are a couple of issues I ran into during a live demo in a recent live stream.

The first is yet another OpenFGA revert issue, apparently I had missed those code paths in my recent fixes. The second is a trivial logic bug in the client which was making the client always take the same-server code path when dealing with a project target, even with the source and target servers were different.


hallyn (approved):
don't know what flag

PR #308, opened 2023-12-15
doc: Fixed typos

colireg:
just ran `sed -i 's/A Incus/An Incus/; s/a Incus/an Incus/' */**/*.*` on the repo to fix the typos :)


stgraber (approved):
Thanks!

I actually had that on my list to do. The errors came from the initial search & replace of LXD to Incus, followed by the same issue in the tool used to include follow-up fixes :)

Issue #255, opened 2023-11-30
Switch OVN/OVS command line calls to libovsdb


stgraber:
The OVN project maintains a Go package to access the databases, it's available at https://github.com/ovn-org/libovsdb

We should slowly transition our various calls to ovn-nbctl/ovn-sbctl to using libovsdb instead, if only to not be dependent on a specific version of the OVN tooling and to also allow for more efficient and race-free interactions with OVN.

Abhiram824:
Hello! I am Abhi, a student for UT Austin. As a part of my final project in virtualization, I was interested in looking into this issue.  I see that you have assigned this already to yourself, but if you need anyone to help out on it, I would be happy to do so!

stgraber:
Hey there,

This one is one of those issues that can be worked on by multiple people in parallel so long as we don't work on the exact same function at the same time :)

Basically there are a whole bunch of functions under `internal/server/network/ovn/ovn_nb_actions.go` that currently call `o.nbctl`. All of those need to be moved to using `o.client` which is the Go-native OVN client (libovsdbclient).

The way I've handled this work so far is by picking a function that's currently using `o.nbctl`, for example let's look at `LogicalRouterAdd`. Then I always go through the same steps:
 - Rename the function to follow our new naming convention, in this case, it would become `CreateLogicalRouter`
 - Add a new `ctx context.Context` argument as the first argument of the function
 - Actually replace the code within the function to use libovsdbclient. In this case, I'd have that function look very similar to `CreateLogicalSwitch` that I've already ported.
 - Commit that work as something like: `incusd/network/ovn/nb: Port CreateLogicalRouter to OVSDB`
 - Then update any existing use of the old function to use the new one and commit that as `incusd/network/ovn: Update for function changes`

Then send that as a pull request and pick the next function to port, rinse repeat until we're done with all of them :)

Feel free to go for the easier ones if a function doesn't seem obvious to map to the structs and actions within the new client logic, just leave them and I can pick them up later.

DhruvNistala:
Hi, I am also working with Abhiram on all the issues for this project. We implemented the CreateLogicalRouter function based on the implementation of the CreateLogicalSwitch function. I created a Draft PR at https://github.com/lxc/incus/pull/750. Let me know if any additional changes need to be made

Additionally, could I be assigned to all the issues that Abhiram is as well?

stgraber:
Thanks, I'll take a look shortly!

[...]

PR #319, opened 2023-12-20
Start transition to libovsdb

stgraber:
Related to #255


stgraber:
DCO check failing because of too many commits

hallyn (approved):

PR #320, opened 2023-12-21
Set minimum OVN version to 22.03.0

stgraber:


stgraber:
Doing a manual test on this one then will force it through as Github Actions doesn't really do any useful tests on OVN.

stgraber:
Confirmed to work fine on my test systems, so the wider compatibility doesn't break the more modern systems.

PR #321, opened 2023-12-21
incusd/network/ovs: Fix empty OVNBridgeMappings

stgraber:


stgraber:
Small regression caught by the OVN tests. I've confirmed that the tests appear to be clean after this one, so will wait for the static analysis, merge it, publish new daily builds and run the full testsuite on that again before tagging 0.4.

PR #327, opened 2023-12-22
Port more code to OVSDB

stgraber:


stgraber:
This eliminates any remaining nbctl/vsctl calls during an Incus daemon restart as well as all calls to nbctl during normal runtime (excluding instance stop/start).

PR #343, opened 2023-12-29
incusd/network/ovn: Properly handle uplink with disabled DHCP/DHCPv6

stgraber:


internal/server/network/driver_ovn.go
```diff
@@ -990,7 +990,16 @@ func (n *ovn) allocateUplinkPortIPs(uplinkNet Network, routerMAC net.HardwareAdd
 					return fmt.Errorf(`Missing required "ipv4.ovn.ranges" config key on uplink network`)
 				}
 
-				ipRanges, err := parseIPRanges(uplinkNetConf["ipv4.ovn.ranges"], uplinkNet.DHCPv4Subnet())
+				dhcpSubnet := uplinkNet.DHCPv4Subnet()
+				allowedNets := []*net.IPNet{}
```
MggMuggins, line 994: Optional: Would it be possible to dispense with the array and just use a `*net.IPNet` instead? Could default to `uplinkIPv4Net` and change the pointer if dhcpSubnet isn't nil.

MggMuggins (commented):
Just one nit.

I assume this makes it possible to use a managed bridge with DHCP disabled as an UPLINK, are there any other use-cases this enables?

PR #357, opened 2024-01-03
incus: Enable cobra's completion support

stgraber:


cmd/incus/main.go
```diff
@@ -274,9 +274,14 @@ For help with any of those, simply call them with --help.`))
 		if globalCmd.flagHelpAll {
 			// Show all commands
 			for _, cmd := range app.Commands() {
+				if cmd.Name() == "completion" {
```
adamcstephens, line 277: This isn't necessary. `go run ./cmd/incus --help --all | grep completion` with just the hidden flag does not show the completion command.

stgraber: Yeah but that's because of our custom help template, that code would have otherwise been needed so I went for just making the logic consistent with the manpage one.

That way should we ever change the help template to something more standard we won't find ourselves with the command showing up.

adamcstephens:
Thanks for the quick change!

PR #364, opened 2024-01-04
lxd-to-incus: touch completion file after migration

adamcstephens:
Creating this touch file allows us to add a ConditionPathExists to the lxd service to make sure it won't start on a host that was migrated. NixOS users will have to reconfigure their host after migration to remove LXD, so there will likely be a non-zero period where both services are installed and enabled.

### DIFF
--- a/cmd/lxd-to-incus/main.go
+++ b/cmd/lxd-to-incus/main.go
@@ -765,6 +765,12 @@ Instead this tool will be providing specific commands for each of the servers.
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/lxd-to-incus/main.go
```diff
@@ -765,6 +765,12 @@ Instead this tool will be providing specific commands for each of the servers.
 		}
 	}
 
+	completeFile, err := os.Create(filepath.Join(targetPaths.Daemon, "migrated-from-lxd.complete"))
```
stgraber, line 768: Maybe `.migrated_from_lxd` for something a bit less visible?

stgraber (approved):

PR #368, opened 2024-01-05
incus/network: add dynamic command line completions

adamcstephens:
The shell handles the filtering for you, but we could try and filter by remote name for example in class someone hits tab again.

~I'll probably finish building out network to see if I can just reuse this function or the bulk of it, but this is a start.~ Done.

Example

``` fish
─❯ incus network show <TAB>
bond0    fast1           incus1:fast0  incus1:oob0      oob0
bridge0  incus1:bond0    incus1:fast1  incus1:vlan2010  vlan2010
fast0    incus1:bridge0  incus1:lo     lo

```

One can also explore these directly

```
─❯ go run ./cmd/incus __complete network show ""
incus1:lo
incus1:oob0
incus1:bond0
incus1:bridge0
incus1:vlan2010
incus1:fast0
incus1:fast1
lo
oob0
bond0
bridge0
vlan2010
fast0
fast1
:4
Completion ended with directive: ShellCompDirectiveNoFileComp
```

### DIFF
--- a/cmd/incus/completion.go
+++ b/cmd/incus/completion.go
@@ -1,0 +1,200 @@
+package main
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+func (g *cmdGlobal) cmpInstances(toComplete string) ([]string, cobra.ShellCompDirective) {
+	results := []string{}
+
+	resources, _ := g.ParseServers(toComplete)
+
+	if len(resources) > 0 {
+		resource := resources[0]
+
+		instances, _ := resource.server.GetInstanceNames("container")
+		vms, _ := resource.server.GetInstanceNames("virtual-machine")
+		instances = append(instances, vms...)
+
+		for _, instance := range instances {
+			var name string
+
+			if resource.remote == g.conf.DefaultRemote {
+				name = instance
+			} else {
+				name = fmt.Sprintf("%s:%s", resource.remote, instance)
+			}
+
+			results = append(results, name)
+		}
+	}
+
+	if !strings.Contains(toComplete, ":") {
+		remotes, _ := g.cmpRemotes()
+		results = append(results, remotes...)
+	}
+
+	return results, cobra.ShellCompDirectiveNoFileComp
+}
+
+func (g *cmdGlobal) cmpNetworks(toComplete string) ([]string, cobra.ShellCompDirective) {
+	results := []string{}
+
+	resources, _ := g.ParseServers(toComplete)
+
+	if len(resources) > 0 {
+		resource := resources[0]
+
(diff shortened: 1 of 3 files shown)


adamcstephens:
This is far from a full set of dynamic completions, but this is a good start and handles all of the basic `network` subcommands and `config show` (as a proof of concept that the functions can be reused).

Marking as ready to review so I can get some feedback.

cmd/incus/network.go
```diff
@@ -915,6 +1007,13 @@ func (c *cmdNetworkList) Command() *cobra.Command {
 	cmd.RunE = c.Run
 	cmd.Flags().StringVarP(&c.flagFormat, "format", "f", "table", i18n.G("Format (csv|json|table|yaml|compact)")+"``")
 
+	cmd.ValidArgsFunction = func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		if len(args) != 0 {
+			return nil, cobra.ShellCompDirectiveNoFileComp
+		}
+
+		return c.global.cmpRemotes()
+	}
```
stgraber, line 1016: Missing line break here

adamcstephens: Fixed. And also fixed my dependencies so I the checks locally.

cmd/incus/completion.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+
+		remoteToParse := toComplete
+
+		if strings.Contains(toComplete, ":") {
+			if remote != strings.Split(toComplete, ":")[0] {
+				continue
+			}
+		} else {
+			if toComplete != "" && remote != "local" {
```
stgraber, line 26: Global isn't always the default remote, so this probably should use `c.global.conf.DefaultRemote` instead.

stgraber:
Looking pretty good. Just the hardcoded `local` which should be replaced by the variable for the default remote.

As someone who runs with well over a dozen different remotes, a bunch of those being clusters with hundreds of instances and some of which requiring VPNs, I don't know that completing across all remotes is necessarily the right thing to do though :)

It may make more sense to only complete over the current default remote and motivate folks to `incus remote switch` as needed if they expect to do a bunch of interactions against a specific remote.

adamcstephens:
With an example:

```
─❯ ~/go/bin/incus network attach <TAB>
bond0    fast0  incus1:  local:  test1
bridge0  fast1  lo       oob0    vlan2010
─❯ ~/go/bin/incus network attach incus1:<TAB>
incus1:enp5s0  incus1:incusbr0  incus1:lo
```

stgraber:
Oh neat, that's definitely the best of all worlds!

stgraber (approved):

stgraber:
Thanks, very cool work!

PR #369, opened 2024-01-06
Rework shared/idmap

stgraber:
This re-organizes and modernizes the `shared/idmap` package to the point of mostly having it fully rewritten. The main changes are:

 - Moved everything to one of three types:
   - Range
   - Entry
   - Set
 - Re-organized filesystem structure to match the various types
 - Made the majority of the code (and test) work on all platforms
 - De-coupled and simplified the set loaders:
   - NewSetFromJSON
   - NewSetFromIncusIDMap
   - NewSetFromCurrentProcess
   - NewSetFromSystem
 - Changes to default Set contents
   - Made all sets return all entries regardless of size by default
   - Introduced `FilterPOSIX` to filter out any entry smaller than 65536 uid/gid
 - Changes to Shift/Unshift functions
   - Removed all existing Shift/Unshift functions
   - Replaced by a set of two functions
     - ShiftPath
     - UnshiftPath
   - Both functions take a path and skipper function which can be set to nil if not useful
   - The skipper function now receives the newuid and newgid and can return filepath.SkipDir/filepath.SkipAll as an error, or any other error to skip (nil to proceed)
  - Added a Split function to Set which allows finding a subset of a given size
  - Added an Includes function to Set which makes it easy to see if a set is a superset of another


tych0 (approved):
Looks good to me, this is some of the oldest code in the tree... wild.

hallyn (approved):
Thanks!

PR #409, opened 2024-01-19
cmd/incusd/api_cluster: Join cluster transactions

monstermunchkin:
This joins two consecutive cluster transactions and fixes the missing
error check in the first transaction.


stgraber:
Thanks!

PR #416, opened 2024-01-22
Add note about scrape_interval and update examples

jalbstmeijer:
The `scrape_interval` is assumed to be 15s by the Grafana Prometheus data source by default.the Grafana `$__rate_interval` value will be calculated incorrectly and possibly cause a `no data` response in queries using it.


stgraber (approved):

stgraber:
Thanks!

PR #417, opened 2024-01-22
cmd/incus: Remove unused `flagContent` variable in `incus file create`

monstermunchkin:
This removes the leftover `flagContent` variable in `incus file create`.


stgraber (approved):
Oops :)

PR #419, opened 2024-01-22
Fix in-cluster storage volume refreshes

stgraber:


stgraber:
When going through this logic, I noticed that our POST handler for the two different kind of storage volume URLs were widely different which seems a bit concerning when the only difference should be that one requires req.Type and the other doesn't.

So I've unified them now too.

hallyn (approved):
Had to look a this for quite a while to convince myself nothing was ended up missing :)

PR #420, opened 2024-01-23
lxd-to-incus: Stop using LXD client package

stgraber:


freeekanayaka (commented):
Nice!

@gibmat with the one we can drop the dependency on lxd

freeekanayaka (approved):

PR #427, opened 2024-01-24
Add support for multiple agent binaries

### LINKED ISSUE #263: Ship multiple VM agent binaries
Currently incus looks for the `incus-agent` binary in the host PATH and copies that into the VM config drive. That works well enough but has quite a few drawbacks:
 - As it's a copy, the guest cannot refresh its agent without a reboot
 - Only a single OS and architecture can be made available (that of the host)
 - Space and time is wasted as the binary is copied for every VM

Instead I believe we should:
 - Ship a directory containing a variety of agent builds, likely named after `OS.ARCH` (e.g. `linux.x86_64`).
 - That directory should be bind-mounted into the config drive
 - The guest should be made to compare its agent with the one in the drive any time it gets reconnected with the host, if the two differ, the agent should exit so it can be restarted and get the new version loaded.

stgraber:
This adds support for a separate 9p/virtiofs share for the agent which will be shared system-wide (saves us from data duplication) and can contain multiple binaries to handle multiple architectures and operating systems.

Closes #263

### DIFF
--- a/doc/environment.md
+++ b/doc/environment.md
@@ -29,6 +29,7 @@ Name                            | Description
 
 Name                            | Description
 :---                            | :----
+`INCUS_AGENT_PATH`              | Path to the directory including the `incus-agent` builds
 `INCUS_CLUSTER_UPDATE`          | Script to call on a cluster update
 `INCUS_DEVMONITOR_DIR`          | Path to be monitored by the device monitor. This is primarily for testing
 `INCUS_DOCUMENTATION`           | Path to the documentation to serve through the web server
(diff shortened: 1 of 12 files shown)


stgraber:
Keeping this as draft for now as I still need to add the virtiofsd support.

Current code only provides the `agent` share over 9p which may not be sufficient with newer distros only shipping virtiofs.

stgraber:
Right, so my plan to use a shared virtiofsd for all VMs doesn't actually work.
It looks like you can't actually share a daemon in that way.

The alternative would be a daemon per VM on top of the existing one for the main config drive.
That doesn't really sound too good to me, especially when configuring how little use those drives get. They're basically just accessed at boot to get some files copied and that's it.

So instead going with the other option which is to remove virtiofs for the config drive too, having both config and agent drives be read-only 9p only.

virtiofs will obviously still be used for any user configured shares as those actually benefit from its better throughput and lower latency.

PR #446, opened 2024-01-28
Add support for an agent drive (ISO)

stgraber:
With this, a new `agent:config` disk source can be used to attach a custom ISO to a VM.
This ISO then contains roughly the same data as the normal `9p` config drive.

The goal behind this is to provide a way for VMs that don't support 9p but do support virtio-serial and virtio-vsock to still run a functional incus-agent.


stgraber:
As that extra drive contains sensitive material (TLS keys to talk to the host), the loader is designed to always eject the drive if found. That's combined with us now having logic to actually handle eject requests.

That way, the VM boots, copies the data from the drive and then immediately ejects it so there's no risk of that drive getting accidentally exposed to an unprivileged user down the line.

(The whole thing also requires the user to have manually asked Incus to add such a drive in the first place, so it's pretty much limited to the few Linux distros that don't offer 9p)

PR #453, opened 2024-01-30
fix: use genisoimage when mkisofs is not found

bketelsen:
Fedora variants don't include `mkisofs`, but do include `genisoimage` which is a command compatible substitute. This change checks for `genisoimage` and uses if `mkisofs` isn't found.

### DIFF
--- a/internal/server/device/disk.go
+++ b/internal/server/device/disk.go
@@ -2208,10 +2208,15 @@ func (d *disk) generateVMAgentDrive() (string, error) {
 	scratchDir := filepath.Join(d.inst.DevicesPath(), linux.PathNameEncode(d.name))
 	defer func() { _ = os.RemoveAll(scratchDir) }()
 
-	// Check we have the mkisofs tool available.
-	mkisofsPath, err := exec.LookPath("mkisofs")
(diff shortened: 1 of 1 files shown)


internal/server/device/disk.go
```diff
@@ -2208,10 +2208,15 @@ func (d *disk) generateVMAgentDrive() (string, error) {
...
-	mkisofsPath, err := exec.LookPath("mkisofs")
+	// Check we have the mkisofs or genisoimage tool available.
+	var mkisofsPath string
+	var err error
+	mkisofsPath, err = exec.LookPath("mkisofs")
 	if err != nil {
-		return "", err
+		mkisofsPath, err = exec.LookPath("genisoimage")
+		if err != nil {
+			return "", fmt.Errorf("Neither mkisofs or genisoimage found in $PATH")
```
joshua-stone, line 2218: I suggest having `"Neither mkisofs nor genisoimage found in $PATH"` instead.

stgraber:
Thanks, I've applied the suggestion and changed the commit title, just going to let some tests run through and then I'll merge this.

stgraber (approved):

PR #467, opened 2024-02-04
Prevent custom storage volumes of type block to be used more than once

monstermunchkin:


internal/server/instance/instance_utils.go
```diff
@@ -1257,3 +1258,97 @@ func SnapshotProtobufToInstanceArgs(s *state.State, inst Instance, snap *migrati
 
 	return &args, nil
 }
+
+// AllowedInstanceDevices checks whether custom block volumes are allowed to be attached to the given instance.
+// This is done by checking whether the provided devices are already attached to other instances.
+func AllowedInstanceDevices(s *state.State, instanceID int, projectName string, devices ...deviceConfig.Devices) error {
+	volumeDevices := make(map[int64][]cluster.Device)
+
+	err := s.DB.Cluster.Transaction(s.ShutdownCtx, func(ctx context.Context, tx *db.ClusterTx) error {
```
stgraber, line 1273: This seems extremely inefficient and will significantly slow down startup in busy environments.

I suspect we'd get much better performance by moving this check from instance startup over to instance config validation.

Basically:
 - Refuse adding a block custom volume to a profile unless security.shared=true is set
 - Refuse adding a block custom volume to an instance if it's already present on another instance
 - Refuse setting security.shared=false on a custom volume that's currently used by more than one instance

Note that your check was also incorrect as you need to keep in mind that volume `foo` in pool `bar` on server `s1` is different from volume `foo` in pool `bar` on server `s2`.

With those checks being at configuration time rather than start time, the performance impact should be significantly lower and we'll save a lot of DB queries on the hot path.

stgraber:
@monstermunchkin oops, looks like you need a quick rebase for the API extension.

monstermunchkin:
> @monstermunchkin oops, looks like you need a quick rebase for the API extension.

Fixed.

internal/server/device/disk.go
```diff
@@ -299,34 +299,92 @@ func (d *disk) validateConfig(instConf instance.ConfigReader) error {
...
 			d.pool, err = storagePools.LoadByName(d.state, d.config["pool"])
 			if err != nil {
 				return fmt.Errorf("Failed to get storage pool %q: %w", d.config["pool"], err)
 			}
 
+			// Derive the effective storage project name from the instance config's project.
+			storageProjectName, err = project.StorageVolumeProject(d.state.DB.Cluster, instConf.Project().Name, db.StoragePoolVolumeTypeCustom)
+			if err != nil {
+				return err
+			}
```
stgraber, line 315: Hmm, is that right?

What happens if you're dealing with a project that has `features.storage.volumes=false`?
In that case, the instance would be in the project but the attached volume would come from the default project instead.

internal/server/device/disk.go
```diff
@@ -299,34 +299,92 @@ func (d *disk) validateConfig(instConf instance.ConfigReader) error {
...
+			contentType, err := storagePools.VolumeContentTypeNameToContentType(dbVolume.ContentType)
+			if err != nil {
+				return err
+			}
+
+			// Check that only shared custom storage block volume are added to profiles, or multiple instances.
+			if util.IsFalseOrEmpty(dbVolume.Config["security.shared"]) && contentType == db.StoragePoolVolumeContentTypeBlock {
+				if instConf.Type() == instancetype.Any {
+					return fmt.Errorf("Cannot add un-shared custom storage block volume to profile")
+				} else {
```
stgraber, line 336: No need for the else statement and indent since the previous block ends with a return.

internal/server/device/disk.go
stgraber: Oh, I see, that's pre-existing logic and I take it StorageVolumeProject performs the correct check.

stgraber:
Just a couple small nits/questions and this will be good to go, thanks!

stgraber (approved):

PR #468, opened 2024-02-05
Improve and document support for encrypted keys

PriceChild:
Gives hints and limitations on encrypting the client key.

### DIFF
--- a/doc/.wordlist.txt
+++ b/doc/.wordlist.txt
@@ -5,6 +5,8 @@ ACLs
 AIO
 allocator
 AMD
+Ansible
+Ansible's
 APIs
 AppArmor
 ARMv
--- a/doc/authentication.md
+++ b/doc/authentication.md
@@ -109,6 +109,28 @@ If the server certificate is valid and signed by the CA, then the connection con
 
 Note that the generated certificates are not automatically trusted. You must still add them to the server in one of the ways described in {ref}`authentication-trusted-clients`.
 
+### Encrypting local keys
+
+The `incus` client also supports encrypted client keys. Keys generated via the methods above can be encrypted with a password, using:
+
+```
+ssh-keygen -p -o -f .config/incus/client.key
+```
+
+```{note}
+Unless you enable [`keepalive` mode](remote-keepalive), then every single call to Incus will cause the prompt which may get a bit annoying:
+
+    $ incus list remote-host:
+    Password for client.key:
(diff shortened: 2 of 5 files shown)


PriceChild:
* I haven't added the Signed-Off line yet. Will sort tomorrow.
* Worried about whether this advice is safe, given discussions like this: https://security.stackexchange.com/a/203384/24342
* Had a look at the ansible connection plugin, saw it'd likely need a change to the incus cli to accept client key before prompt?
* While building docs, had to install a few other dependencies not mentioned on https://linuxcontainers.org/incus/docs/main/installing/#installing-from-source (go 1.21 - jammy defaulted to 1.18, and also python3.10-venv for sphinx for `make doc`) Might look at pr'ing that too.
* I couldn't find any better link for keepalive than the news page? Is there a reference for remote server config?

PriceChild:
@stgraber I'm missing something obvious with `make doc-lint`, running it locally gives:
```
 make doc-lint
doc/.sphinx/.markdownlint/doc-lint.sh
Passed!
```
This didn't match the output of github actions?

I can run the actions in my own personal fork but that's obviously not ideal: https://github.com/PriceChild/incus/actions/runs/7803489771/job/21283375479

stgraber:
> [make doc-spellcheck](https://linuxcontainers.org/incus/docs/main/contributing/#automatic-documentation-checks) has a couple of complaints...

Fixed those, it's the usual issue with capitalization of Incus and having a bunch of other stuff be escaped.

> I couldn't find any better link for keepalive than the news page? Is there a reference for remote server config?

I've added a commit which adds a section for this now.

> Had a look at the ansible connection plugin, saw it'd likely need a change to the incus cli to accept client key before prompt?

Yeah, that may need tweaking. You can work around it through the `keepalive` by first making a manual `incus info remote:` call to prompt for the key and start keepalive, then run Ansible before the timeout hits.

> Worried about whether this advice is safe, given discussions like this: https://security.stackexchange.com/a/203384/24342

I've added support for SSH's encrypted key format, currently limited to ECDSA and update the documentation to recommend that for the EC case.

stgraber (approved):

stgraber:
I've expanded support to also cover RSA keys through SSH encryption and so updated the instructions to always use ssh-keygen.

PriceChild:
Thank you for sorting that all out. I'm still not sure why the linting tools gave different results locally. Looking..

PR #516, opened 2024-02-22
incusd/instance/qemu: Cap hotplug CPU slots to 64

stgraber:
QEMU doesn't like dealing with more than 240 hotplug CPU slots and some OS struggle with more than 64, so put a cap to the max exposed CPUs to 64 unless the user has purposefully exposed more.

Sponsored-by: ActivePort (https://www.activeport.com.au)


stgraber:
Adding a couple of extra changes in this one to fix some test failures.

PR #528, opened 2024-02-23
Add new image requirement for cdrom drive

stgraber:


internal/server/instance/drivers/driver_qemu.go
```diff
@@ -1041,6 +1041,31 @@ func (d *qemu) validateStartup(stateful bool, statusCode api.StatusCode) error {
...
+		found := false
+		for _, dev := range d.expandedDevices {
+			if dev["type"] == "disk" && dev["source"] == "agent:config" {
+				found = true
+				break
+			}
+		}
+
+		if !found {
+			return fmt.Errorf("This virtual machine image requires an agent cdrom drive be added.")
```
monstermunchkin, line 1065:
```suggestion
			return fmt.Errorf("This virtual machine image requires an agent cdrom drive be added")
```

monstermunchkin (changes requested):

internal/server/instance/drivers/driver_qemu.go
stgraber: Thanks!

monstermunchkin (approved):

hallyn (approved):

PR #529, opened 2024-02-23
incusd/device/disk: Fix incorrect block volume usage

### LINKED ISSUE #530: vm with custom storage block volume fails to start with: Cannot add un-shared custom storage block volume
* Distribution: Ubuntu 
 * Distribution version: 22.04
 * Incus version: 0.6
 * Storage backend in use: zfs

existing vm having a second disk attached fails to start after update to incus version 0.6 with:
```
Error: Failed start validation for device "c1-swap": Cannot add un-shared custom storage block volume to more than one instance
```

# Steps to reproduce

 1. create  and stop vm
 2. create custom storage volume
 3. attach custom storage volume to the created vm
 4. start vm --> fails with above error message

```
incus storage volume show ssd1 c1-swap
config:
  security.shared: "true"
  size: 8GB
description: ""
name: c1-swap
type: custom
used_by:
- /1.0/instances/c1
location: none
content_type: block
project: default
```

stgraber:
Closes #530

### DIFF
--- a/internal/server/device/disk.go
+++ b/internal/server/device/disk.go
@@ -339,6 +339,11 @@ func (d *disk) validateConfig(instConf instance.ConfigReader) error {
 				var usedBy []string
 
 				err = storagePools.VolumeUsedByInstanceDevices(d.state, d.pool.Name(), storageProjectName, &dbVolume.StorageVolume, true, func(inst db.InstanceArgs, project api.Project, usedByDevices []string) error {
+					// Don't count the current instance.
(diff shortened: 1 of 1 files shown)


stgraber:
Our CI testing did catch this, but a bit too late as it was clean prior to the 0.6 release but turned red shortly after. This suggests that the Jenkins test ran right before the new build with the security.shared change :(

hallyn (approved):

PR #513, opened 2024-02-20
Disable the architecture check on `incus copy`/`incus mv`

presztak:
1. Disable the architecture check when performing `incus cp`/`incus mv`.
2. Check architecture when trying to run instance.

Fixes: #478

PR #535, opened 2024-02-24
Disable architecture check on `incus cp` with snapshots

presztak:
Fixes issue with copying instances which have snapshots across machines with different architecture.


stgraber (approved):

stgraber:
@presztak any reason to keep this one as draft?

presztak:
@stgraber it's in draft because I was not able to reproduce second issue from here: https://github.com/lxc/incus/pull/513#issuecomment-1962582642. But in overall I believe we can undraft.

stgraber:
Okay, should be an improvement anyway, then we can track the other problem on its own if it still occurs.

Issue #245, opened 2023-11-28
Add suport for forceful project deletion


stgraber:
When dealing with temporary projects, it's a bit annoying having to figure out everything that needs to be removed before being able to delete the project.

It'd be useful to have a force flag which just goes and delete everything in the project instead.

dnegreira:
Can I have a stab at this one?

stgraber:
@dnegreira of course!

stgraber:
@dnegreira I suspect you're going to want to:
 - Add an API extension string for this in doc/api-extensions.md and internal/version/api.go
 - Add a DeleteProjectForce function to the client package which passes a `?force=true` to the API
 - Update the projectDelete API handler to handle the `force` variable and if set to true, then go through everything the project owns and delete it. You probably want to start with instances, then volumes and buckets, then profiles, then networks (starting with acls, zones, peering before networks themselves).

stgraber:
Unassigned this one per @dnegreira's request, it's available for anyone who wants it!

MaheshPunjabi:
Can I try?

stgraber:
Of course!

MaheshPunjabi:
Few follow up questions:
1. Is it fine to specify force as a query type parameter ? Didn't see any existing query parameters, so I thought its best to check. I assume this is yes coz of use of ?force=true
2. When deleting artifacts, do we also want to delete images(if features.images is set) ?
3. Does this need any changes to client/simplestreams ?

Proposed changes:
- [ ] Add an API extension string for this in doc/api-extensions.md and internal/version/api.go. _call it "projects_force_delete"_

- [ ] Update the projectDelete API handler,_in "cmd/incusd/api_project.go"_ to handle the force variable and if set to true, then go through everything the project owns and delete it. You probably want to start with instances, then volumes and buckets, then profiles, then networks (starting with acls, zones, peering before networks themselves).

- [ ] _Re-generate the swagger YAML data (make update-api)._

- [ ] Add a DeleteProjectForce function to "client/incus_projects.go" which passes a ?force=true to the API. _add that function to the ImageServer interface._

- [ ] _Add new "force" argument to "cmd/incus/project.go". when passed, call DeleteProjectForce_
- [ ] _Re-generate the translation template for the new CLI string._

Proposed commits:
api: projects_force_delete
incusd/api_project: Add support for force delete
doc/rest-api: Refresh swagger YAML
client: Add DeleteProjectForce 
incus/projects: add --force flag
[...]

PR #536, opened 2024-02-26
I245 add suport for forceful project deletion

MaheshPunjabi:
What was changed?
from https://github.com/lxc/incus/issues/245 
- [x] api: projects_force_delete
Add an API extension string for this in doc/api-extensions.md and internal/version/api.go. call it "projects_force_delete"
---
Work-In-Progress:
- [ ] Update the projectDelete API handler,in "cmd/incusd/api_project.go" to handle the force variable and if set to true, then go through everything the project owns and delete it. You probably want to start with instances, then volumes and buckets, then profiles, then networks (starting with acls, zones, peering before networks themselves).

- [ ] Re-generate the swagger YAML data (make update-api).

- [ ] Add a DeleteProjectForce function to "client/incus_projects.go" which passes a ?force=true to the API. add that function to the ImageServer interface.

- [ ] Add new "force" argument to "cmd/incus/project.go". when passed, call DeleteProjectForce

- [ ] Re-generate the translation template for the new CLI string.

---
Tests:
gofmt -d passed(for api.go)
make completed fine


stgraber:
@MaheshPunjabi how's this going? Need any help?

MaheshPunjabi:
> @MaheshPunjabi how's this going? Need any help?
Yes, thank you. I've posted a couple of questions on line 1570, 71 and 91.
Also, I'm still trying to figure out the step "Empty the default profile" . Should I just call projectCreateDefaultProfile ?

stgraber:
Ah, in the future please use messages in the PR directly or use Github's in-line comment feature as we don't normally review the code within draft PRs so will not notice code comments like the ones you've left.

cmd/incusd/api_project.go
```diff
@@ -1541,3 +1566,93 @@ func projectValidateRestrictedSubnets(s *state.State, value string) error {
 
 	return nil
 }
+
+// Do I need to check the StatusCode and/or ExpandedConfig similar to "cmd/incus/delete.go"
```
stgraber, line 1570: No, that shouldn't apply to an internal call., but you do need to make sure the instance is stopped before you delete it.

cmd/incusd/api_project.go
```diff
@@ -1541,3 +1566,93 @@ func projectValidateRestrictedSubnets(s *state.State, value string) error {
...
+// Do I need to check if this a snapshot or of type container similar to "cmd/incusd/instance_delete.go"
```
stgraber, line 1571: GetInstances() should only return you instances so that should be fine, their snapshots will go away with them.

cmd/incusd/api_project.go
```diff
@@ -1541,3 +1566,93 @@ func projectValidateRestrictedSubnets(s *state.State, value string) error {
...
+
+	if len(instances) == 0 {
+		return nil
+	}
+
+	for _, instance := range instances {
+		if err := cluster.DeleteInstance(ctx, tx.Tx(), project.Name, instance.Name); err != nil {
+			return err
+		}
+	}
```
stgraber, line 1587: Note that this is only deleting the database records so that's not going to work.

You instead need to load the instance, make sure it's stopped and then call `inst.Delete(true)` to have it and all its snapshots properly deleted.

cmd/incusd/api_project.go
```diff
@@ -1541,3 +1566,93 @@ func projectValidateRestrictedSubnets(s *state.State, value string) error {
...
+	return nil
+}
+
+// is there any value in checking the UsedBy ?
```
stgraber, line 1591: Not if all instances are deleted first.

cmd/incusd/api_project.go
```diff
@@ -1541,3 +1566,93 @@ func projectValidateRestrictedSubnets(s *state.State, value string) error {
...
+		}
+		for _, profile := range profiles {
+			if profile.Name == "default" {
+				continue
+			}
+			if err := cluster.DeleteProfile(ctx, tx.Tx(), project.Name, profile.Name) ; err != nil {
+				return err
+			}
+		}
+	}
```
stgraber, line 1612: You don't need that much indentation.
You can probably just use:
```
if len(profiles) == 1 {
    return nil
}

for _, profile := range profiles {
    if profile.Name == "default" {
        continue
    }

    err := cluster.DeleteProfile(ctx, tx.Tx(), project.Name, profile.Name)
    if err != nil {
        return err
    }
}
```

cmd/incusd/api_project.go
MaheshPunjabi: fixed code to check if its running and stop

cmd/incusd/api_project.go
MaheshPunjabi: fixed in code

cmd/incusd/api_project.go
MaheshPunjabi: thanks. I added a comment about this.

cmd/incusd/api_project.go
MaheshPunjabi: got it. made the change.

cmd/incusd/api_project.go
```diff
@@ -1541,3 +1566,93 @@ func projectValidateRestrictedSubnets(s *state.State, value string) error {
...
+	if len(profiles) > 0 {
+		// Consider the project empty if it is only used by the default profile.
+		if len(profiles) == 1 && profiles[0].Name == "default" {
+			return nil
+		}
+		for _, profile := range profiles {
+			if profile.Name == "default" {
+				continue
+			}
+			if err := cluster.DeleteProfile(ctx, tx.Tx(), project.Name, profile.Name) ; err != nil {
```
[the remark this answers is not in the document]
MaheshPunjabi: understood. I made changes just now.

MaheshPunjabi:
@stgraber Can you take a quick look at the changes in recent commits to see if they look good? I'm doubtful on the "Empty the default profile" part.

stgraber:
I think the DB code to empty the default profile should work. That said, I don't think it's a good idea to directly poke the database as in many other cases (networks, storage, instance, ...) there is more to Incus than just some DB records and this will not properly clean things.

A more reliable approach would be to use our own Go API to perform those actions, effectively acting as if the `incus` command line tool was issuing those commands.

Basically doing something like:
```
s, err := incus.ConnectIncusUnix("", nil)
if err != nil {
    return err
}
```

And then using `s` to call things like `DeleteInstance`, `DeleteProfile`, `UpdateProfile` (to empty the `default` profile`, `DeleteStoragePoolVolume`, `DeleteStoragePoolBucket`, ...

You can compute the actual list of stuff to delete based on a `GetProject` call and looking at the `UsedBy` field. This also makes it possible to immediately fail if there are things in `UsedBy` that the function doesn't yet know how to delete (future proofing).

stgraber:
Going to take this one and the work done in #850 to get something we can merge in 6.2.

PR #552, opened 2024-02-28
incus: add completions for remotes and projects

adamcstephens:
I couldn't find an available list of project config keys to replace the ones from the legacy bash completions. I suspect those were manually entered into that file? This missing available list of all config keys only affects `project set` as the retrieval of keys can use the existing project's config.

These project configs appear to be only codified here, unless I'm missing something https://github.com/lxc/incus/blob/d4d6c93283f0d266d283f949c0c9dd03a297d62a/cmd/incusd/api_project.go#L1066

### DIFF
--- a/cmd/incus/project.go
+++ b/cmd/incus/project.go
@@ -97,6 +97,14 @@ func (c *cmdProjectCreate) Command() *cobra.Command {
 
 	cmd.RunE = c.Run
 
+	cmd.ValidArgsFunction = func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		if len(args) == 0 {
+			return c.global.cmpRemotes(false)
+		}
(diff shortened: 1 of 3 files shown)


cmd/incus/project.go
```diff
@@ -797,14 +888,22 @@ type cmdProjectInfo struct {
 
 func (c *cmdProjectInfo) Command() *cobra.Command {
 	cmd := &cobra.Command{}
-	cmd.Use = usage("info", i18n.G("[<remote>:]<project> <key>"))
+	cmd.Use = usage("info", i18n.G("[<remote>:]<project>"))
```
adamcstephens, line 891: Corrected this, which doesn't actually accept a second position argument.

stgraber (approved):

stgraber:
Thanks!

PR #565, opened 2024-02-29
incusd/certificates: Improve token handling when clustered

montag451:
Hi,

I have an Incus cluster which is reachable through a VIP managed by Keepalived. The `core.https_address` of each node is set to listen on the VIP (net.ipv4.ip_nonlocal_bind=1 on each node). To add a remote client I followed this procedure:

1. From one of the node (**not the one with the VIP**), I created a token using `incus config trust add <client_name>`
2. On the client I created a new remote using `incus remote add <remote_name> <token>`

 But unfortunately it failed with a cryptic error message `Error: Failed to create certificate: Bad certificate add operation data`. If I create the token on the node with the VIP, everything is OK.

Digging through the code I discovered that the problem is due to the token operation retrieval. The code assume that the token operation is a local operation but in fact the operation can be a remote one : in my case, the token was created on one node, but the token processing was done on the node with the VIP. This PR fixes this issue.

### DIFF
--- a/cmd/incusd/certificates.go
+++ b/cmd/incusd/certificates.go
@@ -574,19 +574,24 @@ func certificatesPost(d *Daemon, r *http.Request) response.Response {
 					return response.Forbidden(fmt.Errorf("No matching certificate add operation found"))
 				}
 
-				tokenReq, ok := joinOp.Metadata["request"].(api.CertificatesPost)
-				if !ok {
-					return response.InternalError(fmt.Errorf("Bad certificate add operation data"))
-				}
-
 				// Create a new request from the token data as the user isn't allowed to override anything.
-				req = api.CertificatesPost{
-					CertificatePut: api.CertificatePut{
-						Name:       tokenReq.Name,
-						Type:       tokenReq.Type,
-						Restricted: tokenReq.Restricted,
-						Projects:   tokenReq.Projects,
-					},
+				req = api.CertificatesPost{}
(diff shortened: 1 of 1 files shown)


cmd/incusd/certificates.go
```diff
@@ -574,19 +574,24 @@ func certificatesPost(d *Daemon, r *http.Request) response.Response {
...
+					req.Type = tokenReq.Type
+					req.Restricted = tokenReq.Restricted
+					req.Projects = tokenReq.Projects
+				case map[string]any:
+					req.Name = tokenReq["name"].(string)
+					req.Type = tokenReq["type"].(string)
+					req.Restricted = tokenReq["restricted"].(bool)
+					for _, project := range tokenReq["projects"].([]any) {
+						req.Projects = append(req.Projects, project.(string))
+					}
```
stgraber, line 591: Is there any possibility (user controlled) for any of those fields to be of different types?

Just want to make sure there's no way for someone to provide a bogus join token and crash Incus as a result.

montag451: I don't think it's possible because the operation is retrieved from the node where the token has been created. `tokenReq` is just the result of the serialization of an `api.CertificatesPost` from the node where the token was issued. But I understand your concern so I can check all fields if you want.

stgraber (approved):

stgraber:
Confirmed that the code path is safe, thanks!

PR #588, opened 2024-03-07
incus/completion: do not add space after remote names completion

montag451:


adamcstephens:
@montag451 Can you share which shell are you using? I haven't experienced this problem in my use with fish, but I'll do some more testing on this. More completions are already in the works, so I'll need to know what the best fix is.

There's also `ShellCompDirectiveNoSpace` which maybe makes some of this extra ceremony unnecessary?

montag451:
@adamcstephens: I'm using Bash. For example, if I type `incus exec loc` and ask for completion I get `incus exec local: `. My fix just uses the completion directive returned by `cmpRemotes` instead of discarding it as `cmpRemotes` already returns the right directive which is `ShellCompDirectiveNoSpace`.

stgraber (approved):

adamcstephens:
I suspect this may flip the problem though. Now if you tab complete a local resource /without/ a remote, it will no longer add the trailing space which would be desirable. If I'm thinking about this correctly, either way someone will have to type the space depending on the situation.

montag451:
@adamcstephens I agree but I think that most of the time it's more convenient to add a space than to remove it. But I agree it's very subjective :)

PR #597, opened 2024-03-08
incus: add completions for storage pools and volumes

adamcstephens:
Is it ok to bring in code from `internal/server` ? Or is there another place to get the list of all available storage config options besides https://github.com/lxc/incus/blob/cff35a29ee3d7a2af1f937cbb6cf23776941854b/internal/server/storage/utils.go#L478 This could be used for get/unset for pools/volumes

I also couldn't find a list of available storage drivers, but this is only needed for pool create.

### DIFF
--- a/cmd/incus/completion.go
+++ b/cmd/incus/completion.go
@@ -104,6 +104,23 @@ func (g *cmdGlobal) cmpInstances(toComplete string) ([]string, cobra.ShellCompDi
 	return results, cobra.ShellCompDirectiveNoFileComp
 }
 
+func (g *cmdGlobal) cmpInstanceNamesFromRemote(toComplete string) ([]string, cobra.ShellCompDirective) {
+	results := []string{}
+
+	resources, _ := g.ParseServers(toComplete)
+
+	if len(resources) > 0 {
+		resource := resources[0]
+
+		containers, _ := resource.server.GetInstanceNames("container")
+		results = append(results, containers...)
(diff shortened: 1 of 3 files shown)


stgraber:
In general you can import from internal in the CLI but importing from internal/server sounds isn't ideal as there's a high chance that this imports a bunch more things and may pull cgo code too.

As we're getting most config keys to be exported to a json file for documentation purposes, I think it will make sense to have the CLI similarly either import the json data at build time or even take it one step further and directly fetch the list of valid config keys through the API.

adamcstephens:
> directly fetch the list of valid config keys through the API.

This sounds like a great idea to me. Does this API exist already? 

There are a handful of key lists manually maintained in the old bash completion, and I'm doing my best to get parity (and then some) with the old completions. For an individual resource's config keys, I can just pull them from the resource.

Unfortunately, I've only found a nice list of all keys for instances here: https://github.com/lxc/incus/blob/cff35a29ee3d7a2af1f937cbb6cf23776941854b/internal/instance/config.go#L30

It's also possible I'm missing something. :)

stgraber:
`incus query /1.0/metadata/configuration` shows you what's currently in there, though I need to clean that stuff up a bit because there's no matching Go client function for it apparently...

We should get a `GetMetadataConfiguration()` function to grab that data.

adamcstephens:
Sounds good. This should still be ok to merge now, and I can add completions for the missing config keys once #600 is done.

stgraber (approved):

PR #610, opened 2024-03-12
internal/usbid: allow path override of usb.ids path

adamcstephens:


internal/usbid/load.go
```diff
@@ -30,7 +30,12 @@ var (
 
 // Load reads the USB database from disk.
 func Load() {
-	usbids, err := os.Open("/usr/share/misc/usb.ids")
+	usbidFile := os.Getenv("INCUS_USBIDS_PATH")
```
stgraber, line 33: Can you add a second commit updating `doc/environment.md` to document that config key?

stgraber (approved):

PR #612, opened 2024-03-12
cmd/incusd: support for JWT authentication

mikerobski:
Added support for JWT authentication of REST API requests.
To use the JWT authentication the client should send Authorization header with Bearer token.
The Bearer token contains the JWT signed with the client certificate.

The JWT token must have the following content:
{
  "sub":"axxxxxxxx",
  "exp":1710288000,
  "nbf":1710201600,
  "iat":1710201600
}

The sub parameter contains the client certificate fingerprint, calculated as the SHA256 checksum of the raw certificate.

The authentication of the API request checks for the presence of the Authorization header and if it is present and contains valid JWT token, uses the subject to find the trusted client certificate and verifies the JWT signature.

### DIFF
--- a/doc/.wordlist.txt
+++ b/doc/.wordlist.txt
@@ -99,6 +99,7 @@ hotplug
 hotplugged
 hotplugging
 HTTPS
+hwdata
 ICMP
 idmap
 idmapped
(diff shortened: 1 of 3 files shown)


stgraber:
Hey there, can you rebase your branch to get rid of those two merge commits and get the first two commits merged into one?

Also, please make sure you sign-off your changes (Signed-off-by line per CONTRIBUTING.md).

PR #613, opened 2024-03-12
cmd/incusd: support for JWT authentication

mikerobski:
Added support for JWT authentication of REST API requests.
To use the JWT authentication the client should send Authorization header with Bearer token.
The Bearer token contains the JWT signed with the client certificate.

The JWT token must have the following content:
{
"sub":"axxxxxxxx",
"exp":1710288000,
"nbf":1710201600,
"iat":1710201600
}

The sub parameter contains the client certificate fingerprint, calculated as the SHA256 checksum of the raw certificate.

The authentication of the API request checks for the presence of the Authorization header and if it is present and contains valid JWT token, uses the subject to find the trusted client certificate and verifies the JWT signature.


internal/server/util/http.go
```diff
@@ -288,3 +291,67 @@ func IsJSONRequest(r *http.Request) bool {
...
+
+// CheckJwtToken checks whether the given request has JWT token that is valid and
+// signed with client certificate from the trusted certificates.
+// Returns whether or not the token is valid, the fingerprint of the certificate and the certificate.
+func CheckJwtToken(r *http.Request, trustedCerts map[string]x509.Certificate) (bool, string, *x509.Certificate) {
+	// Check if the request has a JWT token
+	auth := r.Header.Get("Authorization")
+	if auth == "" {
+		return false, "", nil
+	}
```
stgraber, line 303: Line break here

internal/server/util/http.go
```diff
@@ -288,3 +291,67 @@ func IsJSONRequest(r *http.Request) bool {
...
+	// Check if the token is valid
+	if time.Now().Before(requestTokenClaims.NotBefore.Time) || time.Now().After(requestTokenClaims.ExpiresAt.Time) {
+		return false, "", nil // token not yet valid or expired
+	}
+
+	// Find the certificate by token subject
+	requestTokenCert, ok := trustedCerts[requestTokenClaims.Subject]
+	if !ok {
+		return false, "", nil // certificate not found
+	}
```
stgraber, line 333: We need to validate that the certificate is still valid at this stage since we don't have the TLS handler doing that for us here.

stgraber (changes requested):
Looking good, just a couple small changes needed.

We'll also want to add a new API extension for this, let's call it `auth_tls_jwt`. For that you want an extra commit title `api: auth_tls_jwt` which adds the extension to both doc/api-extensions.md and internal/version/api.go.

And we should also ideally add a test for this. Probably introduce a small test tool in tests/ which you can feed a client.crt+client.key, a not-before and not-after timestamp and have it return you the JWT.

We can then use that in the testsuite to validate that:
 - An invalid token is rejected
 - A valid token is properly reported as the user that signed it
 - A token that's not valid yet is not accepted
 - A token that's no longer valid is not accepted

stgraber:
So per above, we'll likely end up with a PR containing:
 - api: auth_tls_jwt
 - incusd: Add support for JWT authentication
 - tests: Introduce tls2jwt tool
 - tests: Test JWT authentication

internal/server/util/http.go
mikerobski: The certificate validation is in the cmd/incusd/daemon.go:470
We can move it here if more appropriate.

stgraber: Oh, sorry I missed that. It's fine to do it from daemon.go.

Issue #598, opened 2024-03-09
Restricted user cannot copy images between projects


victoitor:
# Issue description

If a restricted user tries to copy an image between projects but does not have access to the default project, then the copy fails with a message of `Error: User does not have permission for project "default"`.

# Steps to reproduce

 1. Have a user with access at least two projects, but not the default project
 2. Take an image from one project (I used publish from a container)
 3. Try to copy it into another project

These steps can be seen in the following commands.

```
victoitor@bastion:~$ incus remote list
+-----------------+------------------------------------+---------------+-------------+--------+--------+--------+
|      NAME       |                URL                 |   PROTOCOL    |  AUTH TYPE  | PUBLIC | STATIC | GLOBAL |
+-----------------+------------------------------------+---------------+-------------+--------+--------+--------+
| images          | https://images.linuxcontainers.org | simplestreams | none        | YES    | NO     | NO     |
+-----------------+------------------------------------+---------------+-------------+--------+--------+--------+
| local           | unix://                            | incus         | file access | NO     | YES    | NO     |
+-----------------+------------------------------------+---------------+-------------+--------+--------+--------+
| pargo (current) | https://10.11.16.11:8443           | incus         | tls         | NO     | NO     | NO     |
+-----------------+------------------------------------+---------------+-------------+--------+--------+--------+
victoitor@bastion:~$ incus project list
+---------------+--------+----------+-----------------+-----------------+----------+---------------+----------------------------------------+---------+
|     NAME      | IMAGES | PROFILES | STORAGE VOLUMES | STORAGE BUCKETS | NETWORKS | NETWORK ZONES |              DESCRIPTION               | USED BY |
+---------------+--------+----------+-----------------+-----------------+----------+---------------+----------------------------------------+---------+
| aux (current) | YES    | YES      | YES             | YES             | NO       | NO            | Containers com baixo recurso           | 6       |
+---------------+--------+----------+-----------------+-----------------+----------+---------------+----------------------------------------+---------+
| compartilhado | YES    | YES      | YES             | YES             | NO       | NO            | Experimentos - máquinas compartilhadas | 7       |
+---------------+--------+----------+-----------------+-----------------+----------+---------------+----------------------------------------+---------+
| dedicado      | YES    | YES      | YES             | YES             | NO       | NO            | Experimentos - máquinas dedicadas      | 4       |
+---------------+--------+----------+-----------------+-----------------+----------+---------------+----------------------------------------+---------+
[...]

PR #618, opened 2024-03-12
Fix config name in ImageProjectFromRecord

presztak:
Fixes: #598


stgraber:
Oh wow, that's a pretty simple issue :)

stgraber:
Nice catch!

PR #624, opened 2024-03-15
incus: add more completions

adamcstephens:
- **incus: add completions for clusters**
- **incus: add completions for cluster groups**
- **incus: add completions for cluster roles**
- **incus: add completions for config devices**
- **incus: add completions for config templates**


adamcstephens:
The link check seems unrelated to this PR.

Issue #600, opened 2024-03-09
Define proper API and Go client function to access API configuration metadata


stgraber:
We've had the `/1.0/metadata/configuration` endpoint for a little while now but there's no matching structs for it in `shared/api` nor is there a Go API client function to access it.

We should:
 - Add a new `MetadataConfiguration` struct in `shared/api` matching the current JSON data
 - Add a new `GetMetadataConfiguration` function in `client` to access the endpoint and return the `api.MetadataConfiguration`

presztak:
@stgraber can I take that one?

stgraber:
For sure!

Can you also rename `documentation.go` to `metadata.go` while you're working on this one?
The file name doesn't make much sense right now.

PR #627, opened 2024-03-16
Define API and Go client function to access API configuration metadata

presztak:
- Adds a new `MetadataConfiguration` struct in `shared/api`
- Adds a new `GetMetadataConfiguration` function in client to access the endpoint
- Renames `incusd/documents.go` to `incusd/metadata.go`

Fixes: #600


presztak:
@stgraber do we need to add API extension as well?

stgraber:
It depends, if no extension was introduced when the server side API was first introduced, then we should add one now.

If there is an existing extension for the current server side API then we should just check for that.

presztak:
> If there is an existing extension for the current server side API then we should just check for that.

yup, we have, and I've added check for it.

stgraber:
Looking great, thanks!

I'll spend a bit of time on the flight back home tweaking the Go struct a bit.
I basically want to see if I can make it contain less hardcoded info about the various types/sections as well as rename some of the Go fields to be clearer about what they are.

I considered also changing the JSON field names but that'd actually be an API breakage so I'll refrain from doing that :)

PR #628, opened 2024-03-17
internal/server/device: Support bus and device number

### LINKED ISSUE #594: bus and device not available as a alternative to vendorid and productid for type usb
# Required information

 * Distribution: Gentoo
   * Kernel version: 6.7.7-gentoo-dist
   * LXC version: 5.0.3
   * Incus version: 0.6

# Issue description

When multiple nearly identical usb devices, is connected to the host, it is not possible to distinguish between them when vendorid and productid is the same. When attaching usb devices, to a instance with vendorid and productid, all matching devices is attached to the instance. If the device type usb, had a bus and device option, it would be possible to distinguish between devices with identical vendorid and productid.

# Steps to reproduce

 1. Plug multiple usb devices, with same vendorid and productid into host
 2. Attach devices to a instance, with device type usb
 3. Enter the instance
 4. List devices in /dev/bus/usb/...
 5. Compare with host devices, to see multiple devices has been attached to instance

monstermunchkin:
If multiple devices with same vendorid and productid are present on the
host, all of these will be attached. There's currently no way of
distinguishing those devices.

This adds support for specifying the bus and device number which are
unique for all USB devices.

Closes #594

### DIFF
--- a/doc/reference/devices_usb.md
+++ b/doc/reference/devices_usb.md
@@ -22,6 +22,8 @@ When a device is passed to the instance, it vanishes from the host.
 
 Key         | Type      | Default           | Description
 :--         | :--       | :--               | :--
+`bus`       | string    | -                 | The bus number of the USB device
+`device`    | string    | -                 | The device number of the USB device
 `gid`       | int       | `0`               | Only for containers: GID of the device owner in the instance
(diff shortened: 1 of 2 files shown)


stgraber:
#629 (and its predecessor, it was closed for a little bit) covers this already with very similar logic.

PR #637, opened 2024-03-21
shared/cliconfig: Ensure client certificate key is 0600

stgraber:


hallyn (changes requested):
This is kind of TOCTTOU-ish.  What about passing the desired mode to copyFile and having it set it before the bits are copied over?

If not, this is definitely an improvement.

stgraber:
@hallyn fixed!
The new logic retains the mode on all files except for the key which is turned into 0600 regardless of pre-existing mode.

hallyn (approved):

PR #639, opened 2024-03-21
lxd-to-incus: Add support for LXD 5.21

stgraber:
This is a bit of a trickier one as LXD 5.21 has added new DB tables and so requires a DB downgrade to the 5.20 schema prior to being able to move the data over to Incus.


adamcstephens:
Should we prompt for confirmation if we're going to drop tables?

stgraber:
@adamcstephens the data is moved away from those tables back to the original ones (it's basically running the schema migration steps in reverse) so that's fine to do. We also always make a full DB backup as part of running the migration.

adamcstephens:
Great, thanks. Supporting 5.21 be nice since it's an LTS and gives a hopefully gives a long timeline to migrate. :)

freeekanayaka (approved):

PR #640, opened 2024-03-21
incusd/instance/qemu: Fix handling of > 64 limits.cpu

stgraber:
Sponsored-by: ActivePort (https://www.activeport.com.au)

### DIFF
--- a/internal/server/instance/drivers/driver_qemu_templates.go
+++ b/internal/server/instance/drivers/driver_qemu_templates.go
@@ -425,6 +425,7 @@ type qemuNumaEntry struct {
 type qemuCPUOpts struct {
 	architecture        string
 	cpuCount            int
+	cpuRequested        int
 	cpuSockets          int
 	cpuCores            int
 	cpuThreads          int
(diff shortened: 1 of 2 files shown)


stgraber:
For whoever reviews this, we were getting the issue because on x86_64, CPUs are always hotplugged unless a specific pin is in place. So in the > 64 case without hotplug on x86_64, the cpuCount was set to 1 which would lead to an upper limit of 64 and therefore be immediately hit when hot plugging the remaining CPUs.

With the update, the hotplug case can now expose a separate cpuRequested with the eventual CPU count, this then sets the max to match that and allows things to behave while still not raising the max above 64 except for those cases.

internal/server/instance/drivers/driver_qemu_templates.go
```diff
@@ -489,6 +490,8 @@ func qemuCPU(opts *qemuCPUOpts, pinning bool) []cfgSection {
 		max := 64
 		if int(cpu.Total) < max {
 			max = int(cpu.Total)
+		} else if opts.cpuRequested > max {
+			max = opts.cpuRequested
```
hallyn, line 494: What if cpuCount is greater than cpuRequested? Do you just want Max to be the max of those three values?

stgraber: cpuCount can't be greated than cpuRequested. At the point where it's set, it's either set to 1 (hotplug case) or set to match cpuRequested.

stgraber: cpuRequested is only set in the hotplug case, in all other cases it's just left unset (0) since it's not otherwise useful.

hallyn (approved):

PR #650, opened 2024-03-23
introduce requirements.nesting for images

adamcstephens:
- **internal/server/instance/lxd: add support for image.requirments.nesting**
- **api: add image_restriction_nesting**
- **doc/images: introduce requirements.nesting**

### DIFF
--- a/internal/server/instance/drivers/driver_lxc.go
+++ b/internal/server/instance/drivers/driver_lxc.go
@@ -2626,6 +2626,21 @@ func (d *lxc) onStart(_ map[string]string) error {
 	return nil
 }
 
+// validateStartup checks any constraints that would prevent start up from succeeding under normal circumstances.
+func (d *lxc) validateStartup(stateful bool, statusCode api.StatusCode) error {
+	err := d.common.validateStartup(stateful, statusCode)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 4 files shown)


internal/server/instance/drivers/driver_lxc.go
```diff
@@ -2626,6 +2626,21 @@ func (d *lxc) onStart(_ map[string]string) error {
...
 
+// validateStartup checks any constraints that would prevent start up from succeeding under normal circumstances.
+func (d *lxc) validateStartup(stateful bool, statusCode api.StatusCode) error {
+	err := d.common.validateStartup(stateful, statusCode)
+	if err != nil {
+		return err
+	}
+
+	// Ensure nesting is turned on for images that require nesting.
+	if util.IsTrue(d.localConfig["image.requirements.nesting"]) && util.IsFalseOrEmpty(d.expandedConfig["security.nesting"]) {
```
adamcstephens, line 2637: Should privileged be allowed here as an alternative?

stgraber: I don't think so as security.privileged is separate. There is some overlap in what it may allow but privileged containers also get a much tighter apparmor profile by default which actually makes them even more likely to need nesting enabled for less common actions.

PR #669, opened 2024-03-26
Change incus-agent install location to writeable location

m2Giles:
By default the install script for incus agent attempts to install to `/lib`. This changes it to write to `/etc` instead. Many systems mount `/usr` read-only and `/lib` is usually a symlink to `/usr/lib`. `/etc/` is normally a writable location on all systems.

### DIFF
--- a/internal/server/instance/drivers/agent-loader/install.sh
+++ b/internal/server/instance/drivers/agent-loader/install.sh
@@ -4,15 +4,34 @@ if [ ! -e "systemd" ] || [ ! -e "incus-agent" ]; then
     exit 1
 fi
 
-if [ ! -e "/lib/systemd/system" ]; then
+# Find target path.
+TARGET=""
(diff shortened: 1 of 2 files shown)


stgraber:
Hello,

It looks like your commit is missing the expected Signed-off-by line.

In general I'd still prefer we attempt to install to /lib/systemd and then fallback to /etc/systemd if we can't.

That's for two reasons:
 - This keeps allowing us to update the units and scripts on systems that come with the agent pre-installed (as those will always be in /lib)
 - This allows the user to decide to mask or fully override the units which they can't do if they're already in /etc

m2Giles:
I apologize for the missing sign-off. I'll add some logic to fallback to /etc.

stgraber:
No worries, thanks for your work!

stgraber (approved):

PR #672, opened 2024-03-26
Add support for creating external interfaces

mikerobski:
This change extends the format for the `bridge.external_interfaces` value when using the `native` driver. The extended format is `<interfaceName>/<parentInterfaceName>/<vlanId>`. When the external interface is added to the list with the extended format, the system will automatically create the interface upon the network's creation and subsequently delete it when the network is terminated. The system verifies that the <interfaceName> does not already exist. If the interface name is in use with a different parent or VLAN ID, or if the creation of the interface is unsuccessful, the system will revert with an error message.

### DIFF
--- a/doc/reference/network_bridge.md
+++ b/doc/reference/network_bridge.md
@@ -110,6 +110,10 @@ Key                                  | Type      | Condition             | Defau
 `tunnel.NAME.ttl`                    | integer   | `vxlan`               | `1`                       | Specific TTL to use for multicast routing topologies
 `user.*`                             | string    | -                     | -                         | User-provided free-form key/value pairs
 
+```{note}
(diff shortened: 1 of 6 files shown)


stgraber:
This seems fine overall. I'll just do the usual bit of rebasing and allocating an API extension string for it.

Though one question, since it looks like your goal is to have an Incus network which is attached to a particular VLAN of an existing interface, did you try macvlan?

```
root@v1:~# incus network create my-network --type=macvlan parent=enp5s0 vlan=1000
Network my-network created
root@v1:~# incus launch images:alpine/edge a1 --network my-network
Launching a1
root@v1:~#
```

mikerobski:
Yes, we considered macvilan, but it has limitations like the ACL support and we need a managed network.

stgraber:
Yeah, the above would have given you a managed network by Incus standard, but indeed ACLs aren't possible on macvlan as the host has no control over the traffic.

PR #691, opened 2024-03-28
Use /1.0/events when under OpenFGA

### LINKED ISSUE #292: Rework event permission handling to be compatible with OpenFGA
Currently the CLI attempts to access /1.0/events to get all possible events.
Unfortunately with OpenFGA enabled, this looks like an attempt to access events of the `default` project which if it's not allowed for the user, causes all operations to fail.

Instead, a distinction should be made between attempting to access the endpoint for a project that the user isn't allowed to access vs attempting to access it with no project passed in.

Then in all cases, the actual permission check should be applied on a per-event basis, filtering the stream to what the user is allowed to see.

There is a current workaround in place which forces those operations to use the `/wait` endpoint rather than rely on the websocket events, but that's not really a good option in the long run.

stgraber:
So I expected that this would need a bunch of work in the event handler and operation handler, but after doing a bit of poking, it looks like we've actually added all the needed filtering already and one can actually call `/1.0/events?all-projects=true` today and only get the events that they should be getting (no logging, only operation and lifecycle for projects that they have access to).

I've done a bunch of testing both from the CLI and UI and can indeed only see the expected events.

Closes #292

### DIFF
--- a/client/incus.go
+++ b/client/incus.go
@@ -366,8 +366,13 @@ func (r *ProtocolIncus) queryStruct(method string, path string, data any, ETag s
 // It sets up an early event listener, performs the query, processes the response, and manages the lifecycle of the event listener.
 func (r *ProtocolIncus) queryOperation(method string, path string, data any, ETag string) (Operation, string, error) {
 	// Attempt to setup an early event listener
+	skipListener := false
 	listener, err := r.GetEvents()
 	if err != nil {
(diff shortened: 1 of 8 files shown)


stgraber:
Let's see if the tests catch anything I didn't consider...

stgraber:
Ah right, I tested project-wide permissions, object-specific ones are the problem, will look into that soon.

PR #693, opened 2024-03-28
incusd: Fix use of JWT TLS authentication when OIDC is also enabled

### LINKED ISSUE #689: Fix use of JWT TLS authentication when OIDC is also enabled
Currently when OIDC is enabled, it will return a top-level error when a user provides a JWT that's derived from a TLS certificate rather than signed by the OIDC provider.

Either the order of operation should be changed to first consider TLS JWT or the OIDC error shouldn't prevent further processing.

mikerobski:
Closes #689


mikerobski:
Continuing the processing when the OIDC authentication fails causes tests to fail.
The reordering of the operations to first check for JWT token works properly.

Issue #590, opened 2024-03-08
Not possible to move instance between projects while also defining a target cluster member


victoitor:
# Issue description

When running `incus move`, it is not allowed to use both `--target-project` together with `--target`.

# Steps to reproduce

```
victoitor@bastion:~$ incus move test --target-project aux --target @compartilhado
Error: The --target-project flag can't be used with --target
```

This was initially discussed in [this forum post](https://discuss.linuxcontainers.org/t/moving-instances-between-projects-with-cluster-groups/).

stgraber:
The easy label here is based on the assumption that this is just a CLI or basic API issue.
This may be hiding more complex server-side issues with cross-project copies/moves.

presztak:
@stgraber as we've discussed I can grab this as next one if you don't mind.

stgraber:
Sure!

presztak:
Hey @stgraber , quick update from my side. it seems that removal of https://github.com/lxc/incus/blob/ce15a0a8c41d96466ff6222f8c2d52051fcfba85/cmd/incus/move.go#L165-L167 is not enough to solve this :) 
For moving instances between cluster members we are using https://github.com/lxc/incus/blob/ce15a0a8c41d96466ff6222f8c2d52051fcfba85/cmd/incusd/instance_post.go#L365 and `migrateInstance` doesn't support moving instance across projects but I already have some changes which seems to work. I'll open tomorrow draft PR because still I've some doubts.

stgraber:
Sounds great, thanks for the update!

stgraber:
Confirmed that #727 fixes this one.
```
root@server01:~# incus list
+------+---------+------+------+-----------+-----------+----------+
| NAME |  STATE  | IPV4 | IPV6 |   TYPE    | SNAPSHOTS | LOCATION |
+------+---------+------+------+-----------+-----------+----------+
| d4   | STOPPED |      |      | CONTAINER | 0         | server01 |
+------+---------+------+------+-----------+-----------+----------+
| d5   | STOPPED |      |      | CONTAINER | 0         | server03 |
+------+---------+------+------+-----------+-----------+----------+
| d6   | STOPPED |      |      | CONTAINER | 0         | server03 |
+------+---------+------+------+-----------+-----------+----------+
root@server01:~# ./incus move d4 --target-project default --target server02
root@server01:~# incus list --all-projects name=d4
+---------+------+---------+------+------+-----------+-----------+----------+
[...]

PR #695, opened 2024-03-28
Allow to use `--target` flag all together with `--target-project`

presztak:
Allow to use `--target` flag all together with `--target-project` when moving instances between cluster members.

Fixes: #590


presztak:
@stgraber actually I'm not sure about part in `instances_post.go`. I've implemented `UpdateProject` function for instances because in first place we need to have possibility to load instance (so we need to use current project) but then this instance is passed to migrationSinkArgs so I'm updating project before that because then in some places `inst.Project().Name` is called and in that point of time we should have already access to new project name.

PR #696, opened 2024-03-29
Keep track of cluster groups for evacuations

stgraber:


stgraber:
The fact that we weren't running cluster requests in notify mode is a bit odd considering the number of places in this code path which was checking whether we were handling a notification (to avoid running config checks twice).

In my local tests, the logic seems to be behaving just fine when in notify mode, so hopefully it was in fact a long standing mistake and not something that we had to do on purpose and failed to document.

hallyn (approved):

PR #699, opened 2024-03-29
incusd/instance/qemu: Make rootfs.img be 0600

stgraber:


internal/server/instance/drivers/driver_qemu.go
```diff
@@ -6278,13 +6278,13 @@ func (d *qemu) Export(w io.Writer, properties map[string]string, expiration time
 	defer revert.Fail()
 
 	// Check for Direct I/O support.
-	from, err := os.OpenFile(mountInfo.DiskPath, unix.O_DIRECT|unix.O_RDONLY, 0)
+	from, err := os.OpenFile(mountInfo.DiskPath, unix.O_DIRECT|unix.O_RDONLY, 0600)
```
brauner, line 6281: The mode will be useless here though, no?

stgraber: Yeah, removed that part of the change now so it doesn't look too confusing.

Issue #702, opened 2024-03-29
Add USB support to `incus info --resources`


stgraber:
Currently `incus info --resources` doesn't show USB device at all.

It would be good to add a concise listing of USB devices based on the data already available in the resources API (`/1.0/resources`).

This list should at least include:
 - Device vendor (name and id)
 - Device product (name and id)
 - Bus and device address
 - Serial number (if present)

So looking at something not too dissimilar to the `lsusb` output.

RaviTriv:
Hey, I'm interested in working on this issue. Would be awesome if I could be assigned this. Thank you!

stgraber:
Done!

PR #723, opened 2024-04-02
Adding USB support to `incus info --resources`

RaviTriv:
This PR is in reference to #702 . It adds Usb information to cmd output of resources.


stgraber:
Reviewed and fixed a few minor things:
 - Split code change and translations into two commits
 - Added bug reference to code commit and re-titled it
 - Reworded things slightly so header is `USB devices`, single device is `USB device` and individual entries are just `Device X`.

stgraber (approved):

PR #747, opened 2024-04-12
Stop networks during cluster evacuations

### LINKED ISSUE #340: Disable OVN gateway chassis when evacuated
As a machine that's evacuated is likely to be going down shortly thereafter, it would make sense for the evacuated state to be tied to either a disabled OVN gateway chassis state or at least a much lower chassis priority.

Abhiram824:
Closes #340

### DIFF
--- a/cmd/incusd/api_cluster.go
+++ b/cmd/incusd/api_cluster.go
@@ -3282,6 +3282,9 @@ func evacuateClusterMember(s *state.State, gateway *cluster.Gateway, r *http.Req
 			return err
 		}
 
+		// Stop networks after evacuation.
+		networkShutdown(s)
+
 		revert.Success()
 		return nil
 	}
@@ -3426,6 +3429,12 @@ func restoreClusterMember(d *Daemon, r *http.Request) response.Response {
 		var sourceNode db.NodeInfo
 
 		metadata := make(map[string]any)
+
+		// Restart the networks.
+		err = networkStartup(d.State())
+		if err != nil {
+			return err
+		}
 
 		// Restart the local instances.
 		for _, inst := range localInstances {
--- a/cmd/incusd/daemon.go
+++ b/cmd/incusd/daemon.go
@@ -1462,10 +1462,13 @@ func (d *Daemon) init() error {
 	}
 
 	// Setup the networks.
-	logger.Infof("Initializing networks")
-	err = networkStartup(d.State())
-	if err != nil {
-		return err
+	if !d.db.Cluster.LocalNodeIsEvacuated() {
+		logger.Infof("Initializing networks")
+
+		err = networkStartup(d.State())
+		if err != nil {
+			return err
+		}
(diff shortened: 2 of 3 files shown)


cmd/incusd/api_cluster.go
```diff
@@ -3265,6 +3266,12 @@ func evacuateClusterMember(s *state.State, gateway *cluster.Gateway, r *http.Req
 			_ = evacuateClusterSetState(s, nodeName, db.ClusterMemberStateCreated)
 		})
 
+		 // TODO stop all networks here
```
stgraber, line 3269: That should be done later, after `evacuateInstances` is done otherwise the instances are going to run into issues as the network disappears underneath them.

cmd/incusd/api_cluster.go
```diff
@@ -3583,6 +3590,14 @@ func restoreClusterMember(d *Daemon, r *http.Request) response.Response {
 			}
 		}
 
+		logger.Infof("Initializing networks")
```
stgraber, line 3593: And here the reverse is true, networks need to be brought back up prior to any instance being started back.

cmd/incusd/api_cluster.go
```diff
@@ -3265,6 +3266,12 @@ func evacuateClusterMember(s *state.State, gateway *cluster.Gateway, r *http.Req
 			_ = evacuateClusterSetState(s, nodeName, db.ClusterMemberStateCreated)
 		})
 
+		 // TODO stop all networks here
+		err = networkStopOVN(s)
```
stgraber, line 3270: So the issue mentioned OVN as that's the biggest issue, but the better approach here is to just stop all networks.

So calling `networkShutdown` to stop all networks.

stgraber:
On top of the few comments left in-line, note that your commit messages are missing the `Signed-off-by` line that we required for all commits (see CONTRIBUTING.md).

You can usually get that line added to the current commit with `git commit --amend -s`, assuming a properly configured git client.

cmd/incusd/api_cluster.go
Abhiram824: Should the networks be started before or after the migration/instance reloading?

stgraber: Before

Abhiram824:
Hello, thanks for the feedback! Made updates to the PR

stgraber:
Okay, got this one ready for Github testing and merging, I've done:
 - Merged all the commits together
 - Applied the result of `gofmt -s -w ./`
 - Removed a TODO comment that was leftover
 - Removed duplicate `networkStartup` call in `restoreClusterMember` (only needed at the start)
 - Reworded the comments a bit.
 - Reworded the commit message a bit.
 - Ran `make static-analysis` and cleaned up some new-line issues in `daemon.go`

stgraber (approved):

PR #750, opened 2024-04-13
incusd/network/ovn: rewrote LogicalRouterAdd to use o.client

DhruvNistala:


stgraber:
New function looks good, so now you need a commit to remove the old `LogicalRouterAdd` and then another one to update everywhere that currently calls it to use `CreateLogicalRouter` instead.

dhruvvnistala:
Hi, I've updated the LogicalRouterAdd usage with CreateLogicalRouter accordingly.

stgraber:
I've done a quick rebase on this one to get rid of the merge commits, I also added the missing description comment on the new function and ran `gofmt -s -w ./` to fix the bad indentation on the new function. After a quick rewording of the commit message, I pushed the updated code to the branch and it should now be good to go once the tests all run.

Issue #752, opened 2024-04-14
Add support for ZFS mirror/striping


stgraber:
We should expand the syntax of `source` on ZFS storage pools to support:
 - striping (multiple zdevs)
 - mirroring (mirrored zdevs)
 - RAID5 (raidz1)
 - RAID6 (raidz2)

Probably something like:
 - `source=/dev/sda,/dev/sdb`
 - `source=mirror=/dev/sda,/dev/sdb`
 - `source=raidz1=/dev/sda,/dev/sdb,/dev/sdc,/dev/sdd,/dev/sde`
 - `source=raidz2=/dev/sda,/dev/sdb,/dev/sdc,/dev/sdd,/dev/sde`

For now we'll stay away from doing multiple mirrors or mix of raidz levels and the like, but the syntax above should hopefully allow for some of that as well as eventual addition of things like arc and log devices.

ps-gill:
Is this one still available? I'm willing to work on it.

stgraber:
Sure, got it assigned to you!

cyphar:
At the moment, if you use an existing ZFS pool name as the storage source you can set it up however you like. Is that not intended?

I'm not sure it would be a good idea to put the initial pool configuration in the incus config format. Once you've configured a ZFS pool, ZFS doesn't care about the names assigned to individual drives (instead, `zpool import` looks at the unique ID stored on each drive in the pool) and the layout is all stored on the drives themselves. Not to mention the pool layout can change afterwards through regular ZFS administration. Almost all ZFS users would expect the pool name to be the thing incus cares about, and IMHO would expect ZFS to be configured directly.

(Also, being able to refer to the pool purely by name means that if you want to migrate the pool you can use `zfs send | zfs recv` to migrate the pool, rename the new pool and AFAIK incus will use the new pool without issues.)

Then again, maybe I'm misunderstanding the aim of this feature.

stgraber:
Incus always rewrites the source property to the pool name after creating it so that won't change with this. 

This will however allow for the creation of more complex zpool through the API alone, making it possible to automate production cluster deployments without having to separately handle this stuff through ssh.

PR #754, opened 2024-04-16
Add support for ZFS mirror/striping

ps-gill:
This pull request resolves issue https://github.com/lxc/incus/issues/752

Summary:
- Updating ZFS driver to support multiple block devices and `vdev` types (`mirror`, `raidz1` and `raidz2`). Default `vdev` type, `stripe`, is already supported.
- Updating ZFS driver docs on how to use the new config.


internal/server/storage/drivers/driver_zfs.go
```diff
@@ -713,3 +734,14 @@ func (d *zfs) patchDropBlockVolumeFilesystemExtension() error {
...
 }
+
+func (d *zfs) parseSource() (string, []string) {
+	sourceParts := strings.Split(d.config["source"], "=")
+	poolType := ""
+	devices := sourceParts[0]
+	if len(sourceParts) > 1 {
+		poolType = sourceParts[0]
+		devices = sourceParts[1]
+	}
```
stgraber, line 745: Needs a line break here

internal/server/storage/drivers/utils.go
```diff
@@ -880,3 +880,21 @@ func wipeBlockHeaders(path string) error {
 func IsContentBlock(contentType ContentType) bool {
 	return contentType == ContentTypeBlock || contentType == ContentTypeISO
 }
+
+func sliceAll[T any](slice []T, predicate func(T) bool) bool {
+	for _, element := range slice {
+		if !predicate(element) {
+			return false
+		}
+	}
```
stgraber, line 889: Needs a line break here.

internal/server/storage/drivers/utils.go
```diff
@@ -880,3 +880,21 @@ func wipeBlockHeaders(path string) error {
...
+	return true
+}
+
+func sliceAny[T any](slice []T, predicate func(T) bool) bool {
+	for _, element := range slice {
+		if predicate(element) {
+			return true
+		}
+	}
```
stgraber, line 898: Needs a line break here.

internal/server/storage/drivers/driver_zfs.go
```diff
@@ -25,6 +25,12 @@ import (
 	"github.com/lxc/incus/v6/shared/validate"
 )
 
+var zfsSupportedPoolTypes = []string{
```
stgraber, line 28: Should probably be `zfsSupportedDevTypes` to be more accurate

internal/server/storage/drivers/driver_zfs.go
```diff
@@ -210,7 +217,7 @@ func (d *zfs) FillConfig() error {
 
 			d.config["size"] = fmt.Sprintf("%dGiB", defaultSize)
 		}
-	} else if filepath.IsAbs(d.config["source"]) {
+	} else if sliceAll(devices, func(device string) bool { return filepath.IsAbs(device) }) {
```
stgraber, line 220: That logic seems incorrect to me as someone feeding "blah,/dev/foo" will be treated as having provided a valid source and will have their pool use the `blah` zpool, completely ignoring the invalid /dev/foo having been passed in.

Instead I think we need to flip around the conditions, basically checking for `if len(devices) == 1 && !filepath.IsAbs(devices[0])` to go with the existing zpool.

And then have the `else` handle the case where devices are passed in, in which case we should ensure that all of them are absolute paths or return an error back to the user.

internal/server/storage/drivers/driver_zfs.go
stgraber, line 28: Probably also add `stripe` as a type and make that the default in parseSource?

That should be cleaner to handle in the code than having to check for an empty value every time.

internal/server/storage/drivers/driver_zfs.go
```diff
@@ -242,8 +249,12 @@ func (d *zfs) Create() error {
 		return err
 	}
 
+	poolType, devices := d.parseSource()
+	if poolType != "" && !slices.Contains(zfsSupportedPoolTypes, poolType) {
```
stgraber, line 253: With above change, we should be able to drop the `poolType != ""` part of that as default would now be `stripe`.

internal/server/storage/drivers/driver_zfs.go
```diff
@@ -242,8 +249,12 @@ func (d *zfs) Create() error {
...
+		return fmt.Errorf("Unsupported pool type: %q. Supported pool types: %s", poolType, zfsSupportedPoolTypes)
```
stgraber, line 254: Should be `Unsupported ZFS vdev type %q. Supported types are %v`

internal/server/storage/drivers/driver_zfs.go
```diff
@@ -242,8 +249,12 @@ func (d *zfs) Create() error {
...
+	}
 	loopPath := loopFilePath(d.name)
-	if d.config["source"] == "" || d.config["source"] == loopPath {
+	if len(devices) == 1 && devices[0] == loopPath {
```
stgraber, line 257: Those two conditions aren't the same, it should likely be:
```
if len(devices) == 0 || (len(devices) == 1 && devices[0] == loopPath)
```

internal/server/storage/drivers/driver_zfs.go
```diff
@@ -261,7 +272,12 @@ func (d *zfs) Create() error {
 		}
 
 		// Create the zpool.
-		_, err = subprocess.RunCommand("zpool", "create", "-m", "none", "-O", "compression=on", d.config["zfs.pool_name"], loopPath)
+		createArgs := []string{"create", "-m", "none", "-O", "compression=on", d.config["zfs.pool_name"]}
+		if poolType != "" {
+			createArgs = append(createArgs, poolType)
+		}
+		createArgs = append(createArgs, loopPath)
+		_, err = subprocess.RunCommand("zpool", createArgs...)
```
stgraber, line 280: This change doesn't make sense, we're dealing with the loop path case which is when the user doesn't provide  devices, so we're guaranteed to only deal with a single device which is `loopPath`, no need to change anything here.

internal/server/storage/drivers/driver_zfs.go
```diff
@@ -273,9 +289,9 @@ func (d *zfs) Create() error {
 				return err
 			}
 		}
-	} else if filepath.IsAbs(d.config["source"]) {
+	} else if sliceAll(devices, func(device string) bool { return filepath.IsAbs(device) }) {
```
stgraber, line 292: Same issue as above, we need the `else if` and `else` statements flipped as described above to avoid incorrectly treating bad user input as an existing zpool name.

internal/server/storage/drivers/driver_zfs.go
```diff
@@ -284,26 +300,30 @@ func (d *zfs) Create() error {
...
-			if err != nil {
-				return err
-			}
+			createArgs = []string{"create", "-m", "none", "-O", "compression=on", d.config["zfs.pool_name"]}
+		}
+
+		// Create the zpool.
+		if poolType != "" {
+			createArgs = append(createArgs, poolType)
+		}
```
stgraber, line 322: Line break here

internal/server/storage/drivers/driver_zfs.go
```diff
@@ -318,7 +338,7 @@ func (d *zfs) Create() error {
 		d.config["source"] = d.config["zfs.pool_name"]
 	} else {
 		// Validate pool_name.
-		if d.config["zfs.pool_name"] != d.config["source"] {
+		if len(devices) != 1 || d.config["zfs.pool_name"] != devices[0] {
```
stgraber, line 341: Ah, so that would have caught some of the issue mentioned above, though I still think flipping the `else if` and `else` sections will make this cleaner.

stgraber:
General structure looks good, just a few styling nits and some smaller bits that need tweaking/fixing.

You should also add an extra commit, `api: storage_zfs_vdev` containing changes to both `doc/api-extensions.md` and `internal/version/api.go` introducing an API extension of that same name so API clients can detect that Incus supports the new syntax.

internal/server/storage/drivers/driver_zfs.go
ps-gill: At the start of this function (`Create`), `FillConfig` is called, which ensures that `devices` isn't empty. Could we rely on that behavior or check within `Create` as well?

internal/server/storage/drivers/driver_zfs.go
ps-gill: Added a vdev type check at the top. Would that suffice?

internal/server/storage/drivers/driver_zfs.go
ps-gill: Would the vdev type check at the start suffice?

internal/server/storage/drivers/driver_zfs.go
ps-gill: Done.

internal/server/storage/drivers/driver_zfs.go
ps-gill: Done

internal/server/storage/drivers/utils.go
ps-gill: Done

internal/server/storage/drivers/driver_zfs.go
ps-gill: Is this still valid, given the other changes?

internal/server/storage/drivers/driver_zfs.go
stgraber: I don't see how the change above makes this any different.

If I pass `source=blah,/dev/foo`, parseSource() will give you a default type (stripe) and two devices.

The check here will fail as `blah` isn't an absolute path which then makes us hit the `else` case that will then use `blah` as the pool, completely ignoring the fact that the user passed two devices and that it's invalid.

internal/server/storage/drivers/driver_zfs.go
stgraber: Ah, so the old logic was wrong and `source` couldn't actually be empty there?

internal/server/storage/drivers/driver_zfs.go
```diff
@@ -261,7 +280,14 @@ func (d *zfs) Create() error {
...
 		// Create the zpool.
-		_, err = subprocess.RunCommand("zpool", "create", "-m", "none", "-O", "compression=on", d.config["zfs.pool_name"], loopPath)
+		createArgs := []string{"create", "-m", "none", "-O", "compression=on", d.config["zfs.pool_name"]}
+		// "zpool create" doesn't have an explicit type for "stripe" vdev type
+		if vdevType != zfsDefaultVdevType {
+			createArgs = append(createArgs, vdevType)
+		}
+
+		createArgs = append(createArgs, loopPath)
+		_, err = subprocess.RunCommand("zpool", createArgs...)
```
stgraber, line 290: Hmm, I thought I pointed that one out already :)

This is the `loopPath` case, this cannot have more than a single device which is the loopPath so there is no need to change anything to the current logic.

internal/server/storage/drivers/driver_zfs.go
stgraber: Unless I missed something, no, the check doesn't solve the problem.

internal/server/storage/drivers/driver_zfs.go
stgraber: Given that I'm not convinced the other change solves the problem, I think this is still valid

internal/server/storage/drivers/driver_zfs.go
ps-gill: I've rearranged the blocks to align with feedback.

internal/server/storage/drivers/driver_zfs.go
ps-gill: Yes, when `source` is empty, `FillConfig` sets it to `loopPath`.

internal/server/storage/drivers/driver_zfs.go
ps-gill: I've rearranged the blocks to align with feedback.

internal/server/storage/drivers/driver_zfs.go
ps-gill: I've rearranged the blocks to align with feedback.

internal/server/storage/drivers/driver_zfs.go
ps-gill: I've updated `FillConfig` to ensure that if there are multiple devices, all of them have to be existing block devices.

stgraber: Ah good, that makes sense

stgraber:
Just did a quick rebase, merged two commits together and re-titled the commits as well as ran `make static-analysis` and fixed the 2-3 things that picked up.

Should be all good now, just waiting for Github to run all the tests.

Thanks!

ps-gill:
The documentation test is failing due to `vdev` being lowercase and `config` (short for `configuration`). If you prefer, I can get those patched.

stgraber:
Ah yeah, you probably need to escape vdev as it's not a dictionary word and config indeed needs to become configuration. 

Please make those two edits directly in the doc commit then when things pass on Github I can merge it. 

I'm on my phone for the rest of the day here so can't directly edit commits.

ps-gill:
> Ah yeah, you probably need to escape vdev as it's not a dictionary word and config indeed needs to become configuration.
> 
> Please make those two edits directly in the doc commit then when things pass on Github I can merge it.
> 
> I'm on my phone for the rest of the day here so can't directly edit commits.

It is done.

stgraber:
The api-extensions.md fix needed to go in the api: commit, not the doc: commit, sorry :)

ps-gill:
> The api-extensions.md fix needed to go in the api: commit, not the doc: commit, sorry :)

Just fixed it.

Sorry, I should have done it that way the first time. A workflow where commit message and contents are tightly integrated with CI is new to me.

Issue #741, opened 2024-04-08
Add ability to read YAML from stdin to all `create` commands


stgraber:
Currently only `incus create` and `incus launch` will read YAML data from stdin to create a new instance (reading an InstancePut which gets embeded in the InstancePost struct).

The same logic should be supported for all other `create` commands:

- [x] incus cluster group create
- [x] incus config template create
- [x] incus network acl create
- [x] incus network forward create
- [x] incus network integration create
- [x] incus network load-balancer create
- [x] incus network peer create
- [x] incus network zone create
- [x] incus network zone record create
- [x] incus profile create
- [x] incus project create
- [x] incus snapshot create
- [x] incus storage bucket create
- [x] incus storage bucket key create
- [x] incus storage create
- [x] incus storage volume create
- [ ] incus storage volume snapshot create

That list was based from a quick grep + some hand filtering, so there may be some entries where reading from stdin doesn't make sense.

awalvie:
Can I be assigned this issue please!

stgraber:
sure, done!

awalvie:
Where can I find valid example `config.yaml` files as reference for these?

rvveber:
@awalvie 
Try to take a look at the `--expanded` flag that is used with `incus config show <instance> --expanded`

If i read the docs correctly, it displays all information there is about the entity.
It needs to get its information from somewhere, hopefully from an object.

This object then could be the source of truth to base all possible config keys from.

From my testing, it looks like, it is not implemented in every show command.
So, implementing it for every show command, would be a thing to do in the future anyways. 
@stgraber Correct my if i'm wrong.

awalvie:
Maybe we could add minimal configs for each resource to the docs? In much the same way the kubernetes docs do? For example: 

- https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#creating-a-deployment
- https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#example

stgraber:
> Where can I find valid example `config.yaml` files as reference for these?

In general, the same data as you get from the matching `show` command.

stgraber:
> Maybe we could add minimal example configs for each resource to the docs? In much the same way the kubernetes docs do? For example:

https://linuxcontainers.org/incus/docs/main/rest-api-spec/

This has example values for all the objects, it's not the best looking thing ever but it doesn't need to be manually updated.

awalvie:
[...]

PR #756, opened 2024-04-16
incus/profile: Add support for creating profile from yaml

awalvie:
Ref: #741


stgraber:
I updated the commit title a bit and aligned the commit author with the Signed-off-by line so the test should clear now. I also added the `make i18n` result to a `i18n: Update translation templates` commit which should then clear the rest of the failures.

Assuming tests behave now, I'll merge this and you can repeat much of the same with the remaining commands, thanks!

awalvie:
> I updated the commit title a bit and aligned the commit author with the Signed-off-by line so the test should clear now. I also added the make i18n result to a i18n: Update translation templates commit which should then clear the rest of the failures.

Thank you! I'll take care of them in the follow-up PRs.

PR #767, opened 2024-04-17
Add yaml support for snapshot create

awalvie:
Part of #741

### DIFF
--- a/cmd/incus/snapshot.go
+++ b/cmd/incus/snapshot.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bufio"
 	"fmt"
+	"io"
 	"os"
 	"slices"
 	"strings"
@@ -16,6 +17,7 @@ import (
 	"github.com/lxc/incus/v6/internal/i18n"
 	"github.com/lxc/incus/v6/internal/instance"
 	"github.com/lxc/incus/v6/shared/api"
+	"github.com/lxc/incus/v6/shared/termios"
 )
 
 type cmdSnapshot struct {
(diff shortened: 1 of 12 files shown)


awalvie:
The failing test seems to be unrelated the the PR?

cmd/incus/snapshot.go
```diff
@@ -11,11 +12,12 @@ import (
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v2"
 
-	"github.com/lxc/incus/v6/client"
+	incus "github.com/lxc/incus/v6/client"
```
stgraber, line 15: That change shouldn't be there

cmd/incus/snapshot.go
```diff
@@ -157,6 +175,13 @@ func (c *cmdSnapshotCreate) Run(cmd *cobra.Command, args []string) error {
 
 	if c.flagNoExpiry {
 		req.ExpiresAt = &time.Time{}
+	} else if !stdinData.ExpiresAt.IsZero() {
+		if stdinData.ExpiresAt.Before(time.Now()) {
+			return fmt.Errorf(i18n.G("Expiry date cannot be in the past"))
+		} else {
+			req.ExpiresAt = &stdinData.ExpiresAt
+			fmt.Printf(req.ExpiresAt.String())
+		}
```
stgraber, line 184: A bit confused about why we have that stuff here, we shouldn't perform input validation on the client side, it's really for the server to handle instead.

If the user provides us a YAML input with a data that's past expiry, it will cause the server to expire the snapshot on the next expiry run, but that's fine, that may actually be useful for testing.

cmd/incus/snapshot.go
awalvie: ah, my bad, I think `gofmt` does that automatically. Is there a rule you'd suggest to configure it to not do this?

cmd/incus/snapshot.go
awalvie: Gotcha, I'll make a fixup!

awalvie:
I'll take care of the conflicts if the changes look good!

cmd/incus/snapshot.go
stgraber: gofmt doesn't do that here but goimports has been known to do that sometimes...

awalvie:
Ready for next round of reviews!

PR #775, opened 2024-04-20
Add yaml support for storage create

awalvie:
Part of #741


stgraber:
I do a git rebase against the main branch and when I hit the commit I either manually solve it or in the case of auto-generated commits like i18n, I just discard its content and regenerate it.

Issue #703, opened 2024-03-29
Add CPU architecture to `incus info --resources`


stgraber:
A very trivial one-line addition, but it'd be pretty useful to have `incus info --resources` show the CPU architecture. That's particularly useful when considering various servers in a cluster.

The `Architecture` flag is already present in the API data, so it's really just a matter of exposing it in the CLI.

sophiezhangg:
Hello!

My partners and I are currently in the Virtualization class at the University of Texas at Austin - we are very interested in trying to implement this issue just to gain some familiarity with this code base and are wondering if you are willing to assign us this?

Thank you so much!

stgraber:
Done!

sophiezhangg:
Hello!

Sorry to bother but we are currently running into some issues with gopath and goroot? more specifically, from the make errors, it seems like the pkgs specified by the makefile aren't able to be found in the goroot (path to my go folder) i specified like this below:

![image](https://github.com/lxc/incus/assets/55890471/efb9c720-840f-4f7d-81f4-d8a9c2e149e6)

We are wondering why this would occur and if there are any fixes? 

thank you so much for your help!

stgraber:
Sounds like you're using an old Go version?

What does `go version` get you? The minimum we support building with is 1.21

sophiezhangg:
it is 1.18.1 T_T 

i see now! we will try this - thank u for the help!

sophiezhangg:
Hello!

I tried running incus after successfully building the repo starting with the cmd below:
sudo -E PATH=${PATH} LD_LIBRARY_PATH=${LD_LIBRARY_PATH} $(go env GOPATH)/bin/incusd --group sudo

and switched out the $(go env GOPATH) with the actual path to go on my laptop - but then this cmd hangs after executing in terminal along with other cmds in the "Get Started with Incus" section like "incus admin init" or "incus list"

I am wondering which step, if any, did I miss or is there something I have done incorrectly? I have tried to go through the docs to see but from what I can tell, I think I followed everything after build in the link: https://linuxcontainers.org/incus/docs/main/installing/#from-source-install

Thank you so much and i really appreciate any suggestion / advice!

stgraber:
It's normal for the `incusd` command to hang, it should print a few things, especially if you pass it `--verbose` or `--debug` but it will keep running so long as Incus is running, so you'll want to keep that one in a terminal.

Then once that one is running, commands like `incus list` should work from another terminal.

[...]

PR #776, opened 2024-04-20
incus/info displays cpu arch when using resources flag

sophiezhangg:
Added a new line for CPU architecture when the client runs incus info --resources

the previous version of this file already has the architecture flag displayed in the terminal, but we just moved it to a separate line for readability

Closes: #703


stgraber:
Looks good. As this changes a translated string, you also need to run `make i18n` and commit the result of that command as `i18n: Update translation templates`

You also need your commits to have a Signed-off-by line.

I've done those two things for you now as well as fixed a tiny nit picked up by running `make static-analysis`, basically just running `gofmt -s -w ./` to fix some formatting issues.

I pushed the resulting changes to your branch and will merge once Github is happy.

Issue #450, opened 2024-01-29
Switch to using minio's `mc` client


stgraber:
Minio has moved its madmin-go to AGPLv3 some time ago, making it unsuitable for use within Incus.

We've so far been able to avoid this issue by sticking to an older version of the package which was still under Apache 2.0. Unfortunately recent Minio has now deprecated the service API supported by that older version of the admin client.

As a result, we need to either switch to another Go package to handle admin functions or need to switch to the `mc` command line tool instead.

dontlaugh:
Was [SeaweedFS](https://github.com/seaweedfs/seaweedfs?tab=readme-ov-file#example-using-seaweed-object-store) ever evaluated? It looks like a more complex project, but it's Apache licensed.

stgraber:
I believe it was back when we were looking at adding S3 storage support and indeed felt like even more of a beast than minio.

I'm hoping that replacing the few admin API calls we have today with command line calls ends up being pretty simple. If we have to somehow completely switch away from minio, I suspect it will be far easier for us to just natively provide our own S3 API than to rely on a 3rd party solution to do that.

presztak:
Hey @stgraber , if you don't mind I can start working on that one.

stgraber:
That'd be great!

knutov:
just wonder - how name conflict with midnight commander (`mc`) is solved here?

stgraber:
The code supports both `mc` and `mcli` as names for the minio client.

It checks the `--version` output for the expected `mc version XYZ` pattern from a valid MinIO client, if it doesn't get it, it will fallback to looking for `mcli` instead.

PR #777, opened 2024-04-21
Switch to using minio's `mc` client.

presztak:
Fixes: #450


presztak:
@stgraber, I'm not sure about  `runClientCommand` method. Very similar implementation we have in `subprocess` package but I've needed to modify this a little bit to support change of working directory. `mc admin cluster iam export` creates file in current working directory so I think it would be better to have possibility to change CWD for any new file.

internal/server/storage/s3/miniod/admin_client.go
```diff
@@ -0,0 +1,250 @@
...
+	return filepath.Join(c.varPath, "mc", "policy")
+}
+
+// runClientCommand runs 'mc' command.
+func (c *AdminClient) runClientCommand(ctx context.Context, workDir string, extraArgs ...string) (string, error) {
+	args := []string{"-C", c.configDir()}
+
+	args = append(args, extraArgs...)
+
+	cmd := exec.CommandContext(ctx, "mc", args...)
```
gibmat, line 54: It's also common for distros to ship the MinIO Client as `mcli` -- would probably be good to add that as a fallback test if `mc` doesn't exist.

internal/server/storage/s3/miniod/admin_client.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	var stdout bytes.Buffer
+	var stderr bytes.Buffer
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	err := cmd.Run()
+	if err != nil {
+		return stdout.String(), subprocess.NewRunError("mc", args, err, &stdout, &stderr)
```
gibmat, line 67: Since there could be people migrating from a version of Incus where the MinIO client library was used, it would be good to differentiate between no `mc` executable being found (ie, "Incus now depends on the MinIO Client; on most distributions you can install the `minio-client` package to fix this.") vs an error running the command itself.

stgraber:
Very good feedback from @gibmat, I'll let you sort those out!

I'm on vacation this week so I don't get to review things quite as quickly than usual but still usually get to process a couple PRs a day. I may get back to that one this evening or otherwise tomorrow morning.

Thanks!

internal/server/storage/s3/miniod/admin_client.go
presztak: Fixed. First we are trying `mc` then `mcli`

internal/server/storage/s3/miniod/admin_client.go
presztak: Fixed. We are returning error whenever `mc` nor `mcli` is not present on the system.

internal/server/storage/s3/miniod/admin_client.go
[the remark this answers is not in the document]
presztak: @gibmat @stgraber any good way to detect difference between MinIO client and Midnight Commander?

stgraber:
```
stgraber@castiana:~/data/code/lxc/incus (735)$ ~/minio-binaries/mc --version
mc version RELEASE.2024-04-18T16-45-29Z (commit-id=3ec030eb26e460ec62655487ba116a05eda3ac99)
Runtime: go1.21.9 linux/amd64
Copyright (c) 2015-2024 MinIO, Inc.
License GNU AGPLv3 <https://www.gnu.org/licenses/agpl-3.0.html>

stgraber@castiana:~/data/code/lxc/incus (735)$ mc --version
GNU Midnight Commander 4.8.29
Built with GLib 2.74.5
Built with S-Lang 2.3.3 with terminfo database
Built with libssh2 1.10.0
With builtin Editor and Aspell support
With subshell support as default
With support for background operations
With mouse support on xterm and Linux console
With support for X11 events
With internationalization support
With multiple codepages support
With ext2fs attributes support
Virtual File Systems:
 cpiofs, tarfs, sfs, extfs, ext2undelfs, ftpfs, sftpfs, fish
Data types:
 char: 8; int: 32; long: 64; void *: 64; size_t: 64; off_t: 64;
stgraber@castiana:~/data/code/lxc/incus (735)$ 
```

So probably look at the first line of output, if you get `mc version` that seems like it's going to be MinIO, if you get anything else, let's assume it's not.

internal/server/storage/s3/miniod/admin_client.go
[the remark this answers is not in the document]
presztak: Fixed.

stgraber (approved):

stgraber:
LGTM, thanks!

PR #785, opened 2024-04-24
incus/doc/installing.md: Add Docker information

cmspam:
As discussed here:
https://github.com/zabbly/incus/issues/41

As I maintain a Dockerfile and docker images for incus, I would like to add that information to make it more accessible to people who don't have better installation options on their distributions.

### DIFF
--- a/doc/.wordlist.txt
+++ b/doc/.wordlist.txt
@@ -199,6 +199,7 @@ pre
 preseed
 proxied
 proxying
+Podman
 PTS
 qdisc
 QEMU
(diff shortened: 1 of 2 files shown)


stgraber (approved):

stgraber:
Github looks a bit messed up right now, I'll retry the tests a bit later.

stgraber:
Looks like Sphinx has changed in some ways and is causing issues, I'll look into this once I'm near my laptop again.

stgraber:
Looking at what's going on with the doc builds now

stgraber:
My guess is that something changed behavior in upstream Sphinx or MyST, wouldn't be the first time...

PR #787, opened 2024-04-24
incus/network_integrations: Add yaml support for create

awalvie:
Part of #741


stgraber:
Not adding an example on this one?

awalvie:
ah crap, I forgot to stage it

PR #794, opened 2024-04-26
incus/network_zone: Add example for create command

awalvie:
`incus network zone create` already has support for creation from a yaml configuration file, but the same wasn't printed in the usage information of the command.

This commit updates `incus network zone create` to include an example

Part of https://github.com/lxc/incus/issues/741


stgraber:
Can you give that a quick rebase?

PR #800, opened 2024-04-27
client/connection: Add support for the socket existing in /run/incus

Conan-Kudo:
Transient sockets are supposed to be in /run rather than /var, so make it possible to detect that automatically when used.

### DIFF
--- a/client/connection.go
+++ b/client/connection.go
@@ -157,7 +157,8 @@ func ConnectIncusUnix(path string, args *ConnectionArgs) (InstanceServer, error)
 //
 // If the path argument is empty, then $INCUS_SOCKET will be used, if
 // unset $INCUS_DIR/unix.socket will be used and if that one isn't set
-// either, then the path will default to /var/lib/incus/unix.socket.
+// either, then the path will default to /run/incus/unix.socket or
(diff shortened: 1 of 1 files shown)


stgraber:
That won't work as /run/incus exists everywhere but most systems use /var/lib/incus/unix.socket

stgraber:
So the check would need to be about the exact socket path rather than just checking for the directory.

derkuci:
It seems that putting the socket in `/run/incus` would require more changes than just this patch.  For example, the pre-start hook only looks at /var/lib/incus (its command line is something like `/proc/1078/exe callhook /var/lib/incus "default" "vm01" start`) and it fails when the socket is in `/run/incus`.

stgraber:
Hmm, the server side is supposed to be handled through the INCUS_SOCKET environment variable being passed to `incusd`, the `callhook` logic does check for that env variable so it should work fine so long as what calls it doesn't strip the env variable somehow.

derkuci:
I see.  I'll report to Fedora that their RPM spec may miss an environment variable setting (`INCUS_SOCKET=/run/incus/unix.socket`).

Issue #701, opened 2024-03-29
Add an `incus top` command


stgraber:
It would be nice to have an `incus top` command which accesses the metrics API (`/1.0/metrics`) and renders a list of instances with their CPU, memory and disk usage columns (similar to `top`).

That should all be possible with our existing APIs, so it's just CLI work to implement such a command and provide some hotkeys to switch the column sorting and refresh frequency.

CharanSriramUni:
Hey! @stgraber I'm a student at UT Austin (same virtualization class as other posters). Could I claim this issue?

stgraber:
Sure!

CharanSriramUni:
Hey @stgraber! I've gotten some work on this issue and had a few questions:
- Where can I find disk usage and CPU info? `/1.0/metrics` has quite a bit of information on memory usage but it doesn't seem to be attached to a specific instance nor is there any disk/CPU usage.
- Are there any arguments this command should take?

stgraber:
> Are there any arguments this command should take?

Not initially no. I suspect that down the line it will be expanded with more options on what columns to show, what columns to sort by, refresh interval, ... but none of that is needed for this initial implementation.

> How should CPU usage be recorded? `/1.0/metrics` displays CPU usage per core in term of seconds(?) utilized (`incus_cpu_seconds_total{cpu="0",mode="user",name="second",project="default",type="container"} 0.534083444` for example), and the total number of effective CPUs

I think we should just compute the total CPU time in seconds, basically adding up user and system on all CPUs.

CharanSriramUni:
Got it. Sounds good, I'll modify it accordingly.

Relatedly I see that `d.GetMetrics` returns a `string` instead of the Metric type that is defined in the codebase. Is there a way to unmarshal the string into a Go struct of this type? Can't find any examples of string -> MetricSet

stgraber:
> Relatedly I see that `d.GetMetrics` returns a `string` instead of the Metric type that is defined in the codebase. Is there a way to unmarshal the string into a Go struct of this type? Can't find any examples of string -> MetricSet

MetricSet is an internal representation of those metrics on the server side, it's not a struct that's available to clients.

Parsing those lines shouldn't be too difficult to do by hand.
Alternatively using https://pkg.go.dev/github.com/prometheus/common/expfmt may be an option, we just want to be pretty careful not to bring in a lot of extra external code just for this.

stgraber:
[...]

PR #801, opened 2024-04-28
cmd/top: First draft, have some issues

CharanSriramUni:
Working on adding support from `incus top`, but having some issues:
1. When attempting to get the metrics data from `1.0/metrics` using the internal RawQuery mechanism, err is always not nil due to an invalid character being recognized
``` d, err := conf.GetInstanceServer(remote)
	    if err != nil {
		    return err
	    }
	    // Perform the query
	    resp, _, err := d.RawQuery("GET", path, nil, "") // <-- err is: invalid character '#' looking for beginning of value
```
2. Following how `cmd/query` is structured, I try to fallback to raw text if RawQuery doesn't work. Is there a way to Unmarshal OpenMetrics structured strings into the `metrics` struct?
3.  CPU info (at least in the raw text form) doesn't contain CPU usage per instance

Would appreciate any guidance!

### DIFF
--- a/internal/server/network/acl/acl_load.go
+++ b/internal/server/network/acl/acl_load.go
@@ -197,8 +197,8 @@ func UsedBy(s *state.State, aclProjectName string, usageFunc func(ctx context.Co
 		return err
 	}
 
-	for _, aclName := range aclNames {
-		err := s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
+	err = s.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
+		for _, aclName := range aclNames {
 			_, aclInfo, err := tx.GetNetworkACL(ctx, aclProjectName, aclName)
 			if err != nil {
 				return err
@@ -233,35 +233,30 @@ func UsedBy(s *state.State, aclProjectName string, usageFunc func(ctx context.Co
 					return err
 				}
 			}
+		}
 
-			// Find instances using the ACLs. Most expensive to do.
-			err = tx.InstanceList(ctx, func(inst db.InstanceArgs, p api.Project) error {
-				// Get the instance's effective network project name.
-				instNetworkProject := project.NetworkProjectFromRecord(&p)
+		// Find instances using the ACLs. Most expensive to do.
+		err = tx.InstanceList(ctx, func(inst db.InstanceArgs, p api.Project) error {
+			// Get the instance's effective network project name.
+			instNetworkProject := project.NetworkProjectFromRecord(&p)
 
-				// Skip instances who's effective network project doesn't match this Network ACL's project.
-				if instNetworkProject != aclProjectName {
-					return nil
-				}
+			// Skip instances who's effective network project doesn't match this Network ACL's project.
+			if instNetworkProject != aclProjectName {
+				return nil
+			}
 
-				devices := db.ExpandInstanceDevices(inst.Devices.Clone(), inst.Profiles)
+			devices := db.ExpandInstanceDevices(inst.Devices.Clone(), inst.Profiles)
 
-				// Iterate through each of the instance's devices, looking for NICs that are using any of the ACLs.
-				for devName, devConfig := range devices {
-					matchedACLNames := isInUseByDevice(devConfig, matchACLNames...)
-					if len(matchedACLNames) > 0 {
-						// Call usageFunc with a list of matched ACLs and info about the instance NIC.
-						err := usageFunc(ctx, tx, matchedACLNames, inst, devName, devConfig)
-						if err != nil {
-							return err
-						}
+			// Iterate through each of the instance's devices, looking for NICs that are using any of the ACLs.
+			for devName, devConfig := range devices {
+				matchedACLNames := isInUseByDevice(devConfig, matchACLNames...)
+				if len(matchedACLNames) > 0 {
+					// Call usageFunc with a list of matched ACLs and info about the instance NIC.
+					err := usageFunc(ctx, tx, matchedACLNames, inst, devName, devConfig)
+					if err != nil {
+						return err
 					}
 				}
-
-				return nil
-			})
-			if err != nil {
-				return err
 			}
 
 			return nil
@@ -269,6 +264,11 @@ func UsedBy(s *state.State, aclProjectName string, usageFunc func(ctx context.Co
 		if err != nil {
 			return err
 		}
+
+		return nil
+	})
+	if err != nil {
+		return err
 	}
 
 	return nil
--- a/internal/server/network/driver_ovn.go
+++ b/internal/server/network/driver_ovn.go
@@ -3693,7 +3693,7 @@ func (n *ovn) InstanceDevicePortStart(opts *OVNInstanceNICSetupOpts, securityACL
 		var dynamicIPs []net.IP
 
 		// Retry a few times in case port has not yet allocated dynamic IPs.
-		for i := 0; i < 5; i++ {
+		for i := 0; i < 10; i++ {
 			dynamicIPs, err = n.state.OVNNB.LogicalSwitchPortDynamicIPs(instancePortName)
 			if err == nil {
 				if len(dynamicIPs) > 0 {


cmd/incus/main.go
```diff
@@ -11,7 +11,7 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"github.com/lxc/incus/v6/client"
+	incus "github.com/lxc/incus/v6/client"
```
stgraber, line 14: Tiny detail but this change should not be included, `vscode` tends to do that through running `goimports`. `incus` is already the package name for that URL so putting in directly as the import name is needless duplication.

cmd/incus/top.go
```diff
@@ -0,0 +1,162 @@
...
+	return cmd
+}
+
+// Run is a method of the cmdTop structure. It implements the logic to call `incus top`.
+// This function implements the `top` command. It queries the metrics API at (/1.0/metrics) and renders a list of
+// instances with their CPU, memory and disk usage columns
+func (c *cmdTop) Run(cmd *cobra.Command, args []string) error {
+	conf := c.global.conf
+
+	remote, path, err := conf.ParseRemote("/1.0/metrics")
```
stgraber, line 43: So I'm not too sure what's going on here.
`conf.ParseRemote` should be passed `args[0]` as is usual as that just points it to the correct server (e.g. `incus top my-cluster:`)

stgraber: The `path` part is actually irrelevant in this case, you only really need `remote`.

cmd/incus/top.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Validate path
+	if !strings.HasPrefix(path, "/") {
+		return fmt.Errorf(i18n.G("Query path must start with /"))
+	}
+
```
stgraber, line 52: That's not relevant and can be dropped.

cmd/incus/top.go
```diff
@@ -0,0 +1,162 @@
...
+	d, err := conf.GetInstanceServer(remote)
+	if err != nil {
+		return err
+	}
+
+	// Perform the query
```
stgraber, line 58: So this part is odd as you're trying to directly hit `/1.0/metrics` to fetch the information when you could just run `d.GetMetrics()` to get the same data.

stgraber (changes requested):

stgraber:
Left a bunch of comments in-line.

cmd/incus/main.go
CharanSriramUni: Got it!

cmd/incus/top.go
CharanSriramUni: If no argument is passed, should we assume that the remote is local then?

cmd/incus/top.go
CharanSriramUni: Got it!

cmd/incus/top.go
CharanSriramUni: Ah silly miss of mine- didn't realize d.GetMetrics existed. Will change this accordingly.

CharanSriramUni:
Just modified it to address these issues. Also, can this PR be linked to #701? Left a few comments there about some assumption I made.

cmd/incus/top.go
stgraber: Yeah, you can look at `monitor.go` for an example command with similar argument handling.

CharanSriramUni:
UI should update in realtime now like `top`.

Small aside, I'm currently reading user shortcuts from stdin via bufio which requires shortcuts like 'd' (changing delay) to need an "enter" keypress to be registered. There are libraries like `term-box` or `tcell` I could pull in, but wanted to ask before modifying dependencies. 

Also, I'm doing parsing for metrics into the MetricSet struct (locally defined) in this code - it could be useful to move this to an external file for future use involving metrics.

stgraber:
> Small aside, I'm currently reading user shortcuts from stdin via bufio which requires shortcuts like 'd' (changing delay) to need an "enter" keypress to be registered. There are libraries like `term-box` or `tcell` I could pull in, but wanted to ask before modifying dependencies.

Plain bufio is probably fine for now, keeping external dependencies down.

stgraber:
Looks like `make static-analysis` caught a few small things to fix :)

cmd/incus/top.go
```diff
@@ -0,0 +1,282 @@
...
+	global *cmdGlobal
+}
+
+type MetricSet struct {
+	set    map[metrics.MetricType][]metrics.Sample
+	labels map[string]string
+}
+
+func (ms *MetricSet) getMetricValue(metricType metrics.MetricType, instanceName string) float64 {
+	value := 0.0
```
stgraber, line 30: Line break here

cmd/incus/top.go
```diff
@@ -0,0 +1,282 @@
...
+	cli "github.com/lxc/incus/v6/internal/cmd"
+	"github.com/lxc/incus/v6/internal/i18n"
+	"github.com/lxc/incus/v6/internal/server/metrics"
+)
+
+type cmdTop struct {
+	global *cmdGlobal
+}
+
+type MetricSet struct {
```
stgraber, line 24: Since it's used internally only it doesn't need to be exported, so should just be `metricSet`

cmd/incus/top.go
```diff
@@ -0,0 +1,282 @@
...
+	"fmt"
+	"os"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	incus "github.com/lxc/incus/v6/client"
```
stgraber, line 14: This should be import as just:
```
"github.com/lxc/incus/v6/client"
```

cmd/incus/top.go
```diff
@@ -0,0 +1,282 @@
...
+	cli "github.com/lxc/incus/v6/internal/cmd"
+	"github.com/lxc/incus/v6/internal/i18n"
+	"github.com/lxc/incus/v6/internal/server/metrics"
```
stgraber, line 17: What are you getting from this one?

The client should normally never import anything from `internal/server/XYZ`

cmd/incus/top.go
```diff
@@ -0,0 +1,282 @@
...
+	return value
+}
+
+// Command is a method of the cmdTop structure that returns a new cobra Command for managing command aliases.
+func (c *cmdTop) Command() *cobra.Command {
+	cmd := &cobra.Command{}
+	cmd.Use = usage("top")
+	cmd.Short = i18n.G("Manage command aliases")
+	cmd.Long = cli.FormatSection(i18n.G("Description"), i18n.G(
+		`Manage command aliases`))
```
stgraber, line 47: That description is incorrect :)

cmd/incus/top.go
```diff
@@ -0,0 +1,282 @@
...
+	if err != nil {
+		return err
+	}
+
+	d, err := conf.GetInstanceServer(remote)
+	if err != nil {
+		return err
+	}
+
+	// // Start the ticker for periodic updates
```
stgraber, line 81: Weird command identifier, probably just need one `//` :)

cmd/incus/top.go
```diff
@@ -0,0 +1,282 @@
...
+	return metricSet, nil
+}
+
+func parseLabels(input string) map[string]string {
+	labels := make(map[string]string)
+	for _, pair := range strings.Split(input, ",") {
+		kv := strings.Split(pair, "=")
+		if len(kv) != 2 {
+			continue
+		}
```
stgraber, line 267: Line break here

cmd/incus/top.go
```diff
@@ -0,0 +1,282 @@
...
+		key := strings.TrimSpace(kv[0])
+		value := strings.Trim(kv[1], "\"")
+		labels[key] = value
+	}
```
stgraber, line 271: Line break here

cmd/incus/top.go
```diff
@@ -0,0 +1,282 @@
...
+	return labels
+}
+
+func findMetricTypeByName(name string) (metrics.MetricType, bool) {
+	for typ, typName := range metrics.MetricNames {
+		if typName == name {
+			return typ, true
+		}
+	}
```
stgraber, line 280: Line break here

stgraber (changes requested):
Just a few small fixes needed.

Also as this introduces new translated strings, you'll need to commit the result of `make i18n` as `i18n: Update translation templates`

cmd/incus/top.go
CharanSriramUni: This brings in the MetricType struct and MetricNames enum that I'm using to work with the logs. I can also copy the file over here if we want - let me know if this is preferred!

CharanSriramUni:
> make i18n

I get this issue when trying to run `make i18n` - tried it on two different Linux machines with Go Version 1.22. Any insight here?
/bin/sh: 1: xgettext-go: not found

CharanSriramUni:
Accidentally closed this branch, replaced with https://github.com/lxc/incus/pull/817

PR #812, opened 2024-05-01
incusd/network/ovn/nb: Port more NAT functions

Abhiram824:


Abhiram824:
I'm doing a nested loop in DNATSNATDeleteLogicalRouter, not sure if this is the best approach

stgraber (approved):

PR #813, opened 2024-05-02
api: Implement new access API

### LINKED ISSUE #724: Implement new `access` API for projects
It would be nice to have a `/1.0/projects/NAME/access` endpoint which one could `GET` to retrieve the list of who's allowed access and what access level they have on the object.

### LINKED ISSUE #725: Implement new `access` API for instances
It would be nice to have a `/1.0/instances/NAME/access` endpoint which one could `GET` to retrieve the list of who's allowed access and what access level they have on the object.

BajuMcBites:
Implements two new API endpoints to fetch the list of users allowed to access specific projects or instances.

Closes #724
Closes #725


cmd/incusd/instance_access.go
```diff
@@ -0,0 +1,92 @@
...
+//		          description: certificate fingerprint
+//		          example: 636b69519d27ae3b0e398cb7928043846ce1e3842f0ca7a589993dd913ab8cc9
+//		        role:
+//		          type: string
+//		          description: the role associated with the certificate
+//		          example: admin
+//		        provider:
+//		          type: string
+//		          description: Which authorization method the certificate uses
+//		          example: tls
```
stgraber, line 50: It's usually not required to fill that stuff in here as we can just reference the api struct (`Access`) and have swagger pull the model from that instead.

So you'd have something like this instead:
```
//  "200":
//    description: Access
//    schema:
//      type: object
//      description: Sync response
//      properties:
//        type:
//          type: string
//          description: Response type
//          example: sync
//        status:
//          type: string
//          description: Status description
//          example: Success
//        status_code:
//          type: integer
//          description: Status code
//          example: 200
//        metadata:
//          $ref: "#/definitions/Access"
```

cmd/incusd/instance_access.go
```diff
@@ -0,0 +1,92 @@
...
+	// Handle requests targeted to a container on a different node
+	resp, err := forwardedResponseIfInstanceIsRemote(s, r, projectName, name, instanceType)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	if resp != nil {
+		return resp
+	}
+
```
stgraber, line 84: This isn't needed, we don't need to run the logic on the machine that's running the instance.

shared/api/access.go
```diff
@@ -0,0 +1,22 @@
+package api
+
+// access represents who has access to an instance or project
```
stgraber, line 3: Should be `Access` to align with the struct name

shared/api/access.go
```diff
@@ -0,0 +1,22 @@
...
+//
+// swagger:model
+//
+// API extension: instances.
+type Access []AccessEntry
+
+type AccessEntry struct {
```
stgraber, line 10: This struct also needs to have a swagger:model comment so it's properly included in the swagger model.

stgraber:
@BajuMcBites note that https://github.com/lxc/incus/pull/846 has an attempt at pulling the details from OpenFGA which may be applicable here too.

stgraber:
Going to be pulling this one and #846 together, do some cleanup and see how well that all works :)

stgraber:
Did a few changes:
 - Rebased everything on main
 - Added an API extension
 - Split commits a bit differently
 - Included the commits from #846 
 - Ran `gofmt -s -w ./` 
 - Ran `make static-analysis`
 - Used `make update-api` to generate the swagger changes

Left to do:
 - More swagger fixes
 - Implementation of GetInstanceAccess on OpenFGA
 - Actually testing the new API

stgraber:
I've resolved the remaining issues and tested it on an OpenFGA system with the new ListUsers API enabled.

hallyn (approved):

PR #817, opened 2024-05-02
Introduce `incus top`

### LINKED ISSUE #701: Add an `incus top` command

CharanSriramUni:
Closes #701

### DIFF
--- a/cmd/incus/main.go
+++ b/cmd/incus/main.go
@@ -257,6 +257,10 @@ Custom commands can be defined through aliases, use "incus alias" to control tho
 	versionCmd := cmdVersion{global: &globalCmd}
 	app.AddCommand(versionCmd.Command())
 
+	// top sub-command
+	topCmd := cmdTop{global: &globalCmd}
+	app.AddCommand(topCmd.Command())
+
 	// warning sub-command
 	warningCmd := cmdWarning{global: &globalCmd}
 	app.AddCommand(warningCmd.Command())
--- a/cmd/incus/top.go
+++ b/cmd/incus/top.go
@@ -1,0 +1,425 @@
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"regexp"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+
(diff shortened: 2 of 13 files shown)


stgraber:
Looking into this one now

CharanSriramUni:
Awesome; just added the updated translations as well.

stgraber:
Did some simple fixes:
 - Rebased everything down to one commit
 - Added missing Signed-off-by line
 - Added missing bug reference
 - Generated translations
 - Ran `make static-analysis` and fixed what it found

After that looked at the logic itself and did:
 - Made all user visible strings translated
 - Made most errors in `handleKeystrokes` return out of the function to avoid infinite error loops

TODO:
 - Remove dependency on `internal/server/metrics`
 - Use the metrics as the source of instance names (save the GetInstanceNames call) which will also make stopped instances disappear

CharanSriramUni:
@stgraber - Should be good now I think!

Removed the dependency and removed the extraneous call to `GetInstanceNames`. Also, reran translations and checked static-analysis. 

Thanks for all the help with this issue! Really made contributing to my first open source project a ton of fun :)

stgraber:
Getting back to working on this one now.

stgraber:
I did:
 - Rebase on main
 - Cleaned up commits
 - Ran `make static-analysis` and fixed what it reported
 - Changed to show absolute values instead of percentages
 - Changed sorting to show largest first

stgraber (approved):

CharanSriramUni:
Do I need to worry about `Tests / System (1.21.x, cluster, dir) (pull_request) `? Seems to be a massive test from the logs but not sure if it's correct.

stgraber:
Nope, that's fine, a random failure can happen every so often

PR #825, opened 2024-05-03
incus/network_peer: Add example for create command

awalvie:
`incus network peer create` already has support for creation from a yaml configuration file, but the same wasn't printed in the usage information of the command.

This commit updates `incus network peer create` to include an example.

Part of https://github.com/lxc/incus/issues/741


stgraber:
@awalvie needs a rebase because of #826 having been merged now

stgraber (approved):

awalvie:
Ah, sorry about that, was away from my desktop for a bit

Issue #729, opened 2024-04-03
Add instance uptime value


stgraber:
It would be pretty useful to know how long an instance has been running.

This should be done with a millisecond value which for containers can be calculated from when the monitor process was first started, whereas for VMs, we can ask the agent for the uptime and fallback to when the qemu process was started.

We'll want the field added to `InstanceState` and then exposed in `incus info` and as a column in `incus list`.

BajuMcBites:
Hello, my friend @NotPranav0, and I are students at UT Austin and after taking a virtualization class, we wanted to contribute to this project. Could we be assigned this issue? Thanks.

stgraber:
Sure, I assigned it to you!

If @NotPranav0 comments in this issue I can also assign them to it.

NotPranav0:
Comment

BajuMcBites:
Hello, I'm looking to start working on this issue,
- The field being added to IncusState would be something along the lines of ```InstanceUptime``` and would track the total time the instance had been up.

- The main question I have is whether we are tracking the total time an instance has been up or just the time since the last start (does the timer reset upon an instance stop)?

In the first case, would this approach work:
- We would update ```InstanceUptime``` whenever the instance was stopped by using the time the monitor process was started and the current time (assuming new monitor process is created upon instance start call, if not we’d have to store the new start time on every instance start call)
-  When returning this information in ```incus info``` or ```incus list``` we would return the value of the ```InstanceUptime``` + the time elapsed since the last start to get the total uptime.

In the second case:
- Would we need the ```instanceUptime``` variable at all because we could just calculate it whenever it is needed.

Finally, I was wondering where I could find the structs/data for the monitor processes, VM agents, and qemu processes. Thanks.

stgraber:
> * The field being added to IncusState would be something along the lines of `InstanceUptime` and would track the total time the instance had been up.

We can probably make that `StartedAt time.Time `json:"started_at" yaml:"started_at"`

That makes it a bit more consistent with other timestamps type stuff we have.
Unix users are usually used to a timestamp in seconds since boot, but a time.Time is cleaner at the API level, so let's go with that.

> * The main question I have is whether we are tracking the total time an instance has been up or just the time since the last start (does the timer reset upon an instance stop)?

Time since last start. We should be able to just read the creation time of the container PID and use that as the start time so we don't need to record anything.

So you're looking at roughly this set of commits:
 - `api: instance_started_at` (doc/api-extensions.md, internal/version/api.go)
[...]

PR #829, opened 2024-05-03
Add instance start timestamp (uptime)

BajuMcBites:
Issue #729 
- implemented common.StartedAt
- added StartedAt to InstanceState
- called StartedAt on api calls to ```1.0/instances/{name}/state```


stgraber:
Sorted this one out:
 - Added the missing API extension commit
 - Added reference to the API extension to the shared/api commit
 - Generated the swagger API commit
 - Added error handling to the driver_lxc implementation
 - Renamed the function to `processStartedAt` as we don't need it public
 - Ran `make static-analysis` and fixed what it reported

I also added a bit of extra logic to the client to show the new fields in both `incus info` and `incus list`

Issue #486, opened 2024-02-11
Cluster-wide baseline CPU definition for virtual machines


stgraber:
Currently our live-migration logic assumes that all servers are the same and that instances can be migrated to any other server within the cluster so long as they are of the same CPU architecture.

That's obviously not correct as variation in CPU features will cause live-migration to fail.

To resolve this, we should do two things:
 - Add a function that will check if source and destination server share the same CPU. That should then be added to our current migration and evacuation logic to only consider target servers that match the source.
 - As an alternative, add an option to generate a baseline CPU based on supported CPU features across the cluster for a given architecture. This baseline will then be used as the CPU definition for any instance that's set to be migratable (`migration.stateful=true`).

christina-zh:
Im interested in working on this issue, can I be assigned to it please?

stgraber:
This one we'll do in two stages as I'm not entirely sure about how I want to go around the second stage yet :)

For the first stage, what we need is expose the CPU flags/extensions in our resources API as that will be needed to actually compare all servers and see what flags/extensions they all have in common (within one CPU architecture).

So for stage one, you'll want to:
 - Add a new API extension, let's go with `resources_cpu_flags` in `internal/version/api.go` and `doc/api-extensions.md`
 - Add a new `Flags []string` to `ResourceCPUCore` in `shared/api/resource.go`
 - Re-generate the API metadata (`make update-api`)
 - Extend the `/proc/cpuinfo` parsing logic in `internal/server/resources/cpu.go` to fill in the new `Flags` field

That should result in the following commits:
 - `api: resources_cpu_flags`
 - `shared/api: Add Flags to ResourceCPUCore`
 - `doc/rest-api: Refresh swagger YAML`
 - `incusd/resources: Add CPU Flags to ResourceCPUCore`

This one is pretty easy to test at least, once you're running an updated `incusd`, you can run `incus query /1.0/resources` to look at the whole resource dump and check that your CPU flags match what you see in `cat /proc/cpuinfo`

milaiwi:
Is this still being worked on, if not I'd love to take it!

christina-zh:
Yes we are still working on it 👍

PR #834, opened 2024-05-04
Extend resources API to include CPU flags

christina-zh:
Part of #486

This implements the first stage of the work to build a profile of CPU flags for servers within a cluster. This stage extends our internal resources API to have access to CPU flags of the servers.


stgraber:
Looking into this one now.

stgraber (approved):

Issue #718, opened 2024-04-01
Add customizable columns to `incus storage list`


stgraber:
Similarly to `incus list`, `incus image list`, `incus config trust list` and others.

It would be good to have `incus project list` also feature a customizable list of columns.

sohamnair:
Hello, I am new to open source contribution and would love to have a go at it.

stgraber:
Sure, assigned it to you!

sohamnair:
could you please clarify this statement "feature a customizable list of columns". Like what do you expect exactly?

stgraber:
If you look at `incus list --help` or `incus image list --help`, you'll see that they have a `-c` argument (flagColumn) which offers shorthand options for a variety of fields.

For `incus storage list`, the default column list should be the same we have currently, probably with the shorthands `nDSdus`. No need to add additional columns at this stage, adding the feature will at least make it easier for people to hide columns they don't want.

Issue #719, opened 2024-04-01
Add customizable columns to `incus network list`


stgraber:
Similarly to `incus list`, `incus image list`, `incus config trust list` and others.

It would be good to have `incus network list` also feature a customizable list of columns.

amroKerkizz:
Hi! My partner and I are a group from UT and would love to take this issue on if it's still up

stgraber:
Done!

amroKerkizz:
Already have too many issues so I unassigned myself!

sharkman424:
Hello, could my group take over this issue please? We are also from UT Austin doing an open source contribution project.

stgraber:
Done!

sharkman424:
PR opened

PR #842, opened 2024-05-06
incus/storage: Add customizable columns in list

trannlong:
Added customizable columns to incus storage list
This solves issue #718 
Signed-off-by: Long Tran<longtran@utexas.edu>

### DIFF
--- a/cmd/incus/storage.go
+++ b/cmd/incus/storage.go
@@ -25,6 +25,11 @@ type cmdStorage struct {
 	flagTarget string
 }
 
+type storageColumn struct {
+	Name string
+	Data func(api.StoragePool) string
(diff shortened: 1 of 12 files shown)


stgraber:
Sorted this one out:
 - Fixed help message indent
 - Ran `make static-analysis` and fixed what it picked up
 - Generate an i18n commit with `make i18n`
 - Changed column list to match existing order
 - Fixed incorrect d/D column (swapped)
 - Fixed driver, source, state and usedBy columns to actually return the correct data

trannlong:
Sorry, will do the additional steps for issue #719 as well, as we've finished the issue in our group.

PR #845, opened 2024-05-06
Add load average information to resources API 

### LINKED ISSUE #828: Add load average information to resources API
Currently the resources API (`/1.0/resources`) does not provide system load information.
While system load average isn't exactly the best metric ever, it would be useful in some situations like our automated re-balancing.

To address that, we should add a new `load` section to the resources API which would contain:
 - Average1Min
 - Average5Min
 - Average10Min
 - Processes

The first 3 refer to the first 3 fields from /proc/loadavg, the last one would be a total process count on the system which one can do through ReadDir of /proc and counting the number of numeric-only directories.

`incus info --resources` should also be extended as part of this to show the extra fields.

camnwalter:
Resolves #828


stgraber:
Looking into this one now

stgraber:
I've done:
 - Add an API extension
 - Re-structed commits a bit
 - Renamed LoadAverage to just Load
 - Added function descriptions

stgraber (approved):

PR #850, opened 2024-05-06
Partial Implementation of project delete --force

trannlong:
Added support, but not implemented, project delete --force, similar to how it's handled in clusterNodeDelete. 
Added force flag, updated api_project.go, updated interface, added confirmation message when force deleting. 

Getting an error where the server is unable to find the project to delete. Not sure if the server API needs to be updated 
to handle the force flag.

This should hopefully make it so anyone else can pull from this pull request and implement the force deletion in api_project.go


stgraber:
Going to take this one and the early work in #536 to come up with something we can merge.

PR #860, opened 2024-05-09
fix: close resources

testwill:


internal/server/db/schema/update.go
```diff
@@ -22,6 +22,7 @@ func DotGo(updates map[int]Update, name string) error {
 	if err != nil {
 		return fmt.Errorf("failed to open schema.go for writing: %w", err)
 	}
+	defer db.Close()
```
stgraber, line 25: Line break before the Close call.

internal/server/db/schema/update.go
```diff
@@ -42,6 +43,7 @@ func DotGo(updates map[int]Update, name string) error {
 	if err != nil {
 		return fmt.Errorf("failed to open Go file for writing: %w", err)
 	}
+	defer file.Close()
```
stgraber, line 46: Line break before the Close call.

stgraber:
Looks good, CI just found some a small formatting issue based on our coding style, so just two line breaks to add. The other failures are unrelated and caused by a bug in go-json.

stgraber:
Thanks!

Issue #875, opened 2024-05-17
incus-simplestreams add: add --alias flag


melato:
*incus-simplestreams-add* generates an alias for the image using {os}/{release}/{variant}.

This is a feature request to add an "--alias <alias>" flag to the command, to override this default alias.

Rationale:
I want to start using simplestreams to share my images across servers, and I want to keep the image aliases that I've been using already, instead of changing them to the above format in all my scripts.  I also don't want to have slashes in the aliases, because I generate and save each image in a directory on disk, and use the directory name as the image alias.  Being able to specify my own alias will make it easier to start using a simplestreams server for my images, which will facilitate sharing images.

melato:
I will provide a pull request for this shortly.

melato:
Added a pull request: https://github.com/lxc/incus/pull/876
I'm not sure that I followed the correct procedure.

melato:
I submitted a new pull request, #890.

PR #876, opened 2024-05-17
Add `--alias` flag to `incus-simplestreams`

melato:
Add an --alias flag to the command *incus-simplestreams add*  See issue https://github.com/lxc/incus/issues/875

### DIFF
--- a/cmd/incus-simplestreams/main_add.go
+++ b/cmd/incus-simplestreams/main_add.go
@@ -23,6 +23,8 @@ import (
 
 type cmdAdd struct {
 	global *cmdGlobal
+
+	flagAlias string
 }
 
 // Command generates the command definition.
@@ -44,8 +46,13 @@ This command parses the metadata tarball to retrieve the following fields from i
 
 It then check computes the hash for the new image, confirm it's not
 already on the image server and finally adds it to the index.
+
+If no --alias is specified, it generates a default alias:
+	{os}/{release}/{variant}
 `)
 	cmd.RunE = c.Run
+
+	cmd.Flags().StringVarP(&c.flagAlias, "alias", "", "", "Alias"+"``")
 
 	return cmd
 }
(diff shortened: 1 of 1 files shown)


stgraber:
Can you edit your commit to include the `Signed-off-by: Firstname Lastname <email>` line we require (see CONTRIBUTING.md).

From a usability standpoint, I think we should instead have:
 - `--no-default-alias` to turn off the default `distro/release/variant` alias
 - `--alias` to add aliases (can be passed multiple times)

melato:
Ok, will submit a new pull request with --no-default-alias, and adding multiple aliases.
--no-default-alias will allow setting no alias at all.
There is an inconsistency with the "incus publish" command, which does not set a default alias.
*incus-simplestreams add* is another way of publishing.

By the way, why do some flag usages have "``" at the end?  For example, in cmd/incus/publish.go:
	cmd.Flags().StringArrayVar(&c.flagAliases, "alias", nil, i18n.G("New alias to define at target")+"``")
It does not show up in the usage.  Does it indicate something to spf13/cobra?

stgraber:
> There is an inconsistency with the "incus publish" command, which does not set a default alias.
> _incus-simplestreams add_ is another way of publishing.

That's why, those are two different commands so them behaving differently doesn't really bother me. We'd probably have had `incus publish` also put in a default alias if we had known what we were publishing and if the native incus image store was capable of handling multiple images having the same alias.


> By the way, why do some flag usages have "`" at the end? For example, in cmd/incus/publish.go: cmd.Flags().StringArrayVar(&c.flagAliases, "alias", nil, i18n.G("New alias to define at target")+"`")

I don't recall exactly what issue that's working around other than it being a cobra behavior around default values which we didn't want and that this syntax makes it avoid.

melato:
Closing this pull request.  I will create a new one.  I'm learning about pull requests and fork branches.

PR #880, opened 2024-05-17
Add support for ARMv8 to use AAVMF EFI firmware

dustins:
I discovered I wasn't able to run VMs on my RK3588 (ARM64). Although ARM64 [supports UEFI](https://github.com/lxc/incus/blob/main/internal/server/instance/drivers/driver_qemu.go#L1969-L1971) it wasn't able to find appropriate `*.fd` files because it is assumed every architecture was using OVMF. On ARM64 it should be using AAVMF as far as I can tell.

This PR adds support for using AAVMF on ARMv8 architectures and updates the documentation to reflect the code changes.

I've tested creating VMs with these changes on my ARM64 and x86_64 machines but don't have many devices to test against.

### DIFF
--- a/doc/.wordlist.txt
+++ b/doc/.wordlist.txt
@@ -1,4 +1,5 @@
 AAAA
+AAVMF
 ABI
 ACL
 ACLs
@@ -68,6 +69,7 @@ Ebit
 eBPF
 ECDHE
 ECDSA
+EDK
 EiB
 Eibit
 endian
--- a/doc/environment.md
+++ b/doc/environment.md
@@ -36,7 +36,7 @@ Name                            | Description
 `INCUS_EXEC_PATH`               | Full path to the Incus binary (used when forking subcommands)
 `INCUS_IDMAPPED_MOUNTS_DISABLE` | Disable idmapped mounts support (useful when testing traditional UID shifting)
 `INCUS_LXC_TEMPLATE_CONFIG`     | Path to the LXC template configuration directory
-`INCUS_OVMF_PATH`               | Path to an OVMF build including `OVMF_CODE.fd` and `OVMF_VARS.ms.fd`
+`INCUS_EDK2_PATH`               | Path to EDK2 firmware build including `*_CODE.fd` and `*_VARS.fd`
 `INCUS_SECURITY_APPARMOR`       | If set to `false`, forces AppArmor off
 `INCUS_UI`                      | Path to the web UI to serve through the web server
 `INCUS_USBIDS_PATH`             | Path to the hwdata `usb.ids` file
--- a/doc/installing.md
+++ b/doc/installing.md
@@ -319,17 +319,15 @@ In addition, for normal operation, you'll also likely need
 
     sudo zypper install dnsmasq squashfs xz rsync tar attr acl qemu qemu-img qemu-spice qemu-hw-display-virtio-gpu-pci iptables ebtables nftables
 
-As OpenSUSE stores QEMU firmware files using an unusual filename and location, you will need to create some symlinks for them:
-
-    sudo mkdir /usr/share/OVMF
-    sudo ln -s /usr/share/qemu/ovmf-x86_64-4m-code.bin /usr/share/OVMF/OVMF_CODE.4MB.fd
-    sudo ln -s /usr/share/qemu/ovmf-x86_64-4m-vars.bin /usr/share/OVMF/OVMF_VARS.4MB.fd
-    sudo ln -s /usr/share/qemu/ovmf-x86_64-ms-4m-vars.bin /usr/share/OVMF/OVMF_VARS.4MB.ms.fd
-    sudo ln -s /usr/share/qemu/ovmf-x86_64-ms-4m-code.bin /usr/share/OVMF/OVMF_CODE.4MB.ms.fd
 ```
 
 
 ````
+
+```{note}
+On ARM64 CPUs you need to install AAVMF instead of OVMF for UEFI to work with virtual machines.
+In some distributions this is done through a separate package.
+```
 
 ### From source: Build the latest version
 
--- a/internal/server/apparmor/instance.go
+++ b/internal/server/apparmor/instance.go
@@ -7,12 +7,14 @@ import (
 	"strings"
 
 	"github.com/lxc/incus/v6/internal/server/cgroup"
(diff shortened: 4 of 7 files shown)


doc/environment.md
```diff
@@ -36,7 +36,7 @@ Name                            | Description
 `INCUS_EXEC_PATH`               | Full path to the Incus binary (used when forking subcommands)
 `INCUS_IDMAPPED_MOUNTS_DISABLE` | Disable idmapped mounts support (useful when testing traditional UID shifting)
 `INCUS_LXC_TEMPLATE_CONFIG`     | Path to the LXC template configuration directory
-`INCUS_OVMF_PATH`               | Path to an OVMF build including `OVMF_CODE.fd` and `OVMF_VARS.ms.fd`
+`INCUS_EFI_PATH`                | Path to EFI firmware build including `*_CODE.fd` and `*_VARS.fd`
```
stgraber, line 39: I'd probably use `INCUS_EDK2_PATH` instead to be a bit more specific about what we're using.

doc/installing.md
```diff
@@ -322,6 +322,10 @@ As OpenSUSE stores QEMU firmware files using an unusual filename and location, y
 
 ````
 
+```{note}
+On ARM64 CPUs you need to install AAVMF instead of OVMF for UEFI to work with virtual machines.
+```
+
```
stgraber, line 328: I don't think we should need that.

Instead we should just remove all that symlink stuff and make sure that the logic in the code knows to look into /usr/share/qemu too in this case.

internal/server/instance/drivers/driver_qemu.go
```diff
@@ -114,40 +114,60 @@ const qemuBlockDevIDPrefix = "incus_"
 // qemuMigrationNBDExportName is the name of the disk device export by the migration NBD server.
 const qemuMigrationNBDExportName = "incus_root"
 
-// OVMF firmwares.
-type ovmfFirmware struct {
+// EFI firmwares.
```
stgraber, line 117: The rework here looks good, but I think we should make a new `internal/server/instance/drivers/edk2` package which would contain those definitions and possibly some functions to access it.

That would then make it easy to consume from the apparmor package too so we don't need to keep duplicating logic between the two.

doc/environment.md
dustins: Based on your other comment about the ed2k driver would prefer `edk2` be used throughout the code instead of the more generic `efi`?

internal/server/instance/drivers/driver_qemu.go
dustins: Thanks for the feedback. I'll give it a go and push some new commits.

doc/environment.md
stgraber: Yeah, I think so, especially as EFI is misleading anyway given that what we're using today is UEFI, not the old EFI (which I had to deal with a long time ago ;)).

internal/server/instance/drivers/driver_qemu.go
stgraber: Thanks!

stgraber:
@dustins have you had any luck with this one?

dustins:
Yes, I actually am almost done just wanted to test on x86 and ARM64 before pushing. Sorry for the long delay! I think I can test tonight and then push up the changes for you to review.

stgraber:
Sounds good. 

We're releasing 6.2 today so that will most likely miss that release by a hair but it will be in 6.3 for sure.

dustins:
I still have to test this on arm64/x86_64 to check that this all still works.

dustins:
Ok I tested this on my x86_64/arm64 boxes and I was able to launch VMs with and without secure boot.

doc/.wordlist.txt
```diff
@@ -1,4 +1,5 @@
 AAAA
+AAVMF
```
stgraber, line 2: The file also needs to have `EDK2` added to it.

internal/server/instance/drivers/edk2/driver_edk2.go
```diff
@@ -0,0 +1,141 @@
+package edk2
```
stgraber, line 1: This commit is missing the Signed-off-by line

internal/server/instance/drivers/driver_qemu.go
```diff
@@ -40,7 +40,7 @@ import (
 	"google.golang.org/protobuf/proto"
```
stgraber, line 40: This commit is missing the Signed-off-by line

internal/server/apparmor/instance.go
```diff
@@ -7,6 +7,7 @@ import (
 	"strings"
```
stgraber, line 7: This commit is missing the Signed-off-by line

doc/environment.md
```diff
@@ -36,7 +36,7 @@ Name                            | Description
 `INCUS_EXEC_PATH`               | Full path to the Incus binary (used when forking subcommands)
```
stgraber, line 36: This commit is missing the Signed-off-by line

internal/server/instance/drivers/driver_qemu.go
```diff
@@ -40,7 +40,7 @@ import (
...
 	"gopkg.in/yaml.v2"
 
-	"github.com/lxc/incus/v6/client"
+	incus "github.com/lxc/incus/v6/client"
```
stgraber, line 43: Needless change caused by goimports.

stgraber:
Looking good, just a few changes needed to get the tests to be happy!

doc/environment.md
dustins: What's the correct way for me to resolve the commits without the Signed-off-by line?

stgraber:
Looking at this one now

doc/environment.md
stgraber: Sorry, I missed that question. The way to do it is `git rebase -i origin`, then use the `reword` action on the commits so you can easily edit them and tweak.

If only sign-off is needed, `git rebase origin --sign-off` can also do it automatically.

stgraber:
I did:
 - Fixed the missing sign-offs
 - Rebased some commits together
 - Fixed the import rename
 - Ran `make static-analysis` and fixed what it picked up

stgraber:
Done cleaning this one up and doing some testing on it.

stgraber (approved):

PR #889, opened 2024-05-22
Various bugfixes

stgraber:


shared/idmap/shift_linux.go
```diff
@@ -280,18 +280,63 @@ static int get_userns_fd_cb(void *data)
...
+		return -errno;
+
+	// Setup gid_map
+	snprintf(path, sizeof(path), "/proc/%d/gid_map", pid);
+	fd = openat(AT_FDCWD, path, O_WRONLY);
+	if (fd < 0)
+		return -errno;
+
+	if (write(fd, "0 0 1", 5) != 5)
+		return -errno;
```
hallyn, line 334: Note that since this is c and not go, fd will not be cleaned up in this error path.

hallyn (changes requested):
(looks good otherwise)

shared/idmap/shift_linux.go
stgraber: Fixed!

stgraber:
Added close calls to the write error handling path, that should do it.

hallyn (approved):

shared/idmap/shift_linux.go
```diff
@@ -279,19 +279,72 @@ static int get_userns_fd_cb(void *data)
...
+	// Setup gid_map
+	snprintf(path, sizeof(path), "/proc/%d/gid_map", pid);
+	file_fd = openat(AT_FDCWD, path, O_WRONLY);
+	if (file_fd < 0)
+		goto err_process;
+
+	if (write(file_fd, "0 0 1", 5) != 5)
+		goto err_process;
+
+	if (close(file_fd) < 0)
```
hallyn, line 331: If you close file_fd here, you should set it back to -EBADF before potentially closing it a second time.

stgraber: Done!

hallyn (approved):

PR #891, opened 2024-05-23
incus/storage_volume/snapshot: Support YAML for creation

### LINKED ISSUE #741: Add ability to read YAML from stdin to all `create` commands

stgraber:
Closes #741

### DIFF
--- a/cmd/incus/storage_volume.go
+++ b/cmd/incus/storage_volume.go
@@ -2313,6 +2313,11 @@ func (c *cmdStorageVolumeSnapshotCreate) Command() *cobra.Command {
 	cmd.Short = i18n.G("Snapshot storage volumes")
 	cmd.Long = cli.FormatSection(i18n.G("Description"), i18n.G(
 		`Snapshot storage volumes`))
+	cmd.Example = cli.FormatSection("", i18n.G(`incus storage volume snapshot create default v1 snap0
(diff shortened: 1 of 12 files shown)


hallyn (approved):
looks fine so long as you intend for the stdinData.Description to be ignored.

stgraber:
@hallyn I think that StorageVolumeSnapshot having an editable Description is actually a bug, the description should be read-only and come from the original volume at the time it was snapshotted (same as instances).

PR #909, opened 2024-06-01
Avoid double user-user- in network description

keestux:
The description was showing a description like this: `Network for user restricted project user-user-1001`.


stgraber (approved):

stgraber:
Oops, nice catch!

PR #922, opened 2024-06-07
Initial support for OCI images

stgraber:


cmd/incusd/main_forknet.go
```diff
@@ -71,6 +71,27 @@ static void forkdonetdetach(char *file) {
...
+
+	// Attach to the network namespace.
+	if (dosetns_file(path, "net") < 0) {
+		fprintf(stderr, "Failed setns to container network namespace: %s\n", strerror(errno));
+		_exit(1);
+	}
+
+	// Jump back to Go for the rest
+}
+
```
awalvie, line 94: Out of curiosity, why do we have C code here? As reference for the go implementation?

stgraber: Because Go can't safely do this. Any switch of network namespace as done here must be done from cgo prior to the Go runtime starting up.

Note that those aren't comments, it's actual code that gets compiled in and run, the comment syntax is just how cgo works.

internal/server/instance/drivers/driver_lxc.go
```diff
@@ -2312,6 +2313,125 @@ func (d *lxc) startCommon() (string, []func() error, error) {
...
+				Terminal bool `json:"terminal"`
+				User     struct {
+					UID uint32 `json:"uid"`
+					GID uint32 `json:"gid"`
+				} `json:"user"`
+				Args []string `json:"args"`
+				Env  []string `json:"env"`
+				Cwd  string   `json:"cwd"`
+			} `json:"process"`
+		}
```
dontlaugh, line 2335: Sooner than later you will want to deserialize into the upstream spec types ([runtime spec and/or image spec](https://github.com/opencontainers/image-spec/blob/v1.1.0/conversion.md#verbatim-fields)). They publish reference struct types in Go, and in implementations I've seen it's common to alias these

```go
import imagev1 "github.com/opencontainers/image-spec/specs-go"

type ImageManifest = imagev1.Manifest  // or maybe struct embedding
```

This will help with determining whether it's a bug in Incus or a malformed image.

stgraber: I'm not saying this struct in that repo. At least not when looking for the fields we're reading here.

dontlaugh: Forgive me, I mixed up image spec vs. runtime spec https://github.com/opencontainers/runtime-spec/blob/main/specs-go/config.go

cmd/incusd/main_forknet.go
awalvie: Oh, gotcha!

client/oci_images.go
```diff
@@ -0,0 +1,312 @@
...
+	err = os.Mkdir(filepath.Join(ociPath, "image"), 0700)
+	if err != nil {
+		return nil, err
+	}
+
+	// Copy the image.
+	_, err = subprocess.RunCommand(
+		"skopeo",
+		"--insecure-policy",
+		"copy",
```
dungdm93, line 119: Is there any reason `skopeo` and `umoci` is called as sub-process instead of used as a library?

cmd/incusd/main_forknet.go
```diff
@@ -188,6 +225,107 @@ func (c *cmdForknet) RunInfo(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+// RunDHCP runs a one time DHCPv4 client and applies address, route and DNS configuration.
+func (c *cmdForknet) RunDHCP(cmd *cobra.Command, args []string) error {
+	var messages []*dhcpv4.DHCPv4
+
+	link := &ip.Link{
+		Name: "eth0",
+	}
```
stgraber, line 234: Skip DHCP if not found.

cmd/incusd/main_forknet.go
```diff
@@ -188,6 +225,107 @@ func (c *cmdForknet) RunInfo(cmd *cobra.Command, args []string) error {
...
+		}
+	}
+
+	if reply == nil {
+		return fmt.Errorf("Didn't get a lease")
+	}
+
+	if reply.YourIPAddr == nil || reply.YourIPAddr.Equal(net.IPv4zero) || reply.SubnetMask() == nil || len(reply.Router()) != 1 || len(reply.DNS()) < 1 {
+		return fmt.Errorf("Didn't get a usable address")
+	}
```
stgraber, line 269: Don't fail, just print error message

client/oci_images.go
stgraber: Mostly simplicity at this stage. Neither appear to be particularly designed to be consumed as Go modules and we also don't want to find ourselves importing too many indirect Go dependencies into the core of our project.

As both `skopeo` and `umoci` are readily packaged in most distributions, just using the external commands is best for now.

client/oci_images.go
```diff
@@ -0,0 +1,361 @@
...
+	// Convert to something usable.
+	if req.ProgressHandler != nil {
+		req.ProgressHandler(ioprogress.ProgressData{Text: "Unpacking the OCI image"})
+	}
+
+	_, err = subprocess.RunCommand(
+		"umoci",
+		"unpack",
+		"--image", filepath.Join(ociPath, "oci"),
+		filepath.Join(ociPath, "image"))
```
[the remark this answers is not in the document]
stgraber: Ah yeah, I wasn't aware of that option, it indeed looks like something we should pass.

test/suites/container_oci.sh
```diff
@@ -0,0 +1,19 @@
...
+  if [ -n "${INCUS_OFFLINE:-}" ]; then
+    echo "==> SKIP: Skipping OCI tests as running offline"
+    return
+  fi
+
+  ensure_has_localhost_remote "${INCUS_ADDR}"
+  incus network create inct$$
+
+  incus remote add docker https://docker.io --protocol=oci
+  incus launch docker:hello-world --console --ephemeral --network=inct$$
```
tych0, line 11: I don't exactly know the semantics of `--network=...` here, but I wonder if this is something that should be implied for OCI containers?

tych0 (approved):
one comment and one question, but lgtm otherwise.

client/oci_images.go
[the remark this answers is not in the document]
tych0: I wonder if we shouldn't make it the default in umoci...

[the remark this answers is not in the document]
stgraber: Done now!

test/suites/container_oci.sh
stgraber: It's just something that's needed for the testsuite.

Normal users will have an `incusbr0` bridge that's part of their default profile, providing them with an `eth0` that has both IPv4 and IPv6 connectivity.

Our testsuite instead normally just has a basic `p2p` device for tests, providing no connectivity. Because OCI containers run a DHCP client on startup, I had to make the test use a temporary network that does provide DHCP so the test would run quickly and not hit the 5s DHCP timeout for every container.

tych0: I see, makes sense, thanks.

stgraber:
Examples of normal usage:
```
stgraber@dakara:~$ incus launch docker:hello-world --ephemeral --console
Launching the instance
Instance name is: native-imp

Hello from Docker!
This message shows that your installation appears to be working correctly.

To generate this message, Docker took the following steps:
 1. The Docker client contacted the Docker daemon.
 2. The Docker daemon pulled the "hello-world" image from the Docker Hub.
    (amd64)
 3. The Docker daemon created a new container from that image which runs the
    executable that produces the output you are currently reading.
 4. The Docker daemon streamed that output to the Docker client, which sent it
    to your terminal.

To try something more ambitious, you can run an Ubuntu container with:
 $ docker run -it ubuntu bash

Share images, automate workflows, and more with a free Docker ID:
 https://hub.docker.com/

For more examples and ideas, visit:
 https://docs.docker.com/get-started/


stgraber@dakara:~$ incus launch docker:caddy
Launching the instance
Instance name is: known-ewe                                    
stgraber@dakara:~$ incus list known-ewe
+-----------+---------+-----------------------+-------------------------------------------+-----------------+-----------+
|   NAME    |  STATE  |         IPV4          |                   IPV6                    |      TYPE       | SNAPSHOTS |
+-----------+---------+-----------------------+-------------------------------------------+-----------------+-----------+
| known-ewe | RUNNING | 172.17.250.189 (eth0) | 2602:fc62:c:250:216:3eff:fe95:8bcb (eth0) | CONTAINER (APP) | 0         |
+-----------+---------+-----------------------+-------------------------------------------+-----------------+-----------+
stgraber@dakara:~$ 
```

client/oci_images.go
adamcstephens: Would it be possible to get new commands/dependencies added to the release notes? Or maybe keep a running list in the packaging doc? 

(This is amazing btw :)

PR #934, opened 2024-06-12
Varied bugfixes

stgraber:


cmd/incusd/networks.go
```diff
@@ -489,6 +489,19 @@ func networksPost(d *Daemon, r *http.Request) response.Response {
...
 		}
 
+		n, err := network.LoadByName(s, projectName, req.Name)
+		if err != nil {
+			return response.SmartError(fmt.Errorf("Failed loading network: %w", err))
+		}
+
+		err = s.Authorizer.AddNetwork(r.Context(), projectName, req.Name)
+		if err != nil {
+			logger.Error("Failed to add network to authorizer", logger.Ctx{"name": req.Name, "project": projectName, "error": err})
```
hallyn, line 499: So you don't want to return an error in this case?

client/incus_oidc.go
```diff
@@ -147,6 +148,40 @@ func (o *oidcClient) do(req *http.Request) (*http.Response, error) {
 	return resp, nil
 }
 
+// dial function executes a websocket request and handles OIDC authentication and refresh.
+func (o *oidcClient) dial(dialer websocket.Dialer, uri string, req *http.Request) (*websocket.Conn, *http.Response, error) {
+	conn, resp, err := dialer.Dial(uri, req.Header)
+	if err != nil && resp == nil {
```
hallyn, line 154: So what if err != nil but resp != nil ?  You will ignore the error in that case.  Is that the right thing?

stgraber: Yeah because that's specifically the case we want to retry.

stgraber: The special logic in this function is to handle cases where we get err != nil (Bad handshake) but have a resp filled in with the response code (403 - Unauthorized) and headers (OIDC issuer and client id).

When that happens, we refresh the OIDC token and then re-issue the dialer.Dial with the new token.

hallyn (approved):

PR #956, opened 2024-06-25
Installation documentation update to work on both Debian and Ubuntu

AJRepo:
Was following the instructions on Debian 12 (bookworm) and had an issue with the command that had 

`sudo apt install .... liblxc-dev ... `

because on Debian `liblxc-dev` could not be found.  

I switched to an Ubuntu 24.04 and found that `liblxc-dev` is a transitional package for `lxc-dev`.  I.e.

```
$ apt-cache search liblxc-dev
liblxc-dev - Transitional package - liblxc-dev -> lxc-dev
``` 

Thus changing `liblxc-dev`  to `lxc-dev`  documents a  process that works on both Debian and Ubuntu.

This PR updates the docs for the above reasons.

### DIFF
--- a/doc/installing.md
+++ b/doc/installing.md
@@ -288,7 +288,7 @@ Also, due to a [`gettext` issue](https://github.com/gosexy/gettext/issues/1), yo
 Install the build and required runtime dependencies with:
 
     sudo apt update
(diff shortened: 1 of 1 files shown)


stgraber:
@AJRepo hey there! There seems to be a mistmatch between the Author of your git commit and the Signed-off-by line which is causing our DCO check to fail. Can you line the two up?

AJRepo:
Will do.

If I do a `git commit --amend --force; git push` would that mess up anything?  If you'd prefer I can close/cancel this PR and start over with a new branch.

AJRepo:
Actually I found another issue related to Debian 12 and compiling. I'll cancel this PR and resubmit with both.

PR #958, opened 2024-06-26
Installation documentation update to work on both Debian and Ubuntu

AJRepo:
Was following the instructions on Debian 12 (bookworm) and had an issue with the command that had 

`sudo apt install .... liblxc-dev ... `

because on Debian `liblxc-dev` could not be found.  

I switched to an Ubuntu 24.04 and found that `liblxc-dev` is a transitional package for `lxc-dev`.  I.e.

```console
$ apt-cache search liblxc-dev
liblxc-dev - Transitional package - liblxc-dev -> lxc-dev
``` 

Thus changing `liblxc-dev`  to `lxc-dev`  documents a  process that works on both Debian 12 and Ubuntu 24.

Also ran into an issue compiling as Debian 12 installs golang-go version 1.19 which is too old to compile successfully. Added a note for those running Debian stable. 

This PR updates the docs for the above reasons.

### DIFF
--- a/doc/installing.md
+++ b/doc/installing.md
@@ -288,7 +288,8 @@ Also, due to a [`gettext` issue](https://github.com/gosexy/gettext/issues/1), yo
 Install the build and required runtime dependencies with:
 
     sudo apt update
-    sudo apt install acl attr autoconf automake dnsmasq-base git golang-go libacl1-dev libcap-dev liblxc1 liblxc-dev libsqlite3-dev libtool libudev-dev liblz4-dev libuv1-dev make pkg-config rsync squashfs-tools tar tcl xz-utils ebtables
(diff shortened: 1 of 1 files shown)


AJRepo:
I see the markdownlinter test failed on a few items. Have fixed and pushed

stgraber:
Looking into this one now

stgraber:
I did a quick rebase of the changes together but then decided that rather than adding a big chunk of text about using Debian's backports repository, we should just point to our Go requirements and mention getting a supported version from upstream.

That's because we only ever support the last two Go versions, which currently means 1.22 and 1.21 but will soon be 1.23 and 1.22 at which point Debian providing Go 1.21 through backports won't be good enough anymore.

stgraber (approved):

PR #971, opened 2024-07-05
Add hint about installing incus-tools

ThomasLeister:
Add hint about installing incus-tools beforehand. If just the incus package was installed, `incus-to-lxd` is not found.


stgraber:
I think I'd prefer for distro specific instructions to be limited to the installing.md page. 

So might be worth adding a NOTE on that page instead about installing incus-tools if you need lxd-to-incus.

PR #973, opened 2024-07-08
Improve storage volume help

AJRepo:
As someone who was new to incus, I had some confusion about "type" as it related to volumes. This help text update hopefully helps others who are  new to incus not have that same confusion. 

This update does the following: 

* Moved description text from "cmd.Example" section to cmd.Long section.  Now the example sections only have examples. 

* Added additional example to --help for "incus storage volume" command to show virtual-machine example

* added Error text for when the default (custom) is used and reports "not found". Note: this introduces an `if` statement which gives different error messages depending on if someone uses `custom` or not. If that is not ok, feel free to remove. 

* Ran `make i18n` (separate commit)

### DIFF
--- a/cmd/incus/storage_volume.go
+++ b/cmd/incus/storage_volume.go
@@ -159,9 +159,9 @@ type cmdStorageVolumeAttach struct {
 func (c *cmdStorageVolumeAttach) Command() *cobra.Command {
 	cmd := &cobra.Command{}
 	cmd.Use = usage("attach", i18n.G("[<remote>:]<pool> <volume> <instance> [<device name>] [<path>]"))
-	cmd.Short = i18n.G("Attach new storage volumes to instances")
+	cmd.Short = i18n.G("Attach new custom storage volumes to instances")
 	cmd.Long = cli.FormatSection(i18n.G("Description"), i18n.G(
-		`Attach new storage volumes to instances`))
+		`Attach new custom storage volumes to instances`))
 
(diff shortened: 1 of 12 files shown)


cmd/incus/storage_volume.go
```diff
@@ -951,12 +951,12 @@ func (c *cmdStorageVolumeEdit) Command() *cobra.Command {
...
-	cmd.Example = cli.FormatSection("", i18n.G(
-		`Provide the type of the storage volume if it is not custom.
-Supported types are custom, image, container and virtual-machine.
+		`Edit storage volume configurations as YAML
 
-incus storage volume edit [<remote>:]<pool> [<type>/]<volume> < volume.yaml
+Provide the type of the storage volume if it is not custom.
+Supported types are custom, image, container and virtual-machine.`))
+	cmd.Example = cli.FormatSection("", i18n.G(
+		`incus storage volume edit [<remote>:]<pool> [<type>/]<volume> < volume.yaml
```
stgraber, line 959: Examples should be functional, so not include variables or optional entries.

So in this case we'd probably have to change it to something like:
```
incus storage volume edit default foo < volume.yaml
  Update the foo custom storage volume in the default pool with YAML data from volume.yaml
```

stgraber:
I'm taking a look at this one as there are a few more edits that are needed here I think.

stgraber (approved):

PR #979, opened 2024-07-09
incusd/device/network: Fix Tap interface MTU when in OVN

trunet:
When creating a virtual-machine with ovn, the tap interface inherits the MTU from ovs-system and therefore doesn't have the MTU.

This doesn't happen on container veth interface, the logic for containers are working fine.

I believe it's a bug, and this PR fixed the issue for me.

Let me know if you need any changes.

Thanks for the awesome work in incus.

### DIFF
--- a/internal/server/device/device_utils_network.go
+++ b/internal/server/device/device_utils_network.go
@@ -330,29 +330,36 @@ func networkCreateTap(hostName string, m deviceConfig.Device) (uint32, error) {
 	// The host side should always line up with the bridge to avoid accidentally lowering the bridge MTU.
 	// The instance side should use the configured MTU (if any), if not, it should match the host side.
 	var mtu uint32
-	if m["mtu"] != "" {
-		nicMTU, err := strconv.ParseUint(m["mtu"], 10, 32)
-		if err != nil {
-			return 0, fmt.Errorf("Invalid MTU specified: %w", err)
-		}
-
-		mtu = uint32(nicMTU)
-	}
+	var instanceMTU uint32
+	var parentMTU uint32
 
 	if m["parent"] != "" {
-		parentMTU, err := network.GetDevMTU(m["parent"])
+		mtu, err := network.GetDevMTU(m["parent"])
 		if err != nil {
 			return 0, fmt.Errorf("Failed to get the parent MTU: %w", err)
 		}
 
-		err = NetworkSetDevMTU(hostName, parentMTU)
+		parentMTU = uint32(mtu)
+	}
+
+	if m["mtu"] != "" {
+		mtu, err := strconv.ParseUint(m["mtu"], 10, 32)
 		if err != nil {
-			return 0, fmt.Errorf("Failed to set the MTU %d: %w", mtu, err)
+			return 0, fmt.Errorf("Invalid MTU specified: %w", err)
 		}
 
-		if mtu == 0 {
-			mtu = parentMTU
-		}
+		instanceMTU = uint32(mtu)
+	}
+
+	if instanceMTU > parentMTU {
+		mtu = instanceMTU
+	} else {
+		mtu = parentMTU
+	}
+
+	err = NetworkSetDevMTU(hostName, mtu)
+	if err != nil {
+		return 0, fmt.Errorf("Failed to set the MTU %d: %w", mtu, err)
 	}
 
 	// Set TX queue length on both ends.


trunet:
If you want more context, my config is like this:

```
❯ incus network show UPLINK
config:
  bgp.peers.router01.address: [REDACTED]
  bgp.peers.router01.asn: "[REDACTED]"
  bgp.peers.router02.address: [REDACTED]
  bgp.peers.router02.asn: "[REDACTED]"
  dns.nameservers: [REDACTED]
  ipv4.gateway: [REDACTED].9.1/24
  ipv4.ovn.ranges: [REDACTED].50-[REDACTED].254
  ipv4.routes: [REDACTED].12.0/22
  mtu: "9000"
  ovn.ingress_mode: routed
  volatile.last_state.created: "false"
description: ""
name: UPLINK
type: physical
used_by:
- /1.0/networks/ovn-platform
managed: true
status: Created
locations:
- srv-[REDACTED]
- srv-[REDACTED]
- srv-[REDACTED]
project: default
❯ incus network show ovn-platform
config:
  bridge.mtu: "8942"
  ipv4.address: [REDACTED].12.1/24
  ipv4.nat: "false"
  ipv6.address: none
  network: UPLINK
  volatile.network.ipv4.address: [REDACTED].9.50
description: ""
name: ovn-platform
type: ovn
used_by:
- /1.0/instances/u1
- /1.0/instances/u2
managed: true
status: Created
locations:
- srv-[REDACTED]
- srv-[REDACTED]
- srv-[REDACTED]
project: default
```

before the change:
```
27: ovs-system: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN group default qlen 1000
    link/ether ca:a2:3a:67:ee:26 brd ff:ff:ff:ff:ff:ff
43: tap4536cadb: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq master ovs-system state UP group default qlen 1000
    link/ether 56:a4:3b:18:f7:e9 brd ff:ff:ff:ff:ff:ff
```

after the change:
```
27: ovs-system: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN group default qlen 1000
    link/ether 9a:a5:c0:a8:66:09 brd ff:ff:ff:ff:ff:ff
36: tap14bd52f8: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 8942 qdisc mq master ovs-system state UP group default qlen 1000
    link/ether 5e:f9:09:04:32:3d brd ff:ff:ff:ff:ff:ff
```

internal/server/device/device_utils_network.go
```diff
@@ -337,6 +337,11 @@ func networkCreateTap(hostName string, m deviceConfig.Device) (uint32, error) {
 		}
 
 		mtu = uint32(nicMTU)
+
+		err = NetworkSetDevMTU(hostName, mtu)
+		if err != nil {
+			return 0, fmt.Errorf("Failed to set the MTU %d: %w", mtu, err)
+		}
```
stgraber, line 344: So the comment above remains true, when dealing with a bridged interface we need to make sure not to set a low MTU on the device as the bridge will change its own MTU to the lowest MTU present.

Setting values higher than the bridge's own value is usually not an issue though.

trunet: if I understand correctly, I need to check the parentMTU and only set if it’s higher or none? in my case with OVN/OVS, parentMTU is empty.

stgraber: Yeah, that's right if `m["parent"]` is set, then we should only set the device MTU if it's higher than the parent. If `m["parent"]` isn't set, then we can just set the provided MTU.

trunet:
It's done, also rebased. can you review again @stgraber ?

internal/server/device/device_utils_network.go
```diff
@@ -330,29 +330,36 @@ func networkCreateTap(hostName string, m deviceConfig.Device) (uint32, error) {
...
+			return 0, fmt.Errorf("Invalid MTU specified: %w", err)
 		}
 
-		if mtu == 0 {
-			mtu = parentMTU
-		}
+		instanceMTU = uint32(mtu)
+	}
+
+	if (parentMTU > 0 && instanceMTU > parentMTU) || parentMTU == 0 {
```
stgraber, line 354: Isn't that the same as just `instanceMTU > parentMTU`?

trunet: instanceMTU > parentMTU doesn't catch parentMTU being 0 (not set). if that's ok for you, I can do that.

stgraber:
Thanks!

PR #981, opened 2024-07-10
Compute a cluster-wide baseline CPU definition for VMs

### LINKED ISSUE #486: Cluster-wide baseline CPU definition for virtual machines

stgraber:
For virtual machines with `migration.stateful=true` and so therefore capable of being live-migrated or just have their runtime state restored on another system, we will now pull the list of CPU flags from all servers of the same architecture across the cluster and generate a CPU definition made from whatever flags they all have in common.

This should result in virtual machines that can be migrated even across different CPU generations.

Closes #486

### DIFF
--- a/cmd/incusd/daemon.go
+++ b/cmd/incusd/daemon.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"crypto/x509"
 	"database/sql"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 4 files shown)


stgraber:
Worth noting that this is a first stab at this and an implementation which is a bit naive but will be backportable to the LTS branch.

A more advanced way of handling this will be added by allowing cluster groups to hold per-architecture CPU definitions. This will then make it possible to have different cluster groups within a cluster with a different virtualized CPU definition and so being able to migrate between servers in that group.

hallyn (approved):

PR #991, opened 2024-07-13
incusd/instance/edk2: Support OVMF filenames on arm64

stgraber:
Quite a few environments use OVMF on arm64 rather than the more technicaly correct AAVMF name.


stgraber:
Going to force that one through as soon as a test build has succeeded as we really need this to unblock issues on arm64.

PR #992, opened 2024-07-13
incusd/instance/drivers/qemu: Limit CPU flag calculation to x86_64

stgraber:
The "kvm64" CPU definition doesn't exist on aarch64 and possibly on other platforms, so restrict the logic to x86_64 for the time being.


stgraber:
Going to be rushing that one too as it's another blocker to running aarch64 VMs at the moment.

PR #1014, opened 2024-07-18
Remove Socket Path to Let ConnectIncusUnix Resolve It

### LINKED ISSUE #1013: Issue locating socket while joining cluster
# Required information

 * Distribution: Rocky Linux
 * Distribution version: 9.4
   * Kernel version: 5.14.0-427.22.1.el9_4.x86_64
   * LXC version: 4.0.12
   * Incus version: 6.2
   * Storage backend in use: dir

# Issue description

I have a cluster leader configured. I am trying to join a cluster member. When I run `incus --force-local admin init --preseed` I get the following error:

> Error: Failed to join cluster: Failed to connect to local server: Get "http://unix.socket/1.0": dial unix /var/lib/incus/unix.socket: connect: no such file or directory

The socket is not located in /var/lib/incus, but is in /run/incus.

# Steps to reproduce

On a Fedora or RHEL based system:

 1. Configure cluster leader with dir storage and bridge network
 2. Create a token for the cluster member on the cluster leader via `incus cluster add <member hostname>`
 3. Attempt to join the cluster
     - I don't believe the method of joining the cluster matters (interactive vs preseed)

# Information to attach

 - [ ] Any relevant kernel output (`dmesg`)
 - [ ] Container log (`incus info NAME --show-log`)
 - [ ] Container configuration (`incus config show NAME --expanded`)
 - [ ] Main daemon log (at /var/log/incus/incusd.log)
 - [ ] Output of the client with --debug
 - [ ] Output of the daemon with --debug (alternatively output of `incus monitor --pretty` while reproducing the issue)

amayer5125:
If the path passed to `ConnectIncusUnix` is empty it will resolve to the correct path.

The [comment above GetUnixSocket](https://github.com/lxc/incus/blob/main/internal/server/sys/os.go#L233) mentions that it is used for tests. I'm not sure if it should be used in these locations or not, but removing it fixed the issue I was having with Incus attempting to use the incorrect socket path.

Fixes #1013

### DIFF
--- a/cmd/incusd/api_cluster.go
+++ b/cmd/incusd/api_cluster.go
@@ -598,7 +598,7 @@ func clusterPutJoin(d *Daemon, r *http.Request, req api.ClusterPut) response.Res
 		// As ServerAddress field is required to be set it means that we're using the new join API
 		// introduced with the 'clustering_join' extension.
 		// Connect to ourselves to initialize storage pools and networks using the API.
-		localClient, err := incus.ConnectIncusUnix(d.os.GetUnixSocket(), &incus.ConnectionArgs{UserAgent: clusterRequest.UserAgentJoiner})
+		localClient, err := incus.ConnectIncusUnix("", &incus.ConnectionArgs{UserAgent: clusterRequest.UserAgentJoiner})
 		if err != nil {
 			return fmt.Errorf("Failed to connect to local server: %w", err)
 		}
--- a/cmd/incusd/api_project.go
+++ b/cmd/incusd/api_project.go
@@ -1014,7 +1014,7 @@ func projectDelete(d *Daemon, r *http.Request) response.Response {
 		}
 
 		// Connect to the local server.
-		target, err := incus.ConnectIncusUnix(s.OS.GetUnixSocket(), nil)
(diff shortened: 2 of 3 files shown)


stgraber:
So I'm not sure that this is really a bug and this fix is actually problematic for a few reasons:

1) `GetUnixSocket` and its result is used in a bunch of other places throughout the daemon and should always return the correct path.
2) The full logic when passing "" also involves trying to talk to the unprivileged incus daemon. If that one is present and the main sock is gone for some reason, we're going to start activating that daemon and cause a whole bunch of confusion.

I suspect the reason why you're ending up in this situation is because your system uses socket activation, so has a systemd unit which binds /run/incus/unix.socket. Incusd then gets started and inherits the socket but is never told where that socket actually lives.

If that's the issue, just edit `incus.service` and set `INCUS_SOCKET=/run/incus/unix.socket` and everything will work as expected.

amayer5125:
I believe you are correct. I commented on the issue that there is a [pull request to the Fedora package](https://src.fedoraproject.org/rpms/incus/pull-request/2) to add the `INCUS_SOCKET` environment variable like you mentioned. I'll close this PR in favor of setting the environment variable.

PR #1015, opened 2024-07-18
incusd/storage: Use writeback mode for qemu-img convert

stgraber:


stgraber:
In my tests, I'm seeing at least a 30-40% improvement in unpack speed on larger images.

internal/server/storage/utils.go
```diff
@@ -690,7 +690,7 @@ func ImageUnpack(imageFile string, vol drivers.Volume, destBlockFile string, sys
 
 		cmd = []string{
 			"nice", "-n19", // Run with low priority to reduce CPU impact on other processes.
-			"qemu-img", "convert", "-f", "qcow2", "-O", "raw",
+			"qemu-img", "convert", "-f", "qcow2", "-O", "raw", "-t", "writeback",
```
nanjj, line 693: May forget put `-p` here since the tracker is trying to get the progress.

stgraber: `-p` sends the progress output to `stdout` and seems to use terminal escape sequences to reset the line. That's a bit tricky for us to handle as we don't need updates to happen that frequently and dealing with the escape sequences isn't super fun.

Instead the approach in this PR uses the `SIGUSR1` handler in `qemu-img` which doesn't require the `-p` argument and will instead print one line to `stderr` every time.

nanjj: Got it

hallyn (approved):

PR #1017, opened 2024-07-19
Incus action support --console vga

nanjj:
Supports below usage:
1. incus start instance --console,
2. incus start instance --console=vga,
3. incus start instance --console vga # my favor one

### DIFF
--- a/cmd/incus/action.go
+++ b/cmd/incus/action.go
@@ -430,6 +430,19 @@ func (c *cmdAction) Run(cmd *cobra.Command, args []string) error {
 			return fmt.Errorf(i18n.G("--console can't be used with --all"))
 		}
 
+		if len(names) > 1 {
+			idx := slices.Index(names, "vga")
+			idx2 := slices.Index(names, "console")
+			if idx == -1 || (idx2 != -1 && idx > idx2) {
+				idx = idx2
+			}
+
+			if idx != -1 {
+				c.flagConsole = names[idx]
+				names = append(names[0:idx], names[idx+1:]...)
+			}
+		}
+
 		if len(names) != 1 {
 			return fmt.Errorf(i18n.G("--console only works with a single instance"))
 		}


stgraber:
That doesn't work.

```
incus create images:ubuntu/24.04 vga --vm
incus start --console vga
```

The above needs to give me the text console for the instance named `vga`.

nanjj:
> That doesn't work.
> 
> ```
> incus create images:ubuntu/24.04 vga --vm
> incus start --console vga
> ```
=>
```
To detach from the console, press: <ctrl>+a q
...
```
So it works. And `incus stop --console vga` =>
```
To detach from the console, press: <ctrl>+a q
[  OK  ] Stopped target Swap.
[  OK  ] Unmounted /boot/efi.
```
also works.
> 
> The above needs to give me the text console for the instance named `vga`.

stgraber:
What happens if you do `incus start --console console vga` ?

I would expect that to start a text console on instance named `vga`, but from my reading of the code, this will instead give me a `vga` console on instance `console`.

nanjj:
> What happens if you do `incus start --console console vga` ?
> 
> I would expect that to start a text console on instance named `vga`, but from my reading of the code, this will instead give me a `vga` console on instance `console`.

Fixed.

stgraber:
I think this will fail if I do `incus start vga --console console` as it will incorrectly try to use the VGA console for it now.

simos:
> Supports below usage:
> 
>     1. incus start instance --console,
> 
>     2. incus start instance --console=vga,
> 
>     3. incus start instance --console vga # my favor one

You are suggesting that `--console` should be able to used either on its own (without parameter, assuming the default to be _vga_),

or as `--console vga`, which means that the next string after `--console` will be the name of reserved parameter `vga`.

Such a feature would add ambiguity to the users. 
In the first example, `--console` does not consume the next token and then it does. 
In the second example, there's ambiguity. Is `vga` a parameter to `--console` or is it the name of the instance?

```
incus start --console mycontainer
incus start --console vga mycontainer
```
and
```
incus init images:debian/12 vga
incus start --console vga
```

PR #1018, opened 2024-07-19
Make run-parts running compatible with different versions

nanjj:
The option `--exit-on-error' exists only on debian run-parts version. On centos, redhat or federa, run-parts has no such option.

The fix checks version and adds options accordingly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -312,7 +312,7 @@ endif
 	flake8 test/deps/import-busybox
 	shellcheck --shell sh test/*.sh test/includes/*.sh test/suites/*.sh test/backends/*.sh test/lint/*.sh
 	shellcheck test/extras/*.sh
-	run-parts --verbose --exit-on-error --regex '.sh' test/lint
+	run-parts $(shell run-parts -V >/dev/null 2>&1 && echo -n "--verbose --exit-on-error --regex '.sh'") test/lint
 
 .PHONY: staticcheck
 staticcheck:


test/lint/licenses.sh
```diff
@@ -1,2 +1,3 @@
 #!/bin/sh -eu
+shell command -v go-licenses || go install github.com/google/go-licenses@latest
```
stgraber, line 2: The `go install` call is already in the Makefile alongside the other checks used for static analysis

Makefile
```diff
@@ -312,7 +312,7 @@ endif
 	flake8 test/deps/import-busybox
 	shellcheck --shell sh test/*.sh test/includes/*.sh test/suites/*.sh test/backends/*.sh test/lint/*.sh
 	shellcheck test/extras/*.sh
-	run-parts --verbose --exit-on-error --regex '.sh' test/lint
+	run-parts $(shell run-parts -V 2> /dev/null 1> /dev/null && echo -n "--verbose --exit-on-error --regex '.sh'") test/lint
```
stgraber, line 315: Why not the traditional `>/dev/null 2>&1`?

test/lint/licenses.sh
nanjj: done

Makefile
nanjj: done

stgraber (approved):

Issue #985, opened 2024-07-11
Implement per cluster group baseline CPU definitions for VMs


stgraber:
Building on top of what we've landed a few days ago, rather than just having migratable VMs use a cluster-wide CPU baseline, it would be good to have that be configurable by cluster group. This would allow larger clusters to have a mix of different hardware generations and so long as the workloads are targeted to the right groups, have them be migratable within those groups.

To do this we'll need a bit of infrastructure as currently there is no ability to put configuration or any kind of extra data onto a cluster group. So extending cluster groups to have our usual key/value config handling would be a good start.

Then either on top of that or as its own concept, have the ability to store data on a per-architecture basis which would contain both a baseline CPU (defaulting to `kvm64`) and then a series of flags to enable or disable. Setting the `flags` field to `auto` will have Incus scan the servers in the cluster group and populate the value with the current shared set of flags.

PR #1034, opened 2024-07-22
Introduce cluster group configuration

stgraber:
This is infrastructure work ahead of #985

This adds the DB table, API and allows the usual `user.XYZ` config keys.


cmd/incus/cluster_group.go
```diff
@@ -783,3 +795,207 @@ func (c *cmdClusterGroupAdd) Run(cmd *cobra.Command, args []string) error {
...
+		return nil, cobra.ShellCompDirectiveNoFileComp
+	}
+
+	return cmd
+}
+
+// Run runs the actual command logic.
+func (c *cmdClusterGroupGet) Run(cmd *cobra.Command, args []string) error {
+	// Quick checks.
+	exit, err := c.global.CheckArgs(cmd, args, 2, 2)
```
hallyn, line 835: do you care if err != nil but !exit?

hallyn (commented):
changing spacing on the sql is unfortunate

stgraber:
> changing spacing on the sql is unfortunate

It's obviously not required, but we had a big mess of tabs and spaces which actually end up making their way into sqlite3 and makes it a bit of a pain to visually inspect the schema. So I figured I'd finally bite the bullet and spend 30min making everything consistent and then re-gen the schema.

It's all whitespace changes basically, so those two first commits can be ignored. If the schema had been altered in some way, the auto-generation of the final schema would most likely have failed (schema.go is generated by applying every single update on top of the initial schema) and if for some reason the regression wasn't leading to invalid SQL, we'd almost certainly catch it in tests.

hallyn (approved):

PR #1038, opened 2024-07-23
GO code staticcheck cleanup

nanjj:
Ignored code generated files and fixed the left checks


cmd/incus/admin_recover.go
```diff
@@ -10,7 +10,7 @@ import (
 	"golang.org/x/text/cases"
 	"golang.org/x/text/language"
 
-	"github.com/lxc/incus/v6/client"
+	incus "github.com/lxc/incus/v6/client"
```
stgraber, line 13: Looks like you ran goimports or something like that which has a tendency to go mess with our import names, can you remove all of those from your commit?

cmd/incusd/instance.go
```diff
@@ -427,7 +427,9 @@ func instanceCreateAsCopy(s *state.State, opts instanceCreateAsCopyOpts, op *ope
 					"path": "/",
 					"pool": instRootDiskDevice["pool"],
 				}
-			} else { //nolint:staticcheck // (keep the empty branch for the comment)
+				//nolint:all
+				//lint:ignore SA9003 keep the empty branch for the comment
+			} else { //nolint:all
```
stgraber, line 432: Not sure what's going on here, that change looks off.

internal/server/db/cluster/warnings.mapper.go
```diff
@@ -3,6 +3,7 @@
 package cluster
 
 // The code below was generated by incus-generate - DO NOT EDIT!
+//lint:file-ignore U1000,SA4006 Ignore staticcheck
```
stgraber, line 6: As suggested by the line above it, your change will get deleted the next time we run `make update-schema`.

I also don't think that it's right to ignore all lint on those files because they are auto-generated. Issues in those files are still issues, they just need to be fixed in the generator code rather than in the files themselves.

internal/server/db/generate/file/write.go
```diff
@@ -28,6 +28,7 @@ func Reset(path string, imports []string, buildComment string, iface bool) error
 	content := fmt.Sprintf(`%spackage %s
 
 // The code below was generated by %s - DO NOT EDIT!
+//lint:file-ignore U1000,SA4006 Ignore staticcheck
```
stgraber, line 31: So just because the files are auto-generated doesn't mean that issues/lints aren't worth fixing.

It just means that rather than fixing the issues directly in the `.mapper.go` files, those issues need to instead be fixed in the generator code.

internal/server/response/swagger.go
```diff
@@ -1,6 +1,8 @@
 // Package response contains helpers for rendering HTTP responses.
 //
-//nolint:deadcode,unused
+//nolint:all
```
stgraber, line 3: Why ignoring all lints?

shared/cliconfig/remote.go
```diff
@@ -332,6 +332,8 @@ func (c *Config) getConnectionArgs(name string) (*incus.ConnectionArgs, error) {
 		// Golang has deprecated all methods relating to PEM encryption due to a vulnerability.
 		// However, the weakness does not make PEM unsafe for our purposes as it pertains to password protection on the
 		// key file (client.key is only readable to the user in any case), so we'll ignore deprecation.
+		//nolint:all
```
stgraber, line 335: Why turn off all linting?

cmd/incus/admin_recover.go
```diff
@@ -241,7 +241,8 @@ func (c *cmdAdminRecover) Run(cmd *cobra.Command, args []string) error {
 	// Send /internal/recover/import request to the daemon.
 	// Don't lint next line with gosimple. It says we should convert reqValidate directly to an RecoverImportPost
 	// because their types are identical. This is less clear and will not work if either type changes in the future.
-	reqImport := recover.ImportPost{ //nolint:gosimple
+	//nolint:all
```
stgraber, line 244: Why turn off all linting?

internal/server/cluster/membership.go
```diff
@@ -593,6 +593,8 @@ func NotifyHeartbeat(state *state.State, gateway *Gateway) {
 		// Wait for heartbeat to finish and then release.
 		// Ignore staticcheck "SA2001: empty critical section" because we want to wait for the lock.
 		gateway.HeartbeatLock.Lock()
+		//nolint:all
```
stgraber, line 596: Why turn off all linting?

internal/server/cluster/membership.go
```diff
@@ -593,6 +593,8 @@ func NotifyHeartbeat(state *state.State, gateway *Gateway) {
...
+		//lint:ignore SA2001 we want to wait for the lock
```
stgraber, line 597: The comment above (`// Ignore staticcheck ...`) can be removed given this one just repeats it

Issue #1084, opened 2024-08-07
Add CLI support for bulk environment variables


bketelsen:
# Required information

 * Distribution: debian
 * Distribution version: bookworm
   * Incus version: 6.3 daily

# Feature Request

I've *really* been enjoying the OCI container support in Incus. In the process of moving my self-hosting bits from docker-compose to incus one thing that I've noticed is how much more typing it is to add `-c environment.MYVAR=myvalue` to the command line.  I'm proposing adding a command line flag to the `init/create` and `launch` subcommands that will read environment variables from a file in standard `KEY=value` format, one pair per line.

## Example
Given a file called `.env` with values:

```
MYSQL_PASSWORD=youruncle
MYSQL_USER=bob
```

Incus command could look like this:

```
incus init docker:mysql:latest --env .env
```

The command logic would read the lines from the .env file and merge them into the configuration before init/create/launch.

If acceptable, I'm willing to make this change as a pull request.

Thanks for incus!

bketelsen:
Flag name `--env` is of course up for debate, but it appears that `--env` is available.

stgraber:
I think that's fine, though I'd go for the longer `--environment-file`, a bit more of a pain to type (though that's what shell completion is for :)) but also more descriptive of what it will do.

PR #1085, opened 2024-08-07
incus: Add support for environment file (.env)

bketelsen:
This change adds `--environment-file` to `init/create/launch` commands and reads environment value key pairs from the referenced file, adding them to the config map for the container.


bketelsen:
fixes #1084

stgraber (approved):

bketelsen:
thanks for catching the println I left.

PR #1092, opened 2024-08-08
incusd/device/disk: Allow relative paths within custom volumes

### LINKED ISSUE #993: Allow passing sub-directories of custom volumes
Something that'd be particularly useful now that we support application containers is to be able to pass in a specific directory from within a custom volume as a disk for a container.

This would then allow creating a single custom volume that holds the configuration/data of multiple application containers. The whole volume can then be passed to a system container so that configuration/backups/whatever can be performed from there.

stgraber:
Closes #993


stgraber:
Opening as draft as the actual path logic isn't in there yet, for now it's just allowing the syntax and I want to make sure that this didn't regress anything.

hallyn (approved):

PR #1104, opened 2024-08-11
client: docs improvements

decentral1se:
Follows https://discuss.linuxcontainers.org/t/getting-starting-with-client-api-incus-golang-api/21332

Unsure re: explanation on using `incus` because my `vim-go` auto converts `"github.com/lxc/incus/client"` into `incus "github.com/lxc/incus/client"` 🤔 Maybe you have a wording here that will clarify? Can update the PR...


stgraber (approved):
Thanks!

PR #1112, opened 2024-08-12
incusd: allow custom oidc scope

james-d-elliott:
This adds the option to customize the effective scope of authorizations via OpenID Connect 1.0.

### DIFF
--- a/cmd/incusd/api_1.0.go
+++ b/cmd/incusd/api_1.0.go
@@ -224,7 +224,7 @@ func api10Get(d *Daemon, r *http.Request) response.Response {
 	// Get the authentication methods.
 	authMethods := []string{api.AuthenticationMethodTLS}
 
-	oidcIssuer, oidcClientID, _, _ := s.GlobalConfig.OIDCServer()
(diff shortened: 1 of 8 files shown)


james-d-elliott:
One key element of note. No validation exists to ensure the theoretically required scopes `openid` and `offline_access` exist. Looking for preferential direction here as it could be left to the user entirely, could be automatically added, or some validation method may exist that I'm unaware of.

stgraber:
Hello,

Can you edit your commit to include the required `Signed-off-by` line?

In general, I think it makes sense to make the OIDC scope list be configurable so we can indeed request access to additional userinfo data that can then be used alongside `oidc.claim`. The `profile` scope for `preferred_username` being the most likely combination.

I'd probably call it `oidc.scopes` and have it default to `openid, offline_access`. We have a convenient helper to parse those kind of comma+space separated lists. `SplitNTrimSpace`

james-d-elliott:
Hey thanks for the feedback. I believe I've made the requested adjustments. I'm not super familiar with the codebase so I may have missed something. I just copied previous similar additions.

stgraber (approved):

PR #1115, opened 2024-08-12
Add support for manual QMP configuration

bensmrs:
This PR allows to inject QMP commands at various stages of a VM initialization

### DIFF
--- a/doc/.wordlist.txt
+++ b/doc/.wordlist.txt
@@ -211,6 +211,7 @@ Podman
 PTS
 qdisc
 QEMU
+QMP
 qgroup
 qgroups
 RADOS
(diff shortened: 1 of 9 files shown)


bensmrs:
HERE BE DRAGONS: if you set one of the keys to `"[0]"`, or basically any integer list, it makes Incus hang indefinitely. I think there is a problem with DigitalOcean’s QMP driver. We need to sanitize the input, but I haven’t had the courage to explore the problem in more detail.

bensmrs:
I’m really not an expert in Go debugging, but Delve suggests that the QMP drivers hangs here: https://github.com/digitalocean/go-qemu/blob/2e3d0186973edb6a2811ee1240651bda63d7687d/qmp/socket.go#L250
Not sure what to do with that :)

stgraber:
Static analysis test suggests you need to run `gofmt -s -w ./`

bensmrs:
Sorry for all this noise, I’m slowly discovering the vast quantity of (yet unsuspected) checks

stgraber (approved):

stgraber:
Thanks, looks good!

I just did a bit of moving things around between commits, added an API extension for this and updated the project low-level keys to add those (and a couple we had missed).

PR #1136, opened 2024-08-19
incusd/storage/lvm: Fix resize logic to conserve LV state

stgraber:
We want the activation state of the LV to be identical following a resize operation, otherwise we may end up having the caller assume it's still active and end up writing data to a file in /dev rather than to an actual device.

This was caught in our daily tests but took a little while to track down.

### DIFF
--- a/internal/server/storage/drivers/driver_lvm_volumes.go
+++ b/internal/server/storage/drivers/driver_lvm_volumes.go
@@ -458,9 +458,15 @@ func (d *lvm) SetVolumeQuota(vol Volume, size string, allowUnsafeResize bool, op
 			}
 
 			// Activate volume if needed.
-			_, err := d.activateVolume(vol)
+			activated, err := d.activateVolume(vol)
 			if err != nil {
 				return err
+			}
+
+			if !activated {
+				defer func() {
+					_, _ = d.activateVolume(vol)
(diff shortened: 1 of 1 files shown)


internal/server/storage/drivers/driver_lvm_volumes.go
```diff
@@ -458,11 +458,17 @@ func (d *lvm) SetVolumeQuota(vol Volume, size string, allowUnsafeResize bool, op
...
 			// Activate volume if needed.
-			_, err := d.activateVolume(vol)
+			activated, err := d.activateVolume(vol)
 			if err != nil {
 				return err
 			}
 
+			if !activated {
+				defer func() {
+					_, _ = d.activateVolume(vol)
```
hallyn, line 468: Long function so I'm probably missing some context, but it would seem like you would want to defer a *de*activate here if !activated?

hallyn: Oh, I see, th ecomment above activateVolume() says:

785 // activateVolume activates an LVM logical volume if not already present. Returns true if activated, false if not.

So it's true if activateVolume() itself managed to activate it, not if it was
activated beforehand.

hallyn (approved):

PR #1144, opened 2024-08-21
lxd-to-incus: Handle Incus socket in /run/incus/

stgraber:


tych0:
Is this for backwards compat? or?

stgraber:
Fedora and distros using EPEL are configuring Incus to have its socket in /run/incus/.

So that change makes our migration tool automatically pick up the correct path on those platforms.

PR #1162, opened 2024-08-29
Add startup scriptlets

bensmrs:
This PR adds the ability to call scriptlets at various stages of a VM initialization.
Several modifications to the codebase are proposed:
* A `qemudefault` subpackage has been created to avoid a dependency cycle when allowing the QEMU driver to call scriptlets;
* The scriptlet logger has been moved to a dedicated file, to be reused by other scriptlets;
* The scriptlet creation functions have been transformed to helper functions, to make writing new scriptlets easy;
* A Starlark value unmarshaller has been written
* And finally, the startup scriptlet has been written

For now, the scriptlet only allows running QMP queries, but feel free to add anything you want there!

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2582,3 +2582,7 @@ This introduces a new `oidc.scopes` server configuration key which can take a co
 ## `network_integrations_peer_name`
 
 This extends `ovn.transit.pattern` to allow `peerName` as a template variable.
+
+## `qemu_scriptlet`
+
+This adds the ability to run a scriptlet at various stages of startup: using the `raw.qemu.scriptlet` configuration key.
--- a/doc/config_options.txt
+++ b/doc/config_options.txt
@@ -705,6 +705,14 @@ See {ref}`instance-options-qemu` for more information.
 
 ```
 
+```{config:option} raw.qemu.scriptlet instance-raw
+:condition: "virtual machine"
+:liveupdate: "no"
+:shortdesc: "QEMU scriptlet to run at early, pre-start and post-start stages"
+:type: "string"
+
+```
+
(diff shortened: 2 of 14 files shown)


doc/api-extensions.md
```diff
@@ -2578,3 +2578,11 @@ The following new configuration options are introduced:
...
 
 This introduces a new `oidc.scopes` server configuration key which can take a comma separate list of OIDC scopes to request from the identity provider.
+
+## `startup_scriptlet`
+
+This adds the ability to run scriptlets at various stages of startup:
+
+* `startup.scriptlet.early`
+* `startup.scriptlet.pre-start`
+* `startup.scriptlet.post-start`
```
stgraber, line 2588: I still think this would be better handled by a single `raw.qemu.scriptlet` which gets passed the hook name as an argument.

The reason to avoid a generic `startup.scriptlet` type infrastructure is that we need to think about security. QMP calls are inherently unsafe so this must be a very restricted key that requires a trusted project, so having that particular scriptlet be under `raw.qemu` with the rest of the low-level qemu keys makes sense.

bensmrs: Got it! I understand your remark, but because it is very dangerous to inject QMP commands, I thought that we might as well allow everything in a generic `startup.scriptlet`.

Before I do the proper renamings, are you talking about passing the hook name as an argument to the Starlark function and letting the user deal with it with an `if`/`elif`/`else` construct? What would you want the function to be called if `startup` isn’t suited?

The changes needed are pretty light, I’ll try to do them today.

stgraber (approved):

stgraber:
Thanks!

PR #1199, opened 2024-09-07
doc: Fix limits.memory default value unit

kznrluk:
The documentation for `limits.memory` currently states `1Gib`, but this doesn't seem to be a valid unit.

https://linuxcontainers.org/incus/docs/main/reference/instance_units/

Today, I tried to configure settings based on this documentation, but I encountered an error stating that the unit was invalid, which made me realize the unit was misspelled.

Using my IDE's search function, I quickly searched for `Gib`, `Mib`, and `Tib`, but found no other occurrences.

### DIFF
--- a/doc/config_options.txt
+++ b/doc/config_options.txt
@@ -829,7 +829,7 @@ See {ref}`instance-options-limits-hugepages` for more information.
 ```
 
 ```{config:option} limits.memory instance-resource-limits
-:defaultdesc: "`1Gib` (VMs)"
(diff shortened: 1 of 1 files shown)


kznrluk:
```
[kznrluk@ex ~]$ incus config set aki0 limits.memory=16Gib
Error: Invalid config: Invalid value: 16Gib
[kznrluk@ex ~]$ incus config set aki0 limits.memory=16GiB
[kznrluk@ex ~]$ echo $?
0
```

stgraber:
Thanks!

PR #1201, opened 2024-09-07
[WIP] Add support for ipv4.dhcp.ranges

winiciusallan:
As long as discussed here [1], OVN has a configuration to exclude IPs from DHCP pool and it was not supported yet. So using 'ipv4.dhcp.ranges', we get reserverd IPs using negative logic for provided subnet

[1]: `https://discuss.linuxcontainers.org/t/ovn-dhcp-ranges/21290/3`i


stgraber:
Replaced by #1341

PR #1214, opened 2024-09-11
Add qcow2 and vmdk support to incus-migrate

gibmat:


cmd/incus-migrate/main_migrate.go
```diff
@@ -348,6 +350,21 @@ func (c *cmdMigrate) RunInteractive(server incus.InstanceServer) (cmdMigrateData
...
 		}
 
+		// When migrating a VM, ensure we only process supported inputs
+		if config.InstanceArgs.Type == api.InstanceTypeVM {
+			if linux.IsBlockdevPath(s) {
+				// pass
+			} else if _, ext, _, _ := archive.DetectCompression(s); ext == ".qcow2" {
+				// pass
+			} else if strings.HasSuffix(s, ".raw") {
+				// TODO -- need to actually inspect the image for MBR/GPT header
```
gibmat, line 360: Still trying to figure out what the best route would be to reliably identify raw images. There's github.com/diskfs/go-diskfs, but that might be a bit heavy just for this. Could also call `file` or `fdisk` to identify the image.

cmd/incus-migrate/main_migrate.go
```diff
@@ -537,7 +554,7 @@ func (c *cmdMigrate) Run(cmd *cobra.Command, args []string) error {
 	// Automatically clean-up the temporary path on exit
 	defer func(path string) {
 		_ = unix.Unmount(path, unix.MNT_DETACH)
-		_ = os.Remove(path)
+		_ = os.RemoveAll(path)
```
gibmat, line 557: The root.img file is still left behind. I think it might still be mounted when the temp directory is cleaned up.

stgraber:
Removed `[WIP]` from title and marked as draft instead.

cmd/incus-migrate/main_migrate.go
stgraber: I'd probably do it as a fallback, so if we don't see the magic for QCOW2 or VMDK, just assume it's a raw image.

We can have the tool indicate the detected format in the overview.

cmd/incus-migrate/main_migrate.go
stgraber: I'd definitely avoid `RemoveAll` as for containers if the unmount fails somehow, we may find ourselves deleting a lot of actual data...

stgraber: I don't know what that Unmount is doing here too, I don't see us ever making that path a mount.

I'd still avoid `RemoveAll` though as for containers we do mount stuff in sub-paths.
If that doesn't go away correctly, we may find ourselves recursively deleting a lot of data with RemoveAll.

We should know what files to expect in there so we can specifically delete those.

gibmat: OK, switched back to just `Remove` and I added the potentially converted image as another path to remove.

Do we want to make an issue about the temp directory itself not being properly cleaned up?

cmd/incus-migrate/main_migrate.go
```diff
@@ -554,11 +554,15 @@ func (c *cmdMigrate) Run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// Define name for converted VM image so it can be easily passed to the cleanup function
+	convertedImg := filepath.Join(path, "converted-raw-image.img")
+
 	// Automatically clean-up the temporary path on exit
-	defer func(path string) {
+	defer func(path, convertedImg string) {
```
stgraber, line 561: Tiny code style nit, but we usually always type each argument even if they're the same type.

stgraber:
```

.tmp/doc/howto/import_machines_to_instances.md:106: MD034 Bare URL used
.tmp/doc/howto/import_machines_to_instances.md:210: MD034 Bare URL used
```

cmd/incus-migrate/main_migrate.go
stgraber: Do you have a list of whats' left in there?

gibmat:
```
gibmat@futurfusion:/tmp$ sudo ls -ltr /tmp/incus-migrate_mount_1349025577/
total 0
-rw-r--r-- 1 root root 0 Sep 13 08:34 root.img
```

stgraber: Okay, so we should go ahead and delete that one prior to attempting to delete the directory.

stgraber (approved):

PR #1217, opened 2024-09-12
internal/server/instance/drivers: Disable 9p and vsock for Windows VMs

gibmat:
Windows guests don't support virtiofs-9p or virtio-vsock devices, so don't include them in the generated qemu config.

Depends on the instance's configuration field `image.os` containing the string "Windows" somewhere in its value.

### DIFF
--- a/internal/server/instance/drivers/driver_qemu.go
+++ b/internal/server/instance/drivers/driver_qemu.go
@@ -3384,25 +3384,28 @@ func (d *qemu) generateQemuConfigFile(cpuInfo *cpuTopology, mountInfo *storagePo
 
 	cfg = append(cfg, qemuTablet(&tabletOpts)...)
 
-	// Existing vsock ID from volatile.
-	vsockID, err := d.getVsockID()
-	if err != nil {
-		return "", nil, err
-	}
+	// Windows doesn't support virtio-vsock.
+	if !strings.Contains(strings.ToLower(d.expandedConfig["image.os"]), "windows") {
+		// Existing vsock ID from volatile.
+		vsockID, err := d.getVsockID()
+		if err != nil {
+			return "", nil, err
+		}
 
-	devBus, devAddr, multi = bus.allocate(busFunctionGroupGeneric)
-	vsockOpts := qemuVsockOpts{
-		dev: qemuDevOpts{
-			busName:       bus.name,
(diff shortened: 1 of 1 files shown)


internal/server/instance/drivers/driver_qemu.go
```diff
@@ -3384,25 +3384,28 @@ func (d *qemu) generateQemuConfigFile(cpuInfo *cpuTopology, mountInfo *storagePo
 
 	cfg = append(cfg, qemuTablet(&tabletOpts)...)
 
-	// Existing vsock ID from volatile.
-	vsockID, err := d.getVsockID()
-	if err != nil {
-		return "", nil, err
-	}
+	// Windows doesn't support virtio-vsock
```
stgraber, line 3387: Tiny nit on comment style, we usually start with a capital letter and end with a period, so would be:

```
// Windows doesn't support virtio-vsock.
```

(We've very slowly been converting some comments that don't fit the pattern as we change code near them)

internal/server/instance/drivers/driver_qemu.go
```diff
@@ -3384,25 +3384,28 @@ func (d *qemu) generateQemuConfigFile(cpuInfo *cpuTopology, mountInfo *storagePo
...
+	if !strings.Contains(d.expandedConfig["image.os"], "Windows") {
```
stgraber, line 3388: I'd get the value through `strings.ToLower()` to avoid `Windows` then not being considered as windows.

internal/server/instance/drivers/driver_qemu.go
```diff
@@ -3450,25 +3453,10 @@ func (d *qemu) generateQemuConfigFile(cpuInfo *cpuTopology, mountInfo *storagePo
...
-		name:     "config",
-		protocol: "9p",
-		path:     d.configDriveMountPath(),
-	}
-
-	cfg = append(cfg, qemuDriveConfig(&driveConfig9pOpts)...)
-
-	// Pass in the agents if INCUS_AGENT_PATH is set.
-	if util.PathExists(os.Getenv("INCUS_AGENT_PATH")) {
+	// Windows doesn't support virtio-9p
```
stgraber, line 3456: Same comment style nit.

PR #1225, opened 2024-09-18
Support for ACLs for bridge NIC device when using nftables driver.

mikerobski:
Allows "security.acls*" fields to be used to apply ACLs to bridge NIC device when the firewall driver is nftables.

Since the nftables do not support "reject" rules, the implementation converts the default rules from "reject" to "drop" when needed.

The ACL rules are applied together with the filtering rules, if specified. The filtering rules are applied before the ACL rules and are enforced even if the ACL definition contains allow rule that permits the traffic.

### DIFF
--- a/doc/howto/network_acls.md
+++ b/doc/howto/network_acls.md
@@ -212,7 +212,7 @@ When using network ACLs with a bridge network, be aware of the following limitat
 - Unlike OVN ACLs, bridge ACLs are applied only on the boundary between the bridge and the Incus host.
   This means they can only be used to apply network policies for traffic going to or from external networks (see exception for `nftables` firewall driver below).
   They cannot be used for to create {spellexception}`intra-bridge` firewalls, thus firewalls that control traffic between instances connected to the same bridge.
-- When using the `nftables` firewall driver you can apply ACLs to the NIC device and control traffic between the instances. In this case the `reject` ACL rules are not permitted and when the default action is set to `reject` it is interpreted as `drop`.
+- When using the `nftables` firewall driver you can apply ACLs to the NIC device and control traffic between the instances. In this case the `reject` ACL rules applied to the ingress traffic are converted to `drop` to address nftables limitation.
 - {ref}`ACL groups and network selectors <network-acls-selectors>` are not supported.
 - When using the `iptables` firewall driver, you cannot use IP range subjects (for example, `192.0.2.1-192.0.2.10`).
 - Baseline network service rules are added before ACL rules (in their respective INPUT/OUTPUT chains), because we cannot differentiate between INPUT/OUTPUT and FORWARD traffic once we have jumped into the ACL chain.
--- a/doc/howto/network_acls.md
+++ b/doc/howto/network_acls.md
@@ -210,9 +210,9 @@ incus config device set <instance_name> <device_name> security.acls.default.ingr
 When using network ACLs with a bridge network, be aware of the following limitations:
 
 - Unlike OVN ACLs, bridge ACLs are applied only on the boundary between the bridge and the Incus host.
-  This means they can only be used to apply network policies for traffic going to or from external networks (see exception for `nftables` firewall driver below).
+  This means they can only be used to apply network policies for traffic going to or from external networks (see exception for {spellexception}`nftables` firewall driver below).
   They cannot be used for to create {spellexception}`intra-bridge` firewalls, thus firewalls that control traffic between instances connected to the same bridge.
-- When using the `nftables` firewall driver you can apply ACLs to the NIC device and control traffic between the instances. In this case the `reject` ACL rules applied to the ingress traffic are converted to `drop` to address nftables limitation.
+- When using the {spellexception}`nftables` firewall driver you can apply ACLs to the NIC device and control traffic between the instances. In this case the `reject` ACL rules applied to the ingress traffic are converted to `drop` to address {spellexception}nftables limitation.
 - {ref}`ACL groups and network selectors <network-acls-selectors>` are not supported.
 - When using the `iptables` firewall driver, you cannot use IP range subjects (for example, `192.0.2.1-192.0.2.10`).
 - Baseline network service rules are added before ACL rules (in their respective INPUT/OUTPUT chains), because we cannot differentiate between INPUT/OUTPUT and FORWARD traffic once we have jumped into the ACL chain.
--- a/doc/howto/network_acls.md
+++ b/doc/howto/network_acls.md
@@ -210,8 +210,9 @@ incus config device set <instance_name> <device_name> security.acls.default.ingr
 When using network ACLs with a bridge network, be aware of the following limitations:
 
 - Unlike OVN ACLs, bridge ACLs are applied only on the boundary between the bridge and the Incus host.
-  This means they can only be used to apply network policies for traffic going to or from external networks.
+  This means they can only be used to apply network policies for traffic going to or from external networks (see exception for `nftables` firewall driver below).
   They cannot be used for to create {spellexception}`intra-bridge` firewalls, thus firewalls that control traffic between instances connected to the same bridge.
+- When using the `nftables` firewall driver you can apply ACLs to the NIC device and control traffic between the instances. In this case the `reject` ACL rules are not permitted and when the default action is set to `reject` it is interpreted as `drop`.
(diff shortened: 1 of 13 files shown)


stgraber:
> Since the nftables do not support "reject" rules, the implementation converts the default rules from "reject" to "drop" when needed.

I'm confused by that comment, nftables does have a `reject` target, what do you mean by that?

stgraber:
@mikerobski any chance that you can open PRs from a personal fork?

That would then let me push rebases and tweaks on top of your change to tweak it until it's in shape for merging, avoiding needless back and forth for tiny changes to your PR.

stgraber:
@mikerobski And sorry for the delay on this one. It's been a bit of a crazy time on my end what between getting married and a whole bunch of travel. I somehow remembered this PR being more complex than it is now, possibly because of a previous version of it. As it stands, this is well structured and pretty minimal so we should be able to get this into shape and merged pretty quick.

It looks like I'll want to do a few small edits to the doc, add an API extension string and some other minor changes, but outside of the `drop` vs `reject` thing, this looks well structured and almost good to go.

mikerobski:
> > Since the nftables do not support "reject" rules, the implementation converts the default rules from "reject" to "drop" when needed.
> 
> I'm confused by that comment, nftables does have a `reject` target, what do you mean by that?

The ACL rules we add are L2 and reject nft rule is not being accepted for the bridge family.

stgraber:
> The ACL rules we add are L2 and reject nft rule is not being accepted for the bridge family.

Ah, right, makes sense when processed at the L2 level.

mikerobski:
> @mikerobski any chance that you can open PRs from a personal fork?
> 
> That would then let me push rebases and tweaks on top of your change to tweak it until it's in shape for merging, avoiding needless back and forth for tiny changes to your PR.

I will look into it, but it may take me some time to figure out.

stgraber:
> Ah, right, makes sense when processed at the L2 level.

```
tgraber@dakara:~/data/code/lxc/incus (pr1225)$ sudo nft delete rule bridge incus in.u1.eth0 handle 34
stgraber@dakara:~/data/code/lxc/incus (pr1225)$ sudo nft insert rule bridge incus in.u1.eth0 iifname "veth078cc364" ip daddr 1.1.1.1 reject
stgraber@dakara:~/data/code/lxc/incus (pr1225)$ 
```

This works perfectly fine here and results in the expected:
```
root@u1:~# ping 1.1.1.1
PING 1.1.1.1 (1.1.1.1) 56(84) bytes of data.
From 1.1.1.1 icmp_seq=1 Destination Port Unreachable
From 1.1.1.1 icmp_seq=2 Destination Port Unreachable
^C
--- 1.1.1.1 ping statistics ---
2 packets transmitted, 0 received, +2 errors, 100% packet loss, time 1022ms

root@u1:~# 
```

mikerobski:
> > Ah, right, makes sense when processed at the L2 level.
> 
> ```
> stgraber@dakara:~/data/code/lxc/incus (pr1225)$ sudo nft delete rule bridge incus in.u1.eth0 handle 34
> stgraber@dakara:~/data/code/lxc/incus (pr1225)$ sudo nft insert rule bridge incus in.u1.eth0 iifname "veth078cc364" ip daddr 1.1.1.1 reject
> stgraber@dakara:~/data/code/lxc/incus (pr1225)$ 
> ```
> 
> This works perfectly fine here and results in the expected:
> 
> ```
> root@u1:~# ping 1.1.1.1
> PING 1.1.1.1 (1.1.1.1) 56(84) bytes of data.
> From 1.1.1.1 icmp_seq=1 Destination Port Unreachable
> From 1.1.1.1 icmp_seq=2 Destination Port Unreachable
> ^C
> --- 1.1.1.1 ping statistics ---
> 2 packets transmitted, 0 received, +2 errors, 100% packet loss, time 1022ms
> 
> root@u1:~# 
> ```

While indeed it works for input, it does not work for forward:
```
table bridge incus {
        chain fwd.vm01.eth0 {
                type filter hook forward priority filter; policy accept;
                ip daddr 1.1.1.1 counter packets 0 bytes 0 drop
                ip daddr 1.1.1.1 counter packets 0 bytes 0 drop
        }

        chain in.vm01.eth0 {
                type filter hook input priority filter; policy accept;
                ip daddr 1.1.1.1 counter packets 0 bytes 0 drop
                ip daddr 1.1.1.1 counter packets 0 bytes 0 reject with icmp port-unreachable
        }
}
```
```
nft add rule bridge incus fwd.vm01.eth0 ip daddr 1.1.1.1 counter reject
Error: Could not process rule: Operation not supported
add rule bridge incus fwd.vm01.eth0 ip daddr 1.1.1.1 counter reject
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
It will be confusing to allow the reject rule when it is not supported for the forward chain, right?

stgraber:
When is the forward table being hit in the bridge case?
Is that for connections between interfaces on the same bridge?

Btw, I just noticed a bit of a problem when testing this branch locally here.
If I create two containers and only apply an ACL to one of them:
```
egress:
- action: allow
  state: enabled
- action: drop
  destination: 1.1.1.1
  state: enabled
ingress: []
```

Then both containers can interact with the internet just fine, the container with no ACL applied can interact with all other containers except for the one with the ACL applied. So far that's normal.

What's not normal is that the container with the ACL applied is completely unable to interact with the container which doesn't have the ACL applied even though the restricted container has an rule allowing all egress.

Trying to ping from the container with the ACL to the one without only shows ARP going out and being responded to, no actual ICMP packet, so that'd suggest that the ARP response isn't making it back to the container with the ACL.

mikerobski:
@stgraber the nftemplate was adjusted to support the setup you were testing and some additional edge cases.

stgraber:
Confirmed that the branch fixes the instance to instance communication issue.

stgraber:
> When is the forward table being hit in the bridge case?
> Is that for connections between interfaces on the same bridge?

Can you respond to those?

mikerobski:
> > When is the forward table being hit in the bridge case?
> > Is that for connections between interfaces on the same bridge?
> 
> Can you respond to those?

Yes, forward table is used for instances with interfaces on the same bridge.

stgraber:
Okay, I think I'd be fine with supporting rejects and just converting them to drop in the forward table.
We'll obviously want a documentation note about this behavior, but that should be fine.

The reality is that 90% of the time folks are going to care about ingress/egress to things outside of the bridge, so having reject support on that would be very good as drop on egress is a pretty brutal behavior that will often cause issues.

mikerobski:
> Okay, I think I'd be fine with supporting rejects and just converting them to drop in the forward table. We'll obviously want a documentation note about this behavior, but that should be fine.
> 
> The reality is that 90% of the time folks are going to care about ingress/egress to things outside of the bridge, so having reject support on that would be very good as drop on egress is a pretty brutal behavior that will often cause issues.

The implementation does the conversion of reject to drop only for the default ACL rules. The rest of the rules are inspected and using a `reject` action will result in an error "Invalid action reject for bridge filter".
This behavior is described in the `network_acls.md` edit.

stgraber:
> The implementation does the conversion of reject to drop only for the default ACL rules. The rest of the rules are inspected and using a `reject` action will result in an error "Invalid action reject for bridge filter".
> This behavior is described in the `network_acls.md` edit.

Yeah, I know. What I'm saying is that given that `reject` only needs to be turned into a `drop` when put in the forward table (instance to instance communication), I'd like to still allow for `reject` as an action and just internally turn those into `drop` when generating the forward entries.

Then the documentation should just mention that when using ACLs with `nft`, instance to instance traffic within the same network is dropped rather than rejected due to a nft limitation.

And so for everything else, which should be the very vast majority of rules, `reject` will work as expected.

stgraber:
I'm looking at this one now. I've got a bunch of smaller changes, mostly rebasing and re-titling things. I also added an extra check in one of the functions so we don't try to update instances that are on another cluster member when updating a bridge ACL.

The current logic with the `force_update` flag has quite a few problems:
 - It relies on the instance having a local device for that network (not common, most use a profile)
 - It triggers a very costly full instance config Update
 - Because updates send out /dev/incus notifications, the guest will be notify of a change despite not seeing any actual change in the configuration

So I'm looking at a better way to directly hit the device update codepath for the device we care about without having to go through the rest of the logic.

PR #1228, opened 2024-09-18
Allow VMs to pull console history similar to containers

gibmat:
This adds the capability for VMs to fetch console history in a similar manner to containers (`incus console --show-log <instance>`).

The console backend device is changed to be a ring buffer by default, which we can query for history since the VM started up. When an interactive console is needed (ie, running `incus console <instance>`) the ring buffer is swapped out for a socket, and when done is replaced again by the ring buffer.

### DIFF
--- a/cmd/incusd/instance_console.go
+++ b/cmd/incusd/instance_console.go
@@ -217,7 +217,18 @@ func (s *consoleWs) doConsole(op *operations.Operation) error {
 		return err
 	}
 
-	defer func() { _ = console.Close() }()
+	// Cleanup the console when we're done.
+	defer func() {
+		_ = console.Close()
+
+		// If this was a text-based console for a VM, restore the default ring buffer device backend.
+		if s.instance.Type() == instancetype.VM && s.protocol == instance.ConsoleTypeConsole {
+			v, ok := s.instance.(instance.VM)
+			if ok {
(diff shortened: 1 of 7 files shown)


cmd/incusd/instance_console.go
```diff
@@ -217,7 +217,17 @@ func (s *consoleWs) doConsole(op *operations.Operation) error {
 		return err
 	}
 
-	defer func() { _ = console.Close() }()
+	// Cleanup the console when we're done.
+	defer func() {
+		_ = console.Close()
```
stgraber, line 222: Nit, I'd do a line break after this line and before the comment below.

internal/server/instance/drivers/driver_qemu.go
```diff
@@ -9307,3 +9322,40 @@ func (d *qemu) postCPUHotplug(monitor *qmp.Monitor) error {
 
 	return nil
 }
+
+// ConsoleLog returns all output sent to the instance's console's ring buffer since startup.
+func (d *qemu) ConsoleLog() (string, error) {
```
stgraber, line 9348: We may want to use an `operationlock` for this one to prevent two concurrent attempts to fetch the console leading to a corrupted `console.log`.

This would need a new operation lock type to be added but that should still be pretty trivial.

PR #1230, opened 2024-09-19
Properly handle `--project` in CLI error messages

bensmrs:
This PR fixes a problem where, in case of an error, the CLI would prompt the user to run an invalid command if they had specified a project other than the default one.

Before:
```
#> incus start debian --project test
Error: Failed to run: <...>
Try `incus info --show-log debian` for more info
```

After:
```
#> incus start debian --project test
Error: Failed to run: <...>
Try `incus info --show-log debian --project test` for more info
```

### DIFF
--- a/cmd/incus/action.go
+++ b/cmd/incus/action.go
@@ -331,7 +331,12 @@ func (c *cmdAction) doAction(action string, conf *config.Config, nameArg string)
 	err = cli.CancelableWait(op, &progress)
 	if err != nil {
 		progress.Done("")
-		return fmt.Errorf("%s\n"+i18n.G("Try `incus info --show-log %s` for more info"), err, nameArg)
+		projectArg := ""
+		if conf.ProjectOverride != "" && conf.ProjectOverride != api.ProjectDefaultName {
+			projectArg = " --project " + conf.ProjectOverride
+		}
+
+		return fmt.Errorf("%s\n"+i18n.G("Try `incus info --show-log %s%s` for more info"), err, nameArg, projectArg)
 	}
 
 	progress.Done("")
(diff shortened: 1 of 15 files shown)


bensmrs:
Quick warning: the i18n strings should now contain an additional `%s`, which needs to be taken into account by translators

cmd/incus/action.go
```diff
@@ -331,7 +331,12 @@ func (c *cmdAction) doAction(action string, conf *config.Config, nameArg string)
 	err = cli.CancelableWait(op, &progress)
 	if err != nil {
 		progress.Done("")
-		return fmt.Errorf("%s\n"+i18n.G("Try `incus info --show-log %s` for more info"), err, nameArg)
+		projectArg := ""
+		if conf.ProjectOverride != "" && conf.ProjectOverride != "default" {
+			projectArg = " --project " + conf.ProjectOverride
+		}
```
stgraber, line 337: I'd use `api.ProjectDefaultName` instead of the hardcoded `"default"` in here.

cmd/incus/launch.go
```diff
@@ -143,7 +143,12 @@ func (c *cmdLaunch) Run(cmd *cobra.Command, args []string) error {
 			prettyName = fmt.Sprintf("%s:%s", remote, name)
 		}
 
-		return fmt.Errorf("%s\n"+i18n.G("Try `incus info --show-log %s` for more info"), err, prettyName)
+		projectArg := ""
+		if conf.ProjectOverride != "" && conf.ProjectOverride != "default" {
```
stgraber, line 147: Same as above

cmd/incus/action.go
bensmrs: Noted. FYI, `"default"` is hardcoded in many places in `cmd/`.

stgraber: Yeah, it's one of those where we're trying to not introduce new ones at least :)

bensmrs:
Should be fine now

stgraber (approved):

PR #1232, opened 2024-09-20
doc: add note for Nvidia usage when building from source in OpenSUSE

javiertoledos:
Using Nvidia GPU when building from source in OpenSUSE requires some extra notes that are missing.

1. LXC hooks (which includes Nvidia hooks) are missing in OpenSUSE unless `lxc` package is installed
2. Nvidia container tools is required but not mentioned anywhere (this could be moved/added in another section in the future).


stgraber:
Hello,

Before we can review this pull request, we need all the commits to include a suitable `Signed-off-by` line. You can see CONTRIBUTING.md for more information on that.

Thanks!

stgraber:
Thanks!

PR #1234, opened 2024-09-20
Add operating system details to instance state

gibmat:
This adds operating system details to an instance's state. `incus-agent` has been extended to populate this information when running within a VM.

```
$ incus info debian
Name: debian
Status: RUNNING
Type: virtual-machine
Architecture: x86_64
PID: 658181
Created: 2024/09/13 16:13 MDT
Last Used: 2024/09/20 11:59 MDT
Started: 2024/09/20 11:59 MDT

Operating System:
  OS: Debian GNU/Linux
  OS Version: 12 (bookworm)
  Kernel Version: 6.1.0-25-amd64
  Hostname: debian
  FQDN: debian.testing.local

Resources:
  Processes: 16
[snip]
```

### DIFF
--- a/shared/api/instance_state.go
+++ b/shared/api/instance_state.go
@@ -65,6 +65,11 @@ type InstanceState struct {
 	//
 	// API extension: instance_state_started_at.
 	StartedAt time.Time `json:"started_at" yaml:"started_at"`
+
+	// OS information.
+	//
+	// API extension: instances_state_os_info.
+	OSInfo *InstanceStateOSInfo `json:"os_info" yaml:"os_info"`
 }
 
 // InstanceStateDisk represents the disk information section of an instance's state.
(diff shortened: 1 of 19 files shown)


stgraber:
Looks like this will need a rebase to fix the i18n conflicts.

The first commit should be split up a bit into:
 - api: instances_state_os_info
 - shared/api: Add OSInfo to InstanceState
 - doc/rest-api: Refresh swagger YAML

I also think it may be cleaner to make `OSInfo` a pointer rather than an attached struct.
With it being a pointer, the CLI can then just do a `OSInfo != nil` and then print all the fields (showing them empty if we don't have them). This will save the per-field checks and will provide more consistent output when the agent is running.

shared/api/instance_state.go
```diff
@@ -220,3 +220,30 @@ type InstanceStateNetworkCounters struct {
 	// Example: 179
 	PacketsDroppedInbound int64 `json:"packets_dropped_inbound" yaml:"packets_dropped_inbound"`
 }
+
+// InstanceStateOSInfo represents the operating system information section of an instance's state.
+//
+// swagger:model
+//
+// API extension: instances_state_os_info.
+type InstanceStateOSInfo struct {
```
stgraber, line 234: This change should be in the `shared/api: Add OSInfo to InstanceState` commit.

API extension commits should only include `doc/api-extensions.md` and `internal/version/api.go` changes.

stgraber:
Looks good, just one little bit of code to move between commits and we'll be good to go!

PR #1235, opened 2024-09-21
Refresh Github Actions environment

stgraber:


stgraber:
As the failure appeared out of nowhere and there has been no obvious change to resize2fs, the current assumption is that this is kernel related, so trying to switch Github jobs to a more recent Ubuntu and kernel.

stgraber:
https://bugs.launchpad.net/ubuntu/+source/linux/+bug/2081231

stgraber:
This PR is still useful on its own, improves error reporting a bit and updates our Github Actions stuff to 24.04.

Sadly the kernel issue is present on both 22.04 and 24.04, so that's not actually helping us here.
We'll just have to ignore LVM/Ceph failures for a bit.

PR #1260, opened 2024-09-27
docs: mention alpine edge repository in alpine installation

thebigbone:
The alpine community repository `(http://dl-cdn.alpinelinux.org/alpine/v3.15/community)` does not contain the `incus` packages. It's only available in `http://dl-cdn.alpinelinux.org/alpine/edge/community` mirror (which is not enabled by default).

### DIFF
--- a/doc/.wordlist.txt
+++ b/doc/.wordlist.txt
@@ -297,6 +297,7 @@ UIDs
 unconfigured
 unevictable
 unixgram
+uncomment
 unmanaged
 unmount
 unmounting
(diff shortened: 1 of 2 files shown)


thebigbone:
Alpine has a different init system so I was getting some errors regarding systemd. I saw one issue which instructed to create `/sys/fs/cgroup/systemd` directory and then mount it with `mount -t cgroup -o none,name=systemd systemd /sys/fs/cgroup/systemd` after which I was able to run `incus`. Should I add this in the docs?

stgraber:
Yeah, you can send a PR to add that or add a link to a doc page on the alpine wiki or doc site if there is one (less likely to get outdated on our side).

stgraber (approved):

PR #1262, opened 2024-09-27
Run cluster evacuate and restore in parallel

gibmat:
Sequentially evacuating or restoring a server can be slow when there are several instances running. This switches to running migrations in parallel, based on the number of detected CPUs available.

### DIFF
--- a/cmd/incusd/api_cluster_evacuation.go
+++ b/cmd/incusd/api_cluster_evacuation.go
@@ -169,84 +169,87 @@ func evacuateInstances(ctx context.Context, opts evacuateOpts) error {
 	group.SetLimit(numParallelEvacs)
 
 	for _, inst := range opts.instances {
-		// Accesses variables groupCtx, inst, and opts
 		group.Go(func() error {
-		metadata := make(map[string]any)
-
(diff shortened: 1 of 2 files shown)


cmd/incusd/api_cluster_evacuation.go
```diff
@@ -161,9 +162,17 @@ func evacuateInstances(ctx context.Context, opts evacuateOpts) error {
...
+	// Limit the number of concurrent evacuations to run at the same time
+	numParallelEvacs := max(runtime.NumCPU()/1, 1) // FIXME after testing, bump NumCPU division to 8 or 16
+
+	group, groupCtx := errgroup.WithContext(ctx)
+	group.SetLimit(numParallelEvacs)
 
 	for _, inst := range opts.instances {
+		// Accesses variables groupCtx, inst, and opts
+		group.Go(func() error {
+		metadata := make(map[string]any) // FIXME indent code block after review (makes diff noisy)
```
gibmat, line 174: Either indent or move to new function before merging (creates a lot of diff noise).

stgraber (approved):

Issue #1176, opened 2024-09-02
Make use of SFTP API for `incus file` operations


stgraber:
Currently all the `incus file` operations are done through the individual REST API endpoints, which Incus then internally converts to SFTP interactions. This isn't exactly the most flexible or efficient.

It'd be good to change all the `incus file` operations to instead use `GetInstanceFileSFTP` to fetch a Go SFTP client and then perform the relevant filesystem operations through it.

- [x] create
- [ ] delete
- [x] push
- [x] pull
- [ ] edit

HassanAlsamahi:
Can I work on this? If that is okay, should the change be done in client/incus_instances and replace the REST API part of the functions CreateInstanceFile..etc with the SFTP part? or directly in the cmd/incus/files? I believe the first but wanted to verify.

stgraber:
Assigned it to you!

The change should be done directly in `cmd/incus/file.go`, the Go client package should be a mostly clear mapping to the API so we don't want it to get too smart on that front.

stgraber:
Hey @HassanAlsamahi, how are things going with this one?

HassanAlsamahi:
Hey @stgraber . I made some progress but I didn't push a commit yet as I got busy and some personal stuff. I'll push a commit very soon for one of the operations, followed by the other operations. Thanks for asking and sorry for the delay.

stgraber:
No worries, thanks for the update!

stgraber:
@HassanAlsamahi any luck with the other ones?

HassanAlsamahi:
Hey @stgraber . Sorry I was on vacation in October totally logged off. Please check the latest PR and let me know if anything is needed regarding it.

stgraber:
Thanks!

HassanAlsamahi:
Hey @stgraber , I have a question. Does the functions `recursivePullFile` and `recursivePushFile` need also to be modified to use the SFTP package instead of `buf, resp, err := d.GetInstanceFile(inst, p)` and `err = d.CreateInstanceFile(inst, targetPath, args)`  respectively?

stgraber:
Hey @HassanAlsamahi, how's it going with this one?

HassanAlsamahi:
Hey @stgraber . It's getting so busy in here, I had the code updated for the pull operation but didn't get the chance to organize, validate, and commit until now. Please check the latest PR and let me know if any modifications needed.

Pending now is
- [ ] edit
- [ ] recursivePullFile
[...]

PR #1266, opened 2024-09-30
Change incus file delete operation to sftp

HassanAlsamahi:
This is the modification of ```incus file delete``` as a start. Please let me know if the implementation is acceptable and I'll proceed with the other operations.

Note: I noticed that it throws an error if trying to delete a directory that includes contents, thus, I've added ```--force``` flag in case the user still wants to delete a directory with all its contents. Please let me know if that is not required I can remove it.

### DIFF
--- a/cmd/incus/file.go
+++ b/cmd/incus/file.go
@@ -18,6 +18,7 @@ import (
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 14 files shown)


HassanAlsamahi:
relates to #1176

stgraber (approved):

Issue #1261, opened 2024-09-27
When limits.memory is set to a percentage, an error occurs when starting a virtual machine


itviewer:
OS: Ubuntu noble 24.04 x86_64
Kernel: Linux 6.8.0-45-generic
incus version 6.5

```
config:
  limits.memory: 50%
```
When set to a percentage, the virtual machine cannot be started, but the system container can run normally.
```
$ incus start ubuntu2204vm
Error: limits.memory invalid: Invalid value: 50%
```

PR #1270, opened 2024-10-01
Allows the VM's limits.memory configuration to be set to a percentage value

itviewer:
fix https://github.com/lxc/incus/issues/1261

Just extract the logic of driver_lxc.go for the memory percentage value into a function and apply it to driver_qemu.go


stgraber (approved):

stgraber:
Thanks!

PR #1274, opened 2024-10-02
doc/howto: Clarify support for QCOW2 and VMDK images

simos:
Clarify which versions of incus-migrate directly support non-raw images.


stgraber:
We generally don't want our documentation to refer to past versions, it should only cover what's currently there.

Especially in this case as `incus-migrate` is a command that one can easily retrieve binaries of from the current Github Actions or Github release for Incus.

We will probably want to spend the time to get our doc website to let you access documentation for specific releases, but we haven't had time to get that done yet.

simos:
I added this patch after the discussion at https://discuss.linuxcontainers.org/t/qcow2-will-not-boot/21700/3?u=simos

Should I have just used the version that Brian wrote in commend 4?

PR #1278, opened 2024-10-02
Allow fully creating LVM cluster pools through the API

### LINKED ISSUE #1065: Allow fully creating LVM cluster pools through the API
Currently for LVM cluster pools, you need to provide an existing shared VG name.
This effectively prevents setting them up fully remotely through the API.

Instead we should be able to provide the name of the shared block device as the source and have Incus create the volume group automatically on the leader and then enable it on all the other servers.

gibmat:
Closes #1065


stgraber (approved):

stgraber:
Added an API extension for this.

PR #1298, opened 2024-10-09
incusd/internal/server/instance/drivers:  support for Chimera Linux (qemu/edk2) pkg layout

mwyvr:
Incus is packaged for Chimera Linux; this addition provides support for edk2 filenames as packaged by Chimera Linux (these are the base filenames supplied by the qemu/edk2 package https://github.com/qemu/edk2).

Have built and tested on x86_64 arch / Chimera Linux (musl libc); both containers and virtual machines (security.secureboot=false) perform as expected.

### DIFF
--- a/internal/server/instance/drivers/edk2/driver_edk2.go
+++ b/internal/server/instance/drivers/edk2/driver_edk2.go
@@ -72,13 +72,16 @@ var architectureInstallations = map[int][]Installation{
 			GENERIC: {
 				{Code: "ovmf-x86_64-4m-code.bin", Vars: "ovmf-x86_64-4m-vars.bin"},
 				{Code: "ovmf-x86_64.bin", Vars: "ovmf-x86_64-code.bin"},
+				{Code: "edk2-x86_64-code.fd", Vars: "edk2-i386-vars.fd"},
 			},
 			SECUREBOOT: {
 				{Code: "ovmf-x86_64-ms-4m-vars.bin", Vars: "ovmf-x86_64-ms-4m-code.bin"},
 				{Code: "ovmf-x86_64-ms-code.bin", Vars: "ovmf-x86_64-ms-vars.bin"},
+				{Code: "edk2-x86_64-secure-code.fd", Vars: "edk2-i386-vars.fd"},
 			},
 			CSM: {
 				{Code: "seabios.bin", Vars: "seabios.bin"},
+				{Code: "bios.bin", Vars: "bios.bin"},
 			},
 		},
 	}, {


mwyvr:
PS, I can do a PR for https://linuxcontainers.org/incus/docs/main/installing/ but will provide them here too.

I'd prefer holding off submitting a PR for install docs because this particular PR is key to making Incus properly functional without manual workarounds on Chimera Linux, and, with this change having been accepted, hope/expect to see Incus move from the Chimera user to main repository.

Install:

To install Incus on Chimera Linux, ensure the `user` repository has been added to the system:

    apk add chimera-repo-user

Then add the `incus` package; this will install other dependencies including `incus-client`. Enable the service.

    apk add incus
    dinitctl enable incus

To run virtual machines, also add the EDK2 firmware. Chimera Linux currently does not provide complete support for Secure Boot, so virtual machines must be launched with this feature disabled.

    apk add qemu-edk2-firmware
    # example: launch virtual machine with secureboot disabled:
    # incus launch images:debian/12 --vm -c security.secureboot=false

Report packaging issues [here](https://github.com/chimera-linux/cports/issues).

hallyn (approved):

hallyn:
Thanks, looks good to me.

I do have to ask though - what would be the chances of just having the incus package in chimera linux add symlinks to the names incus already knows them by?

Also, I wonder if we should start adding comments to each line to say which distros they are intended to support

hallyn:
Interesting, SOB check wants the same username that shows up in From.  I didn't know it checked for that...

mwyvr:
On Fri, 2024-10-11 at 12:32 -0700, Serge Hallyn wrote:
> Thanks, looks good to me.
> I do have to ask though - what would be the chances of just having
> the incus package in chimera linux add symlinks to the names incus
> already knows them by?
Indeed, that's what I had been doing, manually. I rationalized
submitting these changes as Chimera Linux maintainers have used the
package/filenames/descriptors from the qemu package, so at least they
will match if other distributions have or will follow suit.

https://github.com/qemu/qemu/tree/05adb38839ba656c7383a548b460d95c91e2febe/pc-bios
> Also, I wonder if we should start adding comments to each line to say
> which distros they are intended to support
Seems like a good idea; after looking at Ubuntu/Debian, Fedora, Arch
and openSUSE (which should package Incus but doesn't to my
disappointment) they are all rather unique combos of paths and EDK2
filenames.

mwyvr:
> Interesting, SOB check wants the same username that shows up in From. I didn't know it checked for that...

What's the right way to correct this? Both email addresses are deliverable to me, btw. I'll change my github setup to git@.

hallyn:
> Indeed, that's what I had been doing, manually. I rationalized submitting these changes as Chimera Linux maintainers have used the package/filenames/descriptors from the qemu package, so at least they will match if other distributions have or will follow suit.

Oh, that seems reasonable.

hallyn:
> > Interesting, SOB check wants the same username that shows up in From. I didn't know it checked for that...
> 
> What's the right way to correct this? Both email addresses are deliverable to me, btw; since this I had github verify the github@. I'll ensure I've used git@mike... in the future, the commit comment was a typo on my part.

Well the easiest way for you is probably to just make the SOB line be

Signed-off-by: Mike Watkins <git@mikewatkins.ca>

Rather than do a git commit --amend --author=...github...

stgraber (approved):
Very weird that Chimera doesn't have an x86_64 version of the vars file.

mwyvr:
> Very weird that Chimera doesn't have an x86_64 version of the vars file.
I thought so too, and went looking for an alternative;  Chimera uses the edk2 contained within the qemu release; in their code I see they define an x86_64 descriptor using the i386 vars file.

https://github.com/qemu/qemu/blob/f1dd640896ee2b50cb34328f2568aad324702954/pc-bios/descriptors/60-edk2-x86_64.json#L13

PR #1318, opened 2024-10-18
incus: Fix display of current project in projects list

montag451:
The current project is not indicated when listing projects with `incus project ls` when the locale is not an english one.

/rant mode on
I really love incus but I don't think translating it was a good idea. Most languages has partial translations and even worth, a lot of strings have incorrect translations which can be quite confusing to the point that I decided to create a `alias incus=LANG=C incus` to get rid of this problem.
/rant mode off


stgraber (approved):

stgraber:
Thanks!

About translations, it's always a very tricky thing and generally an issue on Linux at large.

Ideally you'd only start showing a translated UI once it's above 90% translated and when there's a group of people who can ensure some consistency in the translation. The problem is that it's very very hard to get to that 90% and it's even harder to get anyone to help with translation if no translation is shown at all.

So that's how you end up with the current Linux translation landscape where everything is partially translated and somewhat inconsistent and effectively relies on the fact that people will get annoyed enough by it to go and contribute the missing translation strings...

In the case of Incus, we really need one language to hit 90%+ just so we can do a proper review of the CLI and notice anything that's not properly exposed as translatable, anything that's not properly formatted and where a modified translation string would help and anything that's getting translated which really shouldn't.

Our great Japanese community is once again ahead of the curve here with 77% currently, so I'm hoping that we can make the translation experience better over the next few months and motivate people to help translating Incus.

PR #1329, opened 2024-10-21
Make ask library shared

gibmat:
The `ask` library can be useful beyond its current internal use. This makes it a shared library.


stgraber:
I'm surprised that our tests are passing when `NewAsker` seems to be missing a godoc comment.

PR #1338, opened 2024-10-25
Implement `incus webui`

stgraber:
This implements a new `incus webui` command which will:
 - Check if the server has the web UI installed
 - Spawn a tiny web server for HTTP requests on localhost
 - Generate an authentication token to avoid other clients on the machine from accessing Incus
 - Open the user's web browser with a URL including that token

The result is that running `incus webui` will open your current web browser to the Incus UI without requiring Incus itself to listen on the network, or have to deal with Incus' TLS authentication and TLS certificate verification.

### DIFF
--- a/client/incus_oidc.go
+++ b/client/incus_oidc.go
@@ -18,6 +18,8 @@ import (
 	httphelper "github.com/zitadel/oidc/v3/pkg/http"
 	"github.com/zitadel/oidc/v3/pkg/oidc"
 	"golang.org/x/oauth2"
+
+	"github.com/lxc/incus/v6/shared/util"
 )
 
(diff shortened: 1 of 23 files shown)


jarrodu:
I tested this locally and got the UI up and running in Firefox.

I get an error when I try to connect to the terminal of a running container.

The UI shows:

```
Error
The connection was closed due to a failure to perform a TLS handshake (e.g., the server certificate can't be verified).
```

The Incus remote is running on another machine. I compiled the project in a fresh system container running on the same machine. I then used the remote extension in VS Code to connect to the container.

If the feature works on your side, then I can try compiling the client locally on my laptop and trying again without the VS Code proxy.

Issue #1226, opened 2024-09-18
Addition of helper functions in the QEMU scriptlet


bensmrs:
Hi!
I'm making good progress with my black magic operations to get macOS working, thanks to the new QEMU scriptlet.
I’m finding myself defining plenty of functions like `qom_get`, `qom_set`,… which are critical to me. Would it be interesting to have them directly available to the scriptlets?
I know it will only serve 5 to 10 people, but if that sounds good to you, I’ll add them to the Incus codebase.

stgraber:
Yeah, that should be fine to add a few of the most common ones as exported functions for this scriptlet.

PR #1340, opened 2024-10-28
Export QMP functions

bensmrs:
Closes: #1226

### DIFF
--- a/doc/reference/instance_options.md
+++ b/doc/reference/instance_options.md
@@ -325,6 +325,59 @@ property = "disable_s4"
 value = "0"
 ```
 
+### Override QEMU runtime objects
+While `raw.qemu` and `raw.qemu.conf` can be used to alter the arguments
+and configuration file that's passed to QEMU, a lot of devices are now
+added through QMP instead.
+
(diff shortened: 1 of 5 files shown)


stgraber:
@bensmrs I wonder, wouldn't it be cleaner if we just added a single command, `run_command` which would take one un-named argument (the command) and treat all kwargs as the arguments?

So we'd then have `run_qmp` that handles the raw JSON in/out QMP and `run_command` that lets you use a nicer syntax like:

```
run_command("qom-get", path="/machine/peripheral/dev-incus_root", property="serial")
```

I'm basically trying to avoid us having to duplicate the entire QMP command list in here.

bensmrs:
I expected the remark but I agree with your last message, I’ll look at it tomorrow

stgraber:
Thanks!

bensmrs:
Just for a bit of context, I’m exploring the call stack to get the position in the Starlark code where there is a problem (in big scriptlets, it’s pretty useful). I haven’t found any other way, and Starlark doesn’t seem to offer much introspection into AST nodes, so that should do the trick…

stgraber (approved):

PR #1356, opened 2024-11-06
Add get-current to show current project

maveonair:
This pull requests adds a new command to show the current project.

```
$ incus project get-current
default

$ incus project create foo
$ incus project switch foo
$ incus project get-current
foo
```


stgraber:
I fixed the code test failure in place, so this should be good to go now.

stgraber (approved):

Issue #485, opened 2024-02-11
Automatic live-migration to balance load on cluster


stgraber:
When running a cluster with one or more virtual-machines that's capable of being live-migrated across the cluster, we should be able to use that to better spread the load across a cluster by evaluating server load across the cluster and deciding whether to automatically move some workloads to re-balance things.

We already have a lot of the right pieces in place:
 - Ability to live-migrate
 - Internal APIs to query load and hardware resources of the servers
 - Flexible scheduler (through scriptlet) to handle placement decisions

We will need to think through ways to avoid instances flip-flopping between servers as well as ways to mitigate the migration itself causing significant load difference on both the source and target server.

One approach would be to only perform a single migration per cycle, while also preventing an instance from being moved again for a number of cycles since its last move.

Ideally we'd be leveraging calls to our existing scheduler to find new locations for existing instances, only considering instances that can be easily live-migrated (no local storage, no local devices, ...).

dontlaugh:
> One approach would be to only perform a single migration per cycle, while also preventing an instance from being moved again for a number of cycles since its last move.

What does "cycle" refer to in this context?

stgraber:
A cycle would be a scheduled task inside Incus basically.

The admin would basically instruct Incus to consider automatic balancing every 15min or every hour or every 3 hours, whatever makes sense based on their environment.

sophiezhangg:
Hello!

My partners and I are currently studying Virtualization in UT Austin and would like to have more experience contributing to open-source repos like this - we are wondering if this issue is still open for solving? If so, we would love to take a chance and work on this because it seems very interesting to us! No worries if not!

Thank you so much!

sophiezhangg:
Hello @stgraber!

My team just had discussion about how to approach this ticket and we are wondering if you mind providing tips, suggestions, and/or clarifications on our approach?

Here is our general logic:

1. We think we will need to write a scriptlet for a custom instance_placement function - and the core idea is that given a request and a list of candidate members, we will loop through the candidate_members and acquire the server and its matching load / resource information. Then we will select the server that has the lowest number of instances currently running on it. After that, we see that if the request's reason field is relocation. If it is and the server load is lower than the current server load that the request is on, then we set the target location to that server. However, if the relocation reason is set and the server load is the same, then we don't do anything. **One question we have about this is that we saw mentions of the default scheduler scriptlet but we aren't able to find it code-wise - we are wondering if you know where it is / if you mind providing it?**
[...]

PR #1369, opened 2024-11-12
Automatic live-migration to balance load on cluster

presztak:
This PR adds feature to automatically move some workloads to re-balance load across cluster.
The logic works in following way:
- Spawns a background task which is performed on the current leader
- Calculates a per-server score (based on CPU and memory)
- Splits the servers by CPU architecture, then sort the servers by their score and check difference between most and least busy for each architecture, exit the re-balance if the difference is less than the threshold
- Pulls the list of instances from the most busy server and filter out those that cannot be live migrated
- Determines what instances should be migrated to equalize the score between the two servers looking at their current memory usage and CPU allocation and migrate them

Fixes: #485
Closes: #835

### DIFF
--- a/cmd/incus/monitor.go
+++ b/cmd/incus/monitor.go
@@ -151,7 +151,13 @@ func (c *cmdMonitor) Run(cmd *cobra.Command, args []string) error {
 
 			entry := &logrus.Entry{Logger: logger}
 			entry.Data = c.unpackCtx(record.Ctx)
-			entry.Message = record.Msg
+
+			if event.Type == "logging" && d.IsClustered() {
+				entry.Message = fmt.Sprintf("[%s] %s", event.Location, record.Msg)
+			} else {
+				entry.Message = record.Msg
+			}
+
 			entry.Time = record.Time
 			entry.Level = msgLevel
 			format := logrus.TextFormatter{FullTimestamp: true, PadLevelText: true}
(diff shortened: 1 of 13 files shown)


stgraber:
A few initial tweaks I've made:
 - Updated the API extension commit to list the new config keys
 - Fixed config/config.go to have `cluster.rebalance.frequency` get a gendoc entry
 - Re-generated gendoc

stgraber:
Going to bump the default cooldown to 6H as otherwise with a 1h frequency, it'd be a bit useless :)

stgraber:
I'm also making `cluster.rebalance.frequency` be `0` by default as we shouldn't do this by default.
And making the default threshold be `20`

stgraber:
I'm also making the threshold be between 10% and 100%

stgraber:
I'm renaming `frequency` to `interval` for consistency with other config keys.

stgraber:
Reducing the default batch size down to just 1 instance per batch as that's the most reliable, even if slowest. Also changing the batching to be per architecture rather than global.

stgraber (approved):

stgraber:
I did a few more tweaks on top of what I commented on before:
 - Extracted the instance resource logic we had in the placement logic and moved it to the instance package
 - Rebased the logic in the placement scriptlet to use that function
 - Replaced the logic in this PR with the shared logic
 - Renamed the main .go file to match the pattern for the clustering related files
 - Renamed some functions to keep them a bit better namespaced toegether
 - Optimized the migration code a bit

I've had it running for a bit on a test cluster now and it seems to be behaving as intended.

Issue #1028, opened 2024-07-20
Make `incus top` output configurable through options


stgraber:
Splitting that one off of #1026 

Currently `incus top` can only be configured through shortcuts.

It'd be good to have it support our standard `--format` and `--columns` options as well as a `--refresh` to configure the refresh delay.

Eglantinee:
Hi @stgraber,
I am new to open source and would like to take my first steps by working on this issue.
Can you assign this issue to me?
TIA.

stgraber:
Done!

Eglantinee:
@stgraber 
I have a question from the beginning: please clarify several things about the new keys:
* refresh - Should this be an int flag or time.Duration?
* columns - I assume this is a list of columns to be shown from the possible ones. Am I right? BTW, do you want to support some ordering for the columns?
* format - I can only guess what this is (format for output values in the columns?). Can you please provide me with an example, or if it is some standard behavior, where it is used, so I can use it as a reference?

TIA.

stgraber:
`flagRefresh` should be an `int` with a default value of 10.
`flagColumns` should be similar to what we do in commands like `incus list` though at the beginning we'd just stick with the set of columns we have right now but make it possible show/hide them and indeed to re-order them.
`flagFormat` should behave like our other `list` commands with the exception that for `incus top`, we'd likely just allow `table` and `compact` as the other formats (`csv`, `json` and `yaml`) don't make as much sense for an interactive tool.

So basically we should be able to do:
`incus top --refresh=30 --all-projects --columns=enDum --format=compact` and get a live updated table that looks like:
```
  PROJECT  NAME          DISK        CPU        MEMORY
  default  incus-ui      1.27GiB     20s        80.78MiB      
  default  kernel-test   5.12GiB     186s       1.17GiB       
  default  ovn           2.94GiB     418s       1.35GiB       
  default  speedtest     294.01MiB   21s        90.04MiB      
  vpn      vpn-foo       153.17MiB   32s        48.89MiB      
  vpn      vpn-bar       960.43MiB   1481s      1.67GiB       
  vpn      vpn-baz       101.76MiB   18s        37.25MiB      
```

stgraber:
Hey @Eglantinee, how is it going with this one?

stgraber:
Unassigning due to inactivity so others can have a go at it.

PR #1370, opened 2024-11-13
Make `incus top` output configurable through options

presztak:
Adds support for following options:
- `--format` - possibility to define output format
- `--columns` - possibility to define column layout
- `--refresh` - possibility to configure the refresh delay

Closes: #1028

### DIFF
--- a/cmd/incus/top.go
+++ b/cmd/incus/top.go
@@ -2,9 +2,11 @@ package main
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"os"
 	"regexp"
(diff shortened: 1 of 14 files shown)


stgraber:
Looks good, just did a few minor edits:
 - Added validation for `--format` so it only allows `table` and `compact` (along with their extra options) as otherwise `--format=json` was considered valid
 - Added validation for `--refresh` to make the minimum refresh rate 10s (matches server side minimum for metrics)
 - Updated a comment (was mentioning making --refresh a flag, now it has been)

PR #1380, opened 2024-11-15
exec: Consume websocket pings for stderr

stefanor:
Fixes: #1289

### DIFF
--- a/cmd/incusd/instance_exec.go
+++ b/cmd/incusd/instance_exec.go
@@ -478,6 +478,15 @@ func (s *execWs) Do(op *operations.Operation) error {
 					}()
 				}
 
+				if i == execWSStderr {
+					// Consume data (e.g. websocket pings) from stderr too to
+					// avoid a situation where we hit an inactivity timeout on
+					// stderr during long exec sessions
+					go func() {
+						_, _, _ = conn.ReadMessage()
+					}()
+				}
+
 				if i == execWSStdin {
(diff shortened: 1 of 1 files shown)


stgraber:
@stefanor could you update the commit message to include the `Signed-off-by` line?

cmd/incusd/instance_exec.go
```diff
@@ -478,6 +478,13 @@ func (s *execWs) Do(op *operations.Operation) error {
 					}()
 				}
 
+				if i == execWSStderr {
+					// Consume websocket pings
```
stgraber, line 482: I'd probably expand a bit on the comment so we don't risk removing this :)

```
// Consume data from stderr too to avoid a situation where we hit an inactivity timeout on stderr during long exec sessions.
```

stgraber:
Very nice catch. I had started to look into this a few weeks ago, but the time needed to reproduce and debug was a bit annoying :)

stefanor:
> Very nice catch. I had started to look into this a few weeks ago, but the time needed to reproduce and debug was a bit annoying :)

@cjwatson pointed me in the right direction, because he'd debugged something similar before. Once I knew what to look for, it was obvious what was missing.

stgraber:
Thanks!

stefanor:
Would you mind cherry-picking this into the stable-6.0 branch?

PR #1381, opened 2024-11-15
incus-simplestreams: Add prune command

### LINKED ISSUE #1303: incus-simplestreams : missing prune features
`incus-simplestreams` is a nice tool, but it's missing two features :
* Removal of orphan image tarballs (not referenced by `images.json`) -> This might be triggered by a `incus-simplestreams add` with an image with identical properties and same date, but different checksum
* Removal of old `version`s of a `product` : When publishing a new image, the old one stays and we need manual tooling to clean it up.

I implemented those two features in a python script : https://github.com/YunoHost/lxd_img_builder/blob/master/prune_incus_simplestreams.py but it would be really nice to have a `incus-simplestreams prune` command that does both jobs.

EDIT: considering the lack of those features, i'm wondering about the intended workflow : maybe it would be better to have a "staging" repository that i'd fill with new versions of the images and then swap the real repo and the staging repo…

presztak:
This PR adds a prune command to incus-simplestreams which:

- Removes orphaned image tarballs (those not referenced by images.json)
- Finds any broken image and deletes them from the index
- Removes older versions of a product, retaining only the most recent one

Closes #1303

### DIFF
--- a/cmd/incus-simplestreams/main_prune.go
+++ b/cmd/incus-simplestreams/main_prune.go
@@ -1,0 +1,217 @@
+package main
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"slices"
+	"sort"
+
+	"github.com/spf13/cobra"
+
+	cli "github.com/lxc/incus/v6/internal/cmd"
+	"github.com/lxc/incus/v6/shared/simplestreams"
+)
+
+type cmdPrune struct {
+	global *cmdGlobal
+
+	flagDryRun    bool
+	flagRetention int
+	flagVerbose   bool
+}
+
+// Command generates the command definition.
+func (c *cmdPrune) Command() *cobra.Command {
+	cmd := &cobra.Command{}
+	cmd.Use = "prune"
+	cmd.Short = "Clean up obsolete files and data"
+	cmd.Long = cli.FormatSection("Description",
+		`Cleans up obsolete tarball files and removes outdated versions of a product
+
+The prune command scans the project directory for tarball files that do not have corresponding references
+in the 'images.json' file. Any tarball file that is not listed in images.json is considered orphaned
+and will be deleted.
+Additionally this command will delete older images, keeping a configurable number of older images per product.`)
+
+	cmd.RunE = c.Run
+	cmd.Flags().BoolVarP(&c.flagDryRun, "dry-run", "d", false, "Preview changes without executing actual operations")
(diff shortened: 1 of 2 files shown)


cmd/incus-simplestreams/main_prune.go
```diff
@@ -0,0 +1,167 @@
...
+	flagVerbose       bool
+}
+
+// Command generates the command definition.
+func (c *cmdPrune) Command() *cobra.Command {
+	cmd := &cobra.Command{}
+	cmd.Use = "prune"
+	cmd.Short = "Clean up obsolete files and data"
+	cmd.Long = cli.FormatSection("Description",
+		`Cleans up obsolote tarball files and removes outdated versions of a product
```
stgraber, line 32: Typo, should be `obsolete`

cmd/incus-simplestreams/main_prune.go
```diff
@@ -0,0 +1,167 @@
...
+
+The prune command scans the project directory for tarball files that do not have corresponding references
+in the 'images.json' file. Any tarball file that is not listed in images.json is considered orphaned
+and will be deleted.
+The prune command identifies and deletes older versions of product data,
+retaining only the most recent existing version.`)
```
stgraber, line 38: That last paragraph and the fact that the `prune` command has `prune-versions` flag is pretty confusing.

I think I'd rephrase that last paragraph as:
```

Additionally this command will delete older images, keeping a configurable number of older images per product.
```

Then instead of `prune-version` we should probably just have a `--retention` flag which defaults to `2` as in we'll keep two older versions of the product by default (same as we do on the public image server).

stgraber: So running `incus-simplestreams prune` should:
 - Find any orphaned files (no record in the index) and delete them
 - Find any broken image and delete them from the index
 - Find any product that has more than 1+flagRetention images and delete the older ones

presztak: What do you mean by `broken image`?

stgraber: That would be an image that's in the index but for which some of the files are missing on disk.

cmd/incus-simplestreams/main_prune.go
presztak: Fixed.

cmd/incus-simplestreams/main_prune.go
presztak: Done.

stgraber (approved):

PR #1384, opened 2024-11-16
zfs: load keys for encrypted datasets during pool import

cyphar:
If a user has set up their own zpools and given them to us to manage,
it's possible they've configured ZFS-native encryption. For the most
part, this works completely transparently to us. However, because we
manually do zpool-import and zpool-export during startup and shutdown of
Incus, ZFS datasets with keys will have their keys unloaded during
shutdown and then the keys are not automatically loaded on startup. This
results in containers being unable to start on startup because all IOs
are blocked indefinitely until the dataset keys are loaded manually by
the admin -- even if the admin has configured automatic key loading on
their system!

The simplest solution would be to pass -l to zfs-import (which causes
ZFS to auto-import all keys for all datasets in the pool). However, it
is slightly nicer to do a separate zfs-load-key so that we can unmount
the pool if the key import fails (zfs-import will leave the pool
imported but without keys loaded).

If the user has configured keylocation=prompt (or otherwise
misconfigured the encryption settings for their pool), the command will
fail and the pool import will fail loudly (rather than silently failing
in the form of an imported pool that is not usable as a filesystem).

Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

### DIFF
--- a/internal/server/storage/drivers/driver_zfs.go
+++ b/internal/server/storage/drivers/driver_zfs.go
@@ -407,41 +407,40 @@ func (d *zfs) Delete(op *operations.Operation) error {
 		return err
 	}
 
-	if !exists {
-		return nil
-	}
-
-	// Confirm that nothing's been left behind
-	datasets, err := d.getDatasets(d.config["zfs.pool_name"], "all")
-	if err != nil {
-		return err
-	}
-
-	initialDatasets := d.initialDatasets()
-	for _, dataset := range datasets {
-		dataset = strings.TrimPrefix(dataset, "/")
-
-		if slices.Contains(initialDatasets, dataset) {
-			continue
-		}
-
-		fields := strings.Split(dataset, "/")
-		if len(fields) > 1 {
-			return fmt.Errorf("ZFS pool has leftover datasets: %s", dataset)
-		}
-	}
-
-	if strings.Contains(d.config["zfs.pool_name"], "/") {
-		// Delete the dataset.
-		_, err := subprocess.RunCommand("zfs", "destroy", "-r", d.config["zfs.pool_name"])
+	if exists {
+		// Confirm that nothing's been left behind
+		datasets, err := d.getDatasets(d.config["zfs.pool_name"], "all")
 		if err != nil {
 			return err
 		}
-	} else {
+
+		initialDatasets := d.initialDatasets()
+		for _, dataset := range datasets {
+			dataset = strings.TrimPrefix(dataset, "/")
+
+			if slices.Contains(initialDatasets, dataset) {
+				continue
+			}
+
+			fields := strings.Split(dataset, "/")
+			if len(fields) > 1 {
+				return fmt.Errorf("ZFS pool has leftover datasets: %s", dataset)
+			}
+		}
+
 		// Delete the pool.
-		_, err := subprocess.RunCommand("zpool", "destroy", d.config["zfs.pool_name"])
-		if err != nil {
-			return err
+		if strings.Contains(d.config["zfs.pool_name"], "/") {
+			// Delete the dataset.
+			_, err := subprocess.RunCommand("zfs", "destroy", "-r", d.config["zfs.pool_name"])
+			if err != nil {
+				return err
+			}
+		} else {
+			// Delete the pool.
+			_, err := subprocess.RunCommand("zpool", "destroy", d.config["zfs.pool_name"])
+			if err != nil {
+				return err
+			}
 		}
 	}
 
--- a/internal/server/storage/drivers/driver_zfs.go
+++ b/internal/server/storage/drivers/driver_zfs.go
@@ -566,11 +566,26 @@ func (d *zfs) importPool() (bool, error) {
 		return false, err
 	}
 
-	if exists {
-		return true, nil
+	if !exists {
+		return false, fmt.Errorf("ZFS zpool exists but dataset is missing")
 	}
 
-	return false, fmt.Errorf("ZFS zpool exists but dataset is missing")
+	// We need to explicitly import the keys here so containers can start. This
+	// is always needed because even if the admin has set up auto-import of
+	// keys on the system, because incus manually imports and exports the pools
+	// the keys can get unloaded.
+	//
+	// We could do "zpool import -l" to request the keys during import, but by
+	// doing it separately we know that the key loading specifically failed and
+	// not some other operation. If a user has keylocation=prompt configured,
+	// this command will fail and the pool will fail to load.
+	_, err = subprocess.RunCommand("zfs", "load-key", "-r", d.config["zfs.pool_name"])
+	if err != nil {
+		_, _ = d.Unmount()
+		return false, fmt.Errorf("Failed to load keys for ZFS dataset %q: %w", d.config["zfs.pool_name"], err)
+	}
+
+	return true, nil
 }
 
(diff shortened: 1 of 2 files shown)


cyphar:
I haven't added tests yet because I'm not sure how you would like me to do that -- should we have just one specific test for this behaviour or run all of the ZFS-related tests on both an encrypted and unencrypted dataset?

Also let me know if you think just doing `zpool import -l` is good enough.

internal/server/storage/drivers/driver_zfs.go
```diff
@@ -567,6 +567,20 @@ func (d *zfs) importPool() (bool, error) {
...
+	// We need to explicitly import the keys here so containers can start. This
+	// is always needed because even if the admin has set up auto-import of
+	// keys on the system, because incus manually imports and exports the pools
+	// the keys can get unloaded.
+	//
+	// We could do "zpool import -l" to request the keys during import, but we
+	// also should give warnings to users if they've configured their ZFS
+	// datasets to have encryption setups that are incompatible with how incus
+	// manages its ZFS pools.
+	if err := d.loadKeys(d.config["zfs.pool_name"]); err != nil {
```
stgraber, line 579: Our coding style is to avoid the one line define+check syntax, can you split that into two separate statements?

internal/server/storage/drivers/driver_zfs_utils.go
```diff
@@ -409,6 +410,54 @@ func (d *zfs) receiveDataset(vol Volume, r io.Reader, tracker *ioprogress.Progre
...
+		// would normally silently fail for keylocation=prompt.
+		if !strings.HasPrefix(keyLocation, "file://") {
+			hasInteractiveKey = true
+			logger.Warnf("Dataset %q key cannot be loaded from keylocation %q non-interactively -- containers might fail to start until manual 'zfs load-key'", subPath, keyLocation)
+		}
+	}
+	if !isEncrypted {
+		return nil
+	}
+	if _, err := subprocess.RunCommand("zfs", "load-key", "-r", dataset); err != nil {
```
stgraber, line 447: Same here

internal/server/storage/drivers/driver_zfs_utils.go
```diff
@@ -409,6 +410,54 @@ func (d *zfs) receiveDataset(vol Volume, r io.Reader, tracker *ioprogress.Progre
...
+		// If all the keys are non-interactive then an error should fail the
+		// load, since this indicates that the keylocation= is misconfigured or
+		// the key itself is wrong.
+		if !hasInteractiveKey {
+			return err
+		}
+		logger.AddContext(logger.Ctx{
+			"err": err,
+		}).Warn(fmt.Sprintf("Dataset %q keys could not be loaded non-interactively", dataset))
```
stgraber, line 456: Can you make that one just one line, easier to grep :)

internal/server/storage/drivers/driver_zfs_utils.go
```diff
@@ -409,6 +410,54 @@ func (d *zfs) receiveDataset(vol Volume, r io.Reader, tracker *ioprogress.Progre
...
+	}
```
stgraber, line 457: Need a line break here

internal/server/storage/drivers/driver_zfs_utils.go
```diff
@@ -409,6 +410,54 @@ func (d *zfs) receiveDataset(vol Volume, r io.Reader, tracker *ioprogress.Progre
...
+	if !isEncrypted {
+		return nil
+	}
+	if _, err := subprocess.RunCommand("zfs", "load-key", "-r", dataset); err != nil {
+		// If all the keys are non-interactive then an error should fail the
+		// load, since this indicates that the keylocation= is misconfigured or
+		// the key itself is wrong.
+		if !hasInteractiveKey {
+			return err
+		}
```
stgraber, line 453: Need a line break here

internal/server/storage/drivers/driver_zfs_utils.go
```diff
@@ -409,6 +410,54 @@ func (d *zfs) receiveDataset(vol Volume, r io.Reader, tracker *ioprogress.Progre
...
+		// We still try to load all of the keys anyway, but we know that this
+		// would normally silently fail for keylocation=prompt.
+		if !strings.HasPrefix(keyLocation, "file://") {
+			hasInteractiveKey = true
+			logger.Warnf("Dataset %q key cannot be loaded from keylocation %q non-interactively -- containers might fail to start until manual 'zfs load-key'", subPath, keyLocation)
+		}
+	}
+	if !isEncrypted {
+		return nil
+	}
```
stgraber, line 446: Need a line break here

internal/server/storage/drivers/driver_zfs_utils.go
```diff
@@ -409,6 +410,54 @@ func (d *zfs) receiveDataset(vol Volume, r io.Reader, tracker *ioprogress.Progre
...
+		isEncrypted = true
+
+		// Log a warning if we hit a keylocation= that is not a regular file.
+		// We still try to load all of the keys anyway, but we know that this
+		// would normally silently fail for keylocation=prompt.
+		if !strings.HasPrefix(keyLocation, "file://") {
+			hasInteractiveKey = true
+			logger.Warnf("Dataset %q key cannot be loaded from keylocation %q non-interactively -- containers might fail to start until manual 'zfs load-key'", subPath, keyLocation)
+		}
+	}
```
stgraber, line 443: Need a line break here

internal/server/storage/drivers/driver_zfs_utils.go
```diff
@@ -409,6 +410,54 @@ func (d *zfs) receiveDataset(vol Volume, r io.Reader, tracker *ioprogress.Progre
...
+		fields := strings.SplitN(line, "\t", 2)
+		subPath, keyLocation := fields[0], fields[1]
+
+		// We need to check that all of the keylocations are either no-ops
+		// (none, -) or are regular files (file://*). Assume any other values
+		// (like prompt) are interactive to avoid hanging.
+		switch keyLocation {
+		case "", "-", "none":
+			continue
+		}
```
stgraber, line 433: Need a line break here

stgraber (changes requested):
Logic looks good to me.

For testing, the easiest would be a system tests which only runs under INCUS_BACKEND=zfs and proceeds to manually create a test loop backed test zpool with encryption, then restarts the test Incus instance to make sure it gets unlocked properly.

stgraber:
`zpool -l` may be easier for this, assuming it behaves correctly with /dev/null as stdin so it doesn't actually end up blocking and instead fails the import.

cyphar:
@stgraber Ah, so you would prefer if the import failed? I assumed you'd prefer a warning (hence all of the logic for figuring out which datasets have non-`file://` `keylocation`s). That would make this patch much simpler. Let me test it out...

stgraber:
A storage pool failing to start should have it go into a failed state that gets retried so we should be okay with it failing to mount.

cyphar:
@stgraber Okay, `zpool import -l` will return an error if `stdin` is `/dev/null` and we have `keylocation=prompt`. However, the pool will still be imported (just without the keys loaded), so I guess we should export the pool in that case? Or should we leave it in the half-imported state?

stgraber (approved):

PR #1385, opened 2024-11-16
incus: Add aliases completion

montag451:
This PR also improves the configuration loading because I didn't want to copy/paste on more time the code used to locate the user configuration file.

### DIFF
--- a/shared/cliconfig/file.go
+++ b/shared/cliconfig/file.go
@@ -3,6 +3,8 @@ package cliconfig
 import (
 	"fmt"
 	"os"
+	"os/user"
+	"path"
 	"path/filepath"
 
 	"gopkg.in/yaml.v2"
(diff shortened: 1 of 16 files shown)


shared/cliconfig/file.go
```diff
@@ -11,17 +13,57 @@ import (
...
+		return "", "", nil
+	}
+
+	configPath := os.ExpandEnv(path.Join(configDir, "config.yml"))
+
+	return configPath, filepath.Dir(configPath), nil
+}
+
 // LoadConfig reads the configuration from the config path; if the path does
 // not exist, it returns a default configuration.
```
stgraber, line 44: That description needs updating as it no longer matches what the function does.

montag451: Something like:
``` go
// LoadConfig reads the configuration from the config path; if the path does
// not exist, it returns a default configuration; if the given path is empty
// it tries to determine automatically the configuration file to load.
```

montag451: Any advice?

stgraber: Sorry, been traveling this week and struggling to keep up with Github :)
Yeah, that comment looks good to me.

montag451: Done!

stgraber (approved):

stgraber:
Thanks!

PR #1387, opened 2024-11-16
incus: Improve completion for config keys

montag451:
The code is mainly copy/paste from the LXD repository. Is it OK to do that ?

### DIFF
--- a/cmd/incus/completion.go
+++ b/cmd/incus/completion.go
@@ -238,10 +238,108 @@ func (g *cmdGlobal) cmpImages(toComplete string) ([]string, cobra.ShellCompDirec
 	return results, cmpDirectives
 }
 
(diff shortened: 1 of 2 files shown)


stgraber:
No, it's not. The LXD repository is under a different license, we cannot legally even look at the content of a pull request which is copying code from the LXD repository.

montag451:
OK but the LXD completion code comes from Incus in the first place as you can see here https://github.com/canonical/lxd/commit/3c0f4633b64eccca6c49f5303c9dad3b0b38feef. Doesn't it make the code a Apache-2.0 licenced code? I really don't understand how licenses are supposed to work in this context.

stgraber:
The original code may have originated with us, but if they then extended it, that change isn't Apache 2.0.

PR #1391, opened 2024-11-17
Use `maps.Clone()` where appropriate

montag451:
Go 1.21 introduced a new `maps` package which provides some utility functions to facilitate some common operations on `map`. One of these functions is `maps.Clone` which eases the cloning of a map. Moreover, the `maps.Clone` function is slightly more efficient than manual cloning (cf https://github.com/golang/go/commit/b5a7f2eef7cb17255cb396cd4ff7df04957dd21e)

### DIFF
--- a/cmd/incusd/api_1.0.go
+++ b/cmd/incusd/api_1.0.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"maps"
 	"net"
 	"net/http"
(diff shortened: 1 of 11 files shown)


montag451:
I missed the fact that `maps.Clone` return a nil map when the passed map is nil which lead to crashes in some parts of the code so I prefer to close the PR because the benefit of using `maps.Clone` is not as high as expected.

Issue #1395, opened 2024-11-19
incus 6.7 fails to (cross)compile on 32bit archs


dkwo:
I believe the relevant errors are
```
# github.com/lxc/incus/v6/internal/linux
In file included from internal/linux/socket_linux_cgo.go:25:
internal/linux/../../shared/cgo/process_utils.h: In function 'setproctitle':
internal/linux/../../shared/cgo/process_utils.h:222:30: error: cast to pointer from integer of different size [-Werror=int-to-pointer-cast]
  222 |                 char *dest = (char *)arg_start;
      |                              ^
cc1: all warnings being treated as errors
github.com/mitchellh/mapstructure
html
github.com/russross/blackfriday/v2
gopkg.in/yaml.v3
github.com/cpuguy83/go-md2man/v2/md2man
net/http/httputil
github.com/gorilla/mux
# github.com/lxc/incus/v6/shared/idmap
In file included from shared/idmap/shift_linux.go:26:
shared/idmap/../../shared/cgo/process_utils.h: In function 'setproctitle':
shared/idmap/../../shared/cgo/process_utils.h:222:30: error: cast to pointer from integer of different size [-Werror=int-to-pointer-cast]
  222 |                 char *dest = (char *)arg_start;
      |                              ^
cc1: all warnings being treated as errors
```
as seen in https://github.com/void-linux/void-packages/pull/53135

stgraber:
I've pinged @brauner who I believe is the original author of that code from liblxc.

It's pretty dark magic that we're pulling in there so I'd prefer someone who knows what they're doing with C to look at sorting this one out ;)

stgraber:
Yeah, which is a bit odd considering it's code we took from liblxc which has shipped for years on 32bit platforms :)

rnalrd:
Adding `-Wno-int-to-pointer-cast` to CGO_CFLAGS allowed to build on Alpine Linux 32-bit

ncopa:
I suppose it is gcc 14 that is stricter about things.

> cc1: all warnings being treated as errors

Generally not a good idea with `-Werror` on releases (but very good in CI /development)

PR #1398, opened 2024-11-19
cgo/process_utils: fix 32bit builds

brauner:
New clang-14 reports errors for directly casting from an integer to a pointer. Use uintptr_t to fix that problem and adhere to the casting rules.

Fixes: https://github.com/lxc/incus/issues/1395


stgraber (approved):

stgraber:
Thanks!

Issue #188, opened 2023-10-22
Basic scriptlet-based authorizer


stgraber:
We now have OpenFGA for fine grained access control, but not everyone needs to run a full authorization server and for simpler cases, it could be very useful to have a basic built-in option that can still be very flexible.

I'm thinking of a new `authorization.scriptlet` server config key which takes a scriplet that would then be called with some user context, object context and action and have the ability to allow or deny the request.

sharathsivakumar:
@stgraber I would be interested to work on this.

hi-ko:
I would be happy to test on this. I guess such a script hook would open for much more scenarios to use incus as remote image server having something like a voter script which could then access external services like ldap or db to block or allow access. Are you thinking of a shell script or a Go construct?

stgraber:
We're looking at using a scriptlet which is already a technology we use for our scheduler.
Those scriptlets are python looking scripts that are run inside of a safe environment directly within the Go daemon.

This uses go-starlark from Google.

stgraber:
Unassigning due to lack of activity.

christina-zh:
Im interested in working on this issue, can I be assigned to it please?

stgraber:
Probably best to tackle last in the current batch of stuff assigned to you as that's a bit trickier than the others. But #652 should help you get familiar with some of that stuff already.

This will need a new kind of scriptlet be defined in `internal/server/scriptlet`. It will also need a new server configuration key `authorization.scriptlet` to be added.

The basic infrastructure you can do with:
 - Add an API extension, let's call it `authorization_scriptlet`
 - Add the new config key in `internal/server/cluster/config/config.go`
 - Re-generate the documentation keys with `make update-metadata`
 - Add the new scriptlet definition to `internal/server/scriptlet`
 - Add the new authorizer to `internal/server/auth`
 - Add the needed glue (handle the new config key) to `cmd/incusd/daemon.go` and `cmd/incusd/api_1.0.go`

stgraber:
@christina-zh hello, do you still intend to work on this one or should I clear the assignee?

christina-zh:
hi @stgraber , our group might not have the time to continue working on this issue. You could clear the assignee. Thank you for your help!

bensmrs:
Hi! I’d like to work on this, you can put it in the “soon” milestone.

stgraber:
There you go!

bensmrs:
A few questions as I’m implementing this…
[...]

PR #1412, opened 2024-11-22
Add authorization scriptlet

bensmrs:
Closes: #188

### DIFF
--- a/cmd/incusd/daemon.go
+++ b/cmd/incusd/daemon.go
@@ -1477,6 +1477,7 @@ func (d *Daemon) init() error {
 	syslogSocketEnabled := d.localConfig.SyslogSocket()
 	openfgaAPIURL, openfgaAPIToken, openfgaStoreID := d.globalConfig.OpenFGA()
 	instancePlacementScriptlet := d.globalConfig.InstancesPlacementScriptlet()
+	authorizationScriptlet := d.globalConfig.AuthorizationScriptlet()
 
 	d.endpoints.NetworkUpdateTrustedProxy(d.globalConfig.HTTPSTrustedProxy())
 	d.globalConfigMu.Unlock()
@@ -1510,6 +1511,14 @@ func (d *Daemon) init() error {
 		err = d.setupOpenFGA(openfgaAPIURL, openfgaAPIToken, openfgaStoreID)
 		if err != nil {
 			return fmt.Errorf("Failed to configure OpenFGA: %w", err)
+		}
+	}
+
+	// Setup the authorization scriptlet.
+	if authorizationScriptlet != "" {
+		err = d.setupAuthorizationScriptlet(authorizationScriptlet)
+		if err != nil {
+			return err
 		}
 	}
 
@@ -2241,6 +2250,29 @@ func (d *Daemon) setupOpenFGA(apiURL string, apiToken string, storeID string) er
 	d.authorizer = openfgaAuthorizer
 
 	revert.Success()
+	return nil
+}
+
+// Setup authorization scriptlet.
+func (d *Daemon) setupAuthorizationScriptlet(scriptlet string) error {
+	err := scriptletLoad.AuthorizationSet(scriptlet)
+	if err != nil {
+		return fmt.Errorf("Failed saving authorization scriptlet: %w", err)
+	}
+
+	if scriptlet != "" {
+		// Fail if not using the default tls or scriptlet authorizer.
+		switch d.authorizer.(type) {
+		case *auth.TLS, *auth.Scriptlet:
+			d.authorizer, err = auth.LoadAuthorizer(d.shutdownCtx, auth.DriverScriptlet, logger.Log, d.clientCerts)
+			if err != nil {
+				return err
+			}
+		default:
+			return errors.New("Attempting to setup scriptlet authorization while a non-default authorizer is already set")
+		}
+	}
+
 	return nil
 }
 
--- a/cmd/incusd/daemon.go
+++ b/cmd/incusd/daemon.go
@@ -2260,17 +2260,26 @@ func (d *Daemon) setupAuthorizationScriptlet(scriptlet string) error {
 		return fmt.Errorf("Failed saving authorization scriptlet: %w", err)
 	}
 
-	if scriptlet != "" {
-		// Fail if not using the default tls or scriptlet authorizer.
-		switch d.authorizer.(type) {
-		case *auth.TLS, *auth.Scriptlet:
-			d.authorizer, err = auth.LoadAuthorizer(d.shutdownCtx, auth.DriverScriptlet, logger.Log, d.clientCerts)
-			if err != nil {
-				return err
-			}
-		default:
-			return errors.New("Attempting to setup scriptlet authorization while a non-default authorizer is already set")
+	if scriptlet == "" {
+		// Reset to default authorizer.
+		d.authorizer, err = auth.LoadAuthorizer(d.shutdownCtx, auth.DriverTLS, logger.Log, d.clientCerts)
+		if err != nil {
+			return err
 		}
+
+		return nil
+	}
+
+	// Fail if not using the default tls or scriptlet authorizer.
+	switch d.authorizer.(type) {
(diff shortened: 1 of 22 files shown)


cmd/incusd/daemon.go
```diff
@@ -2205,6 +2214,37 @@ func (d *Daemon) setupOpenFGA(apiURL string, apiToken string, storeID string) er
...
+		if err != nil {
+			return err
+		}
+
+		return nil
+	} else {
+		// Fail if not using the default tls or scriptlet authorizer.
+		switch d.authorizer.(type) {
+		case *auth.TLS, *auth.Scriptlet:
+			d.authorizer, err = auth.LoadAuthorizer(d.shutdownCtx, auth.DriverScriptlet, logger.Log, d.clientCerts)
```
bensmrs, line 2236: If you want to add a `StopService` there, please tell. I didn’t add it because basically, declaring two authorizers is forbidden, and both the default authorizer and the scriptlet authorizer’s `StopService`s are a no-op.

bensmrs:
I don’t really get why the code checks are failing, the errors are not something my knowledge of Go allows me to decipher :)

stgraber:
Alright, got all the tests to behave. Now I can actually review this PR :)

cmd/incusd/daemon.go
```diff
@@ -2205,6 +2214,38 @@ func (d *Daemon) setupOpenFGA(apiURL string, apiToken string, storeID string) er
...
+	// Fail if not using the default tls or scriptlet authorizer.
+	switch d.authorizer.(type) {
+	case *auth.TLS, *auth.Scriptlet:
+		d.authorizer, err = auth.LoadAuthorizer(d.shutdownCtx, auth.DriverScriptlet, logger.Log, d.clientCerts)
+		if err != nil {
+			return err
+		}
+
+	default:
+		return errors.New("Attempting to setup scriptlet authorization while a non-default authorizer is already set")
```
stgraber, line 2243: Maybe `while another authorizer is already set`? (non-default sounds kinda weird as something that may be shown to a user)

cmd/incusd/daemon.go
stgraber: That should be fine

bensmrs: Well now that I’m thinking about it, if we create a new authorizer that needs to be properly cleaned (closing a fd, releasing a lock…), that may be problematic.
On the other hand, the way authorizers work now feels like it has only be thought for two authorizers (the default one and one non-default).

I feel that the {Stop the default -> Load the new -> [if something isn’t right,] Stop the new -> Load the default again} process needs to be rewritten for the OpenFGA authorizer, and it needs to be more generic to allow new authorizers to be written easily.

We should also probably be more defensive here and check the unicity of authorizer configuration keys before initializing the authorizers…

So we keep it this way for this PR and later think of a cleaner approach? I can probably work on that next month. If that sounds good, I’ll summarize it and open another issue.

bensmrs:
Hi! Does it help if I rebase the PR?

stgraber:
Yeah, that'd be quite useful, thanks!

bensmrs:
There you go. I started freaking out when I saw changes to the OpenFGA driver, but it wasn’t that dramatic in the end.

cmd/incusd/daemon.go
stgraber: Yeah, it's fine for now, but definitely can do with improvement down the line.

doc/authorization.md
```diff
@@ -64,3 +65,21 @@ Users that you do not trust with root access to the host should not be granted t
...
 ```
+
+(authorization-scriptlet)=
+## Scriptlet authorization
+
+Incus supports defining a scriptlet to manage fine-grained authorization, allowing to write precise authorization rules with no dependency on external tools.
+
+To use scriptlet authorization, you can write a scriptlet in the `authorization.scriptlet` server configuration option implementing a function `authorize`, which takes three arguments:
+
+- `details`, the authorization request details
```
stgraber, line 76: Can we expand a bit on what's part of `details`?

doc/authorization.md
```diff
@@ -64,3 +65,21 @@ Users that you do not trust with root access to the host should not be granted t
...
+- `object`, the object on which the user requests authorization
+- `entitlement`, the authorization level asked by the user
+
+This function must return a boolean indicating whether the user has access or not to the given object with the given entitlement.
```
stgraber, line 80: Documentation so far uses `Boolean` which is allowed already in the dictionary

doc/.wordlist.txt
```diff
@@ -25,6 +25,7 @@ BGP
 bibi
 BitLocker
 bool
+boolean
```
stgraber, line 28: This change shouldn't be needed

stgraber:
Looking good.

On the `details` part, I don't think we want to expose a whole `RequestDetails` to the scriptlet as the forwarded pieces are confusing and shouldn't be handled by the scriptlet.

Instead we should expose a script which just includes:
 - Protocol string
 - Username string
 - Project string
 - AllProjects bool

And have the username and protocol be pre-expanded as is done in `username` and `authenticationProtocol`.

stgraber:
Also, this is a feature which we should be able to readily test, so it would be good to introduce a test for it in `test/`, basically just need to connect over the network (`incus_remote` with the correct remote) and then load some scriptlets that allow/reject some of the requests and make sure it's working as expected.

bensmrs:
> On the `details` part, I don't think we want to expose a whole `RequestDetails` to the scriptlet as the forwarded pieces are confusing and shouldn't be handled by the scriptlet.

Ok that makes part of the refactoring useless; I’m de-refactoring a bit :)

bensmrs:
Well, something’s gone terribly wrong with this rebase :D

bensmrs:
> Also, this is a feature which we should be able to readily test, so it would be good to introduce a test for it in `test/`, basically just need to connect over the network (`incus_remote` with the correct remote) and then load some scriptlets that allow/reject some of the requests and make sure it's working as expected.

I’m having quite some trouble navigating through the testing infrastructure’s code. For example, when am I supposed to use `incus_remote` vs `incus`? `openfga.sh` doesn’t use `incus_remote`, so I’ll try tinkering about from it and we’ll see.

internal/server/auth/common/common.go
```diff
@@ -0,0 +1,9 @@
+package common
+
+// RequestDetails is a type representing an authorization request.
+type RequestDetails struct {
+	UserName             string
```
stgraber, line 5: Tiny nit, but lets make that `Username` to line up with our usual spelling across the codebase (there are two exceptions currently, but neither are user visible).

doc/authorization.md
```diff
@@ -64,3 +65,21 @@ Users that you do not trust with root access to the host should not be granted t
...
 ```
+
+(authorization-scriptlet)=
+## Scriptlet authorization
+
+Incus supports defining a scriptlet to manage fine-grained authorization, allowing to write precise authorization rules with no dependency on external tools.
+
+To use scriptlet authorization, you can write a scriptlet in the `authorization.scriptlet` server configuration option implementing a function `authorize`, which takes three arguments:
+
+- `details`, an object with attributes `UserName` (the user name or certificate fingerprint), `Protocol` (the authentication protocol), `IsAllProjectsRequest` (whether the request is made on all projects) and `ProjectName` (the project name)
```
stgraber, line 76: Will need to update to `Username` once the RequestDetails struct is updated.

stgraber:
Looks good, just a capitalization nit left ;)

It's nice that the test uses OIDC, makes it easier to try multiple users.

bensmrs:
> It's nice that the test uses OIDC, makes it easier to try multiple users.

It was actually a mix of “I don’t want to generate certificates and pass them cleanly to Incus” and ”Oh! The tests make it sooo easy to use OIDC”, combined to “I prefer to use explicit user names in the tests”.
Kudos for all the very useful helper functions!

bensmrs:
Hey! all these rebases are starting to become quite boring :D

stgraber (approved):

stgraber:
Thanks for the great work!

PR #1427, opened 2024-11-26
incusd/instances_post: Fix cluster internal migrations

stgraber:
The StoragePool field should only be set if a full cross-pool migration is expected to occur. This isn't normally the case within a cluster when the pool name itself doesn't change.


presztak (approved):
LGTM

PR #1431, opened 2024-11-27
Add support for VGA console screenshots

### LINKED ISSUE #1419: Add support for VGA console screenshots
One feature that'd be quite useful for the web UI is the ability to show a screenshot of the current VGA console. This is particularly useful as only one user may be connected to the VM at any one time.

To do this, there is the `screendump` QMP command (https://qemu-project.gitlab.io/qemu/interop/qemu-qmp-ref.html#qapidoc-1550) which should be able to give us what we want. As we can't directly write to a file, we'll need to use our usual trick of sending a file over through QMP first, then once we know the slot number, have `screendump` dump its data through `/proc/self/fd/NUMBER`. We should try to request the screenshot be taken as a `png`. This has apparently been supported for a couple of years now.

On the client side, we should have this implemented by extending `GET /1.0/instances/NAME/console` to take a `?type=` variable, which if set to `vga` will then have us return the VGA console screenshot.

A rough list of expected changes are:
 - Allocate an API extension (`api: instance_console_screenshot`)
 - Add a binding for `screendump` in `internal/server/instance/drivers/qmp`
 - Add `ConsoleScreenshot` to the `VM` interface in `internal/server/instance`
 - Implement `ConsoleScreenshot` in `internal/server/instance/drivers/driver_qemu.go`
 - Extend `GET /1.0/instances/NAME/console` to take a `type` input variable (`cmd/incusd/instance_console.go`)
 - Re-generate the swagger metadata for previous change (`make update-api`)
 - Implement `type=vga` in the `GET /1.0/instances/NAME/console` handler

breml:
Resolves #1419

### DIFF
--- a/cmd/incusd/instance_console.go
+++ b/cmd/incusd/instance_console.go
@@ -533,9 +533,10 @@ func instanceConsolePost(d *Daemon, r *http.Request) response.Response {
 
 // swagger:operation GET /1.0/instances/{name}/console instances instance_console_get
 //
-//	Get console log
+//	Get console output
 //
-//	Gets the console log for the instance.
+//	Gets the console output for the instance either as text log or as vga
+//	screendump.
 //
 //	---
 //	produces:
@@ -546,14 +547,27 @@ func instanceConsolePost(d *Daemon, r *http.Request) response.Response {
 //	    description: Project name
 //	    type: string
 //	    example: default
+//	  - in: query
+//	    name: type
+//	    description: Console type
+//	    type: string
+//	    enum: [log, vga]
+//	    default: log
+//	    example: vga
 //	responses:
 //	  "200":
-//	     description: Raw console log
+//	     description: |
+//	       Console output either as raw console log or as vga screendump in PNG
+//	       format depending on the `type` parameter provided with the request.
 //	     content:
 //	       application/octet-stream:
 //	         schema:
 //	           type: string
 //	           example: some-text
+//	       image/png:
+//	         schema:
+//	           type: string
+//	           format: binary
 //	  "400":
 //	    $ref: "#/responses/BadRequest"
 //	  "403":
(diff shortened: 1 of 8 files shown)


breml:
@stgraber I don't think https://github.com/lxc/incus/actions/runs/12065465418/job/33644309917?pr=1431#step:14:21168 is related to my change, but it is maybe still something we would like to look into. On first sight, I can not see anything that is off with the use of the mutex in [func (g *Group) Start(ctx context.Context) {](https://github.com/breml/incus/blob/2eb09202d2f5565611d1ff2e1201eeb801483746/internal/server/task/group.go#L38).

The only thing I wonder is, if the usage of the wait group is correct in this code. I see, that we add all the tasks to the wait group in https://github.com/breml/incus/blob/2eb09202d2f5565611d1ff2e1201eeb801483746/internal/server/task/group.go#L46, but we do not always call [`wg.Done()`](https://github.com/breml/incus/blob/2eb09202d2f5565611d1ff2e1201eeb801483746/internal/server/task/group.go#L69), since it is not called with `defer` inside of the goroutine and we skip some tasks, if these are already in running state. Therefore I wonder, if `Start` might be called multiple times after adding more tasks. If so, we should only add the tasks to the wait group, that we have not yet added before, because otherwise we end up with a wait group, that will never be drained.

I wonder, if it would be better to rewrite `Start` like this:

```golang
// Start all the tasks in the group.
func (g *Group) Start(ctx context.Context) {
	// Lock access to the g.running and g.tasks map for the entirety of this function so that
	// concurrent calls to Start() or Add(0) don't race. This ensures all tasks in this group
	// are started based on a consistent snapshot of g.running and g.tasks.
	g.mu.Lock()
	defer g.mu.Unlock()

	ctx, g.cancel = context.WithCancel(ctx)

	if g.running == nil {
		g.running = make(map[int]bool)
	}

	for i := range g.tasks {
		if g.running[i] {
			continue
		}

		g.running[i] = true
		task := g.tasks[i] // Local variable for the closure below.
		g.wg.Add(1)

		go func(i int) {
			defer g.wg.Done()

			task.loop(ctx)

			// Ensure running map is updated before wait group Done() is called.
			g.mu.Lock()
			defer g.mu.Unlock()

			// FIXME: is this even possible?
			if g.running != nil {
				g.running[i] = false
			}
		}(i)
	}
}
```

I wonder, if it is even possible for `g.running` to be `nil` inside of the goroutine.

stgraber:
Code looks good, just had to tweak the commit structure to match what we normally do.

stgraber (approved):

PR #1445, opened 2024-12-01
incus: Improve completion for `file push` and `file pull`

montag451:
Completion for files inside instances is not provided as I don't know how to do it properly for the moment, the improvement concerns the fact that instances names are now returned as completion candidates. Before only local file names was provided.


stgraber (approved):

montag451:
@stgraber Can you rerun the checks please, I don't think my changes are responsible for the failure so running the checks again should fix the failure.

PR #1447, opened 2024-12-02
incusd/device/tpm: Wait for swtpm to be ready

stgraber:


internal/server/device/tpm.go
```diff
@@ -227,6 +227,9 @@ func (d *tpm) startVM() (*deviceConfig.RunConfig, error) {
 		return nil, fmt.Errorf("Failed to save swtpm state for device %q: %w", d.name, err)
 	}
 
+	// swtpm unfortunately isn't always immediately ready for QEMU...
+	time.Sleep(time.Second)
```
breml, line 231: I wonder, if there isn't a nicer way to make sure, that the swtpm is indeed ready. With this sleep, we always loose 1 second here. On the other hand, waiting 1 second might also not always be enough for the swtmp to become ready e.g. on a heavily loaded or very minimal system.

stgraber: I was somehow remembering our older implementation where we were passing an fd which would have been pretty problematic to monitor, but given that following a fix for a path length issue, we're now using a regular socket, we can just wait for it to exist (after ensuring it doesn't exist prior to starting up).

internal/server/device/tpm.go
```diff
@@ -227,6 +230,21 @@ func (d *tpm) startVM() (*deviceConfig.RunConfig, error) {
 		return nil, fmt.Errorf("Failed to save swtpm state for device %q: %w", d.name, err)
 	}
 
+	// Wait for the socket to be available.
+	exists := false
+	for i := 0; i < 20; i++ {
+		if util.PathExists(socketPath) {
+			exists = true
+			continue
```
gibmat, line 238: Wouldn't `break` be better?

stgraber: doh, yes, that's what I meant to put there ;)

stgraber: Fixed

PR #1473, opened 2024-12-07
Switch to mount.ceph

MadnessASAP:
Currently when mounting CephFS Incus will construct the mount syscall by attempting to parse `/etc/ceph/ceph.conf `. Unfortunately this is fragile given the varieties of locations, layouts, and forms the necessary options can take.

This PR fixes that by instead diverting ceph mounts through the mount.ceph helper which will do all the necessary lifting. This obviously implies a dependency on the ceph client tools, however that dependency already existed for other ceph operations and isn't particularly unreasonable for a host interacting with ceph.

* Currently this is untested *

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -200,7 +200,6 @@ jobs:
             bind9-dnsutils \
             btrfs-progs \
             busybox-static \
-            ceph-common \
             dnsmasq-base \
             easy-rsa \
             gettext \
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -1,0 +1,318 @@
+name: Dev
+on:
+  push:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  code-tests:
+    name: Code
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        go:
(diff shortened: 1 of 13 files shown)


MadnessASAP:
```
24-12-07T21:37:02.6077872Z Error: Failed to mount "admin@.cephfs=/eZ5" on "/tmp/incus_cephfs_44833359/mount" using "ceph": Failed to run: mount.ceph admin@.cephfs=/eZ5 /tmp/incus_cephfs_44833359/mount -o : exit status 32 (mount error 2 = No such file or directory)
```

The CI failed because it doesn't have mount.ceph, for this to work either the tests will need to be adjusted, or using the helper is gated behind a config variable with the old methods intact.

My preference would be to still remove the config file scraping but add options for fsid, monitor addresses, and keyring. With the option to use the mount.ceph helper instead of specifying those.

MadnessASAP:
9dc3cm

MadnessASAP:
I'm going to close this PR as it doesn't really cover the scope of the changes needed and was going to cause more problems then it fixed.  However a much more promising 2nd attempt is continuing on this branch [ceph-refactor](https://github.com/MadnessASAP/incus/tree/ceph-refactor) which will be hopefully seeing a PR in the near future.

MadnessASAP:
I'm going to close this PR as it doesn't really cover the scope of the changes needed and was going to cause more problems then it fixed.  However a much more promising 2nd attempt is continuing on this branch [ceph-refactor](https://github.com/MadnessASAP/incus/tree/ceph-refactor) which will be hopefully seeing a PR in the near future.

PR #1478, opened 2024-12-07
OCI image debugging improvements

danbiagini:
### Problem

I ran into some challenges while trying to get the new incus OCI image support working with a Google Cloud registry (Artifact Registry).  The good news it was all self inflicted and the feature worked as designed, but it wasn't easy to debug given the reliance on skopeo/umoci and lack of logging and error pass throughs.  

### Solution
Added debug logging of the skopeo and umoci std err.  This is especially useful with authenticated connections that need to use the auth.json configuration file to work.

### Potential Improvements
I considered changing the error handling / pass through logic in `ImageDownload`, however I'm not familiar enough with the code and the tests aren't working in my development environment.  There might be cases where `GetImageAliasType` can fail however `GetImage` will not.

```		if args.Secret == "" {
			// Look for a matching alias.  Note, this err message is lost!
			entry, _, err := remote.GetImageAliasType(args.Type, fp)
			if err != nil {
				return nil, false, fmt.Errorf("Failed getting remote image alias: %w", err)
			}
                         fp = entry.Target

			// Expand partial fingerprints
			info, _, err = remote.GetImage(fp)
			if err != nil {
				return nil, false, fmt.Errorf("Failed getting remote image info: %w", err)
			}
```

### DIFF
--- a/client/oci_images.go
+++ b/client/oci_images.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/lxc/incus/v6/shared/api"
 	"github.com/lxc/incus/v6/shared/ioprogress"
+	"github.com/lxc/incus/v6/shared/logger"
 	"github.com/lxc/incus/v6/shared/osarch"
 	"github.com/lxc/incus/v6/shared/subprocess"
 	"github.com/lxc/incus/v6/shared/units"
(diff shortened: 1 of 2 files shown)


stgraber (approved):
Thanks, looks good!

I did a tiny bit of rebasing and re-titling but that was it.

PR #1490, opened 2024-12-11
Accept io.Writer in RenderTable

breml:
Makes `RenderTable` more flexible and allows to pass a different `io.Writer` instead of `os.Stdout` e.g. for testing.

This can be seen as a preparation step to replace all all direct writing to Stdout (and Stderr) with the facilities provided by `github.com/spf13/cobra`, namely [`command.Print`](https://pkg.go.dev/github.com/spf13/cobra#Command.Print) and the likes.

### DIFF
--- a/cmd/incusd/main_cluster.go
+++ b/cmd/incusd/main_cluster.go
@@ -369,7 +369,7 @@ func (c *cmdClusterListDatabase) Run(cmd *cobra.Command, args []string) error {
 		data[i] = []string{address}
 	}
 
-	_ = cli.RenderTable(c.flagFormat, columns, data, nil)
+	_ = cli.RenderTable(os.Stdout, c.flagFormat, columns, data, nil)
 
 	return nil
 }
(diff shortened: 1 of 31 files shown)


cmd/incusd/main_cluster.go
```diff
@@ -369,7 +369,7 @@ func (c *cmdClusterListDatabase) Run(cmd *cobra.Command, args []string) error {
 		data[i] = []string{address}
 	}
 
-	_ = cli.RenderTable(c.flagFormat, columns, data, nil)
+	_ = cli.RenderTable(os.Stdout, c.flagFormat, columns, data, nil)
```
gibmat, line 372: I would put the other changes to this file in a separate commit.

stgraber:
I'll take care of re-slicing this one. I'm actively trying to clear the PR backlog for Incus ahead of the 6.8 release.

stgraber (approved):

PR #1491, opened 2024-12-11
Add `--force` flag to the console command

presztak:
This PR introduces a `--force` flag to the console command. When used, it forces the connection and closes any existing ones. Without the flag, the user will receive a message about active connections. This change addresses an issue with Spice/QEMU, where connections are sometimes not properly closed after disconnection.

### DIFF
--- a/client/incus_instances.go
+++ b/client/incus_instances.go
@@ -2528,6 +2528,10 @@ func (r *ProtocolIncus) ConsoleInstance(instanceName string, console api.Instanc
 		return nil, fmt.Errorf("The server is missing the required \"console_vga_type\" API extension")
 	}
 
+	if console.Force && !r.HasExtension("console_force") {
+		return nil, fmt.Errorf(`The server is missing the required "console_force" API extension`)
+	}
+
 	// Send the request
 	op, _, err := r.queryOperation("POST", fmt.Sprintf("%s/%s/console", path, url.PathEscape(instanceName)), console, "")
 	if err != nil {
@@ -2614,6 +2618,10 @@ func (r *ProtocolIncus) ConsoleInstanceDynamic(instanceName string, console api.
 
 	if console.Type == "vga" && !r.HasExtension("console_vga_type") {
 		return nil, nil, fmt.Errorf("The server is missing the required \"console_vga_type\" API extension")
+	}
+
+	if console.Force && !r.HasExtension("console_force") {
+		return nil, nil, fmt.Errorf(`The server is missing the required "console_force" API extension`)
 	}
 
 	// Send the request.
--- a/cmd/incus/console.go
+++ b/cmd/incus/console.go
@@ -26,6 +26,7 @@ import (
 type cmdConsole struct {
 	global *cmdGlobal
 
+	flagForce   bool
 	flagShowLog bool
 	flagType    string
 }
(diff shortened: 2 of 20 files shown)


cmd/incus/console.go
```diff
@@ -41,6 +42,7 @@ This command allows you to interact with the boot console of an instance
 as well as retrieve past log entries from it.`))
 
 	cmd.RunE = c.Run
+	cmd.Flags().BoolVarP(&c.flagForce, "force", "f", false, i18n.G("Forces a connection to the console, even if there is already an active session")+"``")
```
breml, line 45: Just curious, what is the suffix
```
"``"
```
for? Why is it appended to the usage string of some of the flags (but not to all of them)?

client/incus_instances.go
```diff
@@ -2616,6 +2620,10 @@ func (r *ProtocolIncus) ConsoleInstanceDynamic(instanceName string, console api.
 		return nil, nil, fmt.Errorf("The server is missing the required \"console_vga_type\" API extension")
 	}
 
+	if console.Force && !r.HasExtension("console_force") {
+		return nil, nil, fmt.Errorf("The server is missing the required \"console_force\" API extension")
```
breml, line 2624: I would write this with back ticks to avoid the need for escaping:
```suggestion
		return nil, nil, fmt.Errorf(`The server is missing the required "console_force" API extension`)
```

client/incus_instances.go
```diff
@@ -2528,6 +2528,10 @@ func (r *ProtocolIncus) ConsoleInstance(instanceName string, console api.Instanc
 		return nil, fmt.Errorf("The server is missing the required \"console_vga_type\" API extension")
 	}
 
+	if console.Force && !r.HasExtension("console_force") {
+		return nil, fmt.Errorf("The server is missing the required \"console_force\" API extension")
```
breml, line 2532: I would write this with back ticks to avoid the need for escaping:
```suggestion
		return nil, fmt.Errorf(`The server is missing the required "console_force" API extension`)
```

cmd/incusd/instance_console.go
```diff
@@ -500,6 +521,41 @@ func instanceConsolePost(d *Daemon, r *http.Request) response.Response {
 		return response.BadRequest(fmt.Errorf("Instance is frozen"))
 	}
 
+	// Find any running 'ConsolShow' operation for the instance.
```
breml, line 524:
```suggestion
	// Find any running 'ConsoleShow' operation for the instance.
```

cmd/incusd/instance_console.go
```diff
@@ -500,6 +521,41 @@ func instanceConsolePost(d *Daemon, r *http.Request) response.Response {
...
+		}
+
+		// Fetch instance name from operation.
+		r := op.Resources()
+		apiUrls := r["instances"]
+		if len(apiUrls) < 1 {
+			return response.SmartError(fmt.Errorf("Operation does not have an instance URL defined"))
+		}
+
+		urlParts := strings.Split(apiUrls[0].URL.Path, "/")
```
breml, line 539: Would it be better to use [`path.Split`](https://pkg.go.dev/path#Split) here?

breml (commented):
Some small findings.

cmd/incusd/instance_console.go
```diff
@@ -500,6 +521,40 @@ func instanceConsolePost(d *Daemon, r *http.Request) response.Response {
...
+		if urlPrefix == "" || instanceName == "" {
+			return response.SmartError(fmt.Errorf("Instance URL has incorrect format"))
+		}
+
+		if instanceName != inst.Name() {
+			continue
+		}
+
+		if !post.Force {
+			return response.SmartError(fmt.Errorf("Active session exists. Use --force to force the connection"))
```
stgraber, line 549: We shouldn't mention command line flags in server-side errors.

So maybe go with `This console is already connected. Force is required to take it over.`

cmd/incus/console.go
presztak: Done.

client/incus_instances.go
presztak: Done.

cmd/incusd/instance_console.go
presztak: Fixed.

cmd/incusd/instance_console.go
presztak: Fixed.

cmd/incusd/instance_console.go
presztak: Fixed.

stgraber (approved):

PR #1507, opened 2024-12-13
Refactor scriptlet loader and improve checks

bensmrs:
The error generation function is pretty heavy, but I think the error messages are way nicer this way.

```
root@test-incus:~# incus config set authorization.scriptlet - <<EOF
def get_project_access(project):
  return ['ben']

get_instance_access = 3
EOF
Error: cannot set 'authorization.scriptlet' to 'def get_project_access(project):
  return ['ben']

get_instance_access = 3
': the function "authorize" is required but has not been found in the scriptlet; additionally, "get_instance_access" should define the scriptlet’s optional function of the same name (found a value of type int instead); finally, the optional function "get_project_access" defines arguments ["project"] (expected ["project_name"])
```

### DIFF
--- a/internal/server/scriptlet/load/load.go
+++ b/internal/server/scriptlet/load/load.go
@@ -1,13 +1,9 @@
 package load
 
 import (
-	"fmt"
-	"slices"
-	"sort"
 	"sync"
 
 	"go.starlark.net/starlark"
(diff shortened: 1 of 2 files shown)


stgraber:
The optional function handling will be quite convenient :)

PR #1513, opened 2024-12-16
incusd/storage/drivers: Force blkdiscard and ignore errors

stgraber:
There can be partitions or other artifacts from snapshots which would normally trip blkdiscard. This is fine in our case and it should proceed regardless.

Also not all devices support discard, so treat it as non-fatal.

### DIFF
--- a/internal/server/storage/drivers/utils.go
+++ b/internal/server/storage/drivers/utils.go
@@ -966,12 +966,9 @@ func clearDiskData(diskPath string, disk *os.File) error {
 	}
 
 	if linux.IsBlockdev(st.Mode()) {
-		// If dealing with a block device, discard its current content.
+		// If dealing with a block device, attempt to discard its current content.
 		// This saves space and avoids issues with leaving zero blocks to their original value.
-		_, err = subprocess.RunCommand("blkdiscard", diskPath)
-		if err != nil {
-			return err
-		}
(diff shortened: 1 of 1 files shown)


internal/server/storage/drivers/utils.go
```diff
@@ -966,12 +966,9 @@ func clearDiskData(diskPath string, disk *os.File) error {
...
 
 	if linux.IsBlockdev(st.Mode()) {
-		// If dealing with a block device, discard its current content.
+		// If dealing with a block device, attempt to discard its current content.
 		// This saves space and avoids issues with leaving zero blocks to their original value.
-		_, err = subprocess.RunCommand("blkdiscard", diskPath)
-		if err != nil {
-			return err
-		}
+		_, _ = subprocess.RunCommand("blkdiscard", "-f", diskPath)
```
breml, line 971: Should this at least be logged just in case? Debugging issues without any trace on what went wrong could be pretty difficult.

hallyn (approved):

internal/server/storage/drivers/utils.go
hallyn: True that could be useful.

stgraber: We could make it a debug level event but not anything more visible than that as this code path is hit quite often.

Issue #1514, opened 2024-12-16
macOS client fails to show VGA console if remote server is addressed via mDNS (*.local) and its IP has changed since the remote was first added


evadne:
# Required information

 * Server: Ubuntu 24.04 + Incus 6.0.0 (from official package)
 * Client: macOS 15.0.1 (24A348)
 * The output of "incus info": Attached below

# Issue description

The Incus client can be instructed to add a remote server such as:

    $ incus remote add hercules hercules.local

However this has only worked once. After the Incus server was modified and its underlying IP changed, the remote will no longer work properly all the time. `incus list` still works but `incus <target> --type=vga` will fail with an error message such as:

    Error: dial tcp: lookup hercules.local: i/o timeout

However if the remote is removed then re-added with the underlying IP address then it would be fine.

This is perplexing as the output of `dns-sd` looks in order:

```
➜  ~ dns-sd -G v4 hercules.local
DATE: ---Mon 16 Dec 2024---
 6:07:09.517  ...STARTING...
Timestamp     A/R  Flags         IF  Hostname                               Address                                      TTL
 6:07:09.644  Add  3             11  hercules.local.                        192.168.255.96                               120
 6:07:09.644  Add  2             11  hercules.local.                        192.168.255.97                               120
```

However the client fails.

When the remote is re-created with the underlying IP address then everything works.

The most perplexing thing is that this had worked for 2 days, then won’t work anymore at all

# Steps to reproduce

 1. Add remote via mDNS name (machine.local)
 2. Cause the remote IP to change via DHCP lease renew etc
 3. Try to open a remote again

# Information to attach

When using mDNS name and things do not work:
 - [x] [Output of `incus info`](https://github.com/user-attachments/files/18145730/incus-info.txt)
 - [x] [Output of the client with --debug](https://github.com/user-attachments/files/18145738/incus-client-debug.txt)
 - [x] [Output of `incus monitor --pretty` while reproducing the issue](https://github.com/user-attachments/files/18145740/incus-monitor-pretty.txt)

When using IP and things work:
 - [x] [Output of the client with --debug](https://github.com/user-attachments/files/18145745/incus-client-debug-good.txt)
 - [x] [Output of `incus monitor --pretty` while reproducing the issue](https://github.com/user-attachments/files/18145748/incus-monitor-pretty-good.txt)

stgraber:
Closing as not an Incus issue.

[...]

PR #1520, opened 2024-12-17
client: allow mDNS resolution when activating console

evadne:
Amends the WebSocket setup code to pass in any existing TLS dialling context, so the Dialer set up on Incus side, alongside with any custom TLS handling logic, including implicit mDNS support by resolving names at the OS level, works consistently

As discussed within #1514

Tested on macOS client vs remote Incus installation over TLS

### DIFF
--- a/client/incus.go
+++ b/client/incus.go
@@ -11,7 +11,6 @@ import (
 	"slices"
 	"strings"
 	"sync"
-	"time"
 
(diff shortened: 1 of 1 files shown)


stgraber:
We already have https://github.com/lxc/incus/pull/1519 which does this :)

Issue #1522, opened 2024-12-17
Issue with qemu 9.2 "Invalid parameter type for 'bootindex'"


cmspam:
When trying to start a VM with qemu 9.2, the below error occurs.

```
localhost:~# incus start debvm
Error: Failed setting up device via monitor: Failed setting up device "eth0": Failed adding NIC device: Invalid parameter type for 'bootindex', expected: integer
Try `incus info --show-log debvm` for more info
```

This is on incus 6.7, using alpine/edge, same on nixos overriding qemu version to 9.2

Has something changed with qemu with bootindex?

# Required information

 * Distribution: Alpine, NixOS
 * Distribution version: edge, unstable
 * The output of "incus info" or if that fails:
   * Kernel version: 6.12.5, 6.6.66
   * LXC version:
   * Incus version: 6.7
   * Storage backend in use: zfs

# Issue description

A brief description of the problem. Should include what you were
attempting to do, what you did, what happened and what you expected to
see happen.

# Steps to reproduce

 1. Install qemu 9.2
 2. Create a VM
 3. Try to start the vm

# Information to attach

 - [ ] Any relevant kernel output (`dmesg`)
 - [ ] Container log (`incus info NAME --show-log`)
 - [ ] Container configuration (`incus config show NAME --expanded`)
 - [ ] Main daemon log (at /var/log/incus/incusd.log)
 - [ ] Output of the client with --debug
 - [ ] Output of the daemon with --debug (alternatively output of `incus monitor --pretty` while reproducing the issue)

bensmrs:
Was able to reproduce. It looks like QEMU now refuses `string` used in place of `integer`s and `uint64`s.
It’s actually very bad, because the change also affects parameters such as `vectors`. Should we convert devices to `map[string]any` to handle it from now on?

stgraber:
Yeah, sounds like we do, which shouldn't be too bad. I'm just hoping that previous QEMU releases didn't expect strings or we'll have a problem :)

bensmrs:
I’ll have a look this week because it’ll definitely impact me next time `unattended-upgrades` runs on my trixie servers

PR #1531, opened 2024-12-18
incusd/instance/qemu: Fix QMP arguments typing

bensmrs:
I was genuinely worried it would be harder
Closes: #1522

### DIFF
--- a/internal/server/instance/drivers/driver_qemu.go
+++ b/internal/server/instance/drivers/driver_qemu.go
@@ -4819,7 +4819,7 @@ func (d *qemu) addGPUDevConfig(cfg *[]cfgSection, bus *qemuBus, gpuConfig []devi
 }
 
 func (d *qemu) addUSBDeviceConfig(usbDev deviceConfig.USBDeviceItem) (monitorHook, error) {
-	device := map[string]any{
+	qemuDev := map[string]any{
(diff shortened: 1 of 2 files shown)


bensmrs:
This gets me wondering… are we *actually* covering VMs in the test suites?

stgraber:
Not through Github Actions as the runners don't universally handle nested virtualization correctly.

We do have a bunch of daily jobs on Jenkins (https://jenkins.linuxcontainers.org) that validate some of those use cases. The tests could be expanded to handle more cases though: https://github.com/lxc/lxc-ci

stgraber (approved):

bensmrs:
> Not through Github Actions as the runners don't universally handle nested virtualization correctly.

Well that’s a pity… I thought I’d read somewhere that Linux runners finally supported it.

> We do have a bunch of daily jobs on Jenkins (https://jenkins.linuxcontainers.org/) that validate some of those use cases. The tests could be expanded to handle more cases though: https://github.com/lxc/lxc-ci

Yeah, but daily jobs are unfortunately not preventing you to merge a bug… and writing tests in two places can get quite hard to track in the long run. Just out of curiosity, have you tried recently to do nested virtualization on Github Actions runners?

Issue #1529, opened 2024-12-18
Allow creation of no-uplink OVN networks with both IPv4 and IPv6 disabled and without an associated logical router.


stevmills:
# Required information

 * Distribution: `Ubuntu`
 * Distribution version: `Ubuntu 22.04.5 LTS`
 * Kernel version: `5.15.0-128-generic`
 * LXC version: `5.0.4`
 * Incus version: `6.8`
 * Storage backend in use: `dir`

# Issue description

Incus already allows creation of OVN networks that specify `network=none`.

This issue is to request a new feature that also allows OVN networks to be created with both IPv4 and IPv6 disabled:

`incus network create my-net network=none --type ovn ipv4.address=none ipv6.address=none`

When both `ipv4.address=none` and `ipv6.address=none` are provided, the expected behavior would be that no associated logical router would get created for the new OVN network.

The use case behind this feature request is an attempt to deploy medium to large virtual environments to Incus using a tool that historically has targeted vSphere instantiations for deployment.

<img width="1148" alt="Screenshot 2024-12-17 at 11 04 28 PM" src="https://github.com/user-attachments/assets/011eefac-b8f5-42ec-afc2-bcf4778ad79c" />

These environments intentionally use real router and firewall OS such Cisco's CSR and ASA virtual appliances and VyOS VMs to enhance realism.

The ability to create OVN networks without logical routers would be required to deploy these types of virtual environments.

winiciusallan:
Hi @stgraber! I'm taking a look to solve this issue. Should change the `State` and `setup` functions to don't create a logical router?

Here, it is also checked if can assign IP addresses to the network

https://github.com/lxc/incus/blob/main/internal/server/network/driver_ovn.go#L2404

stevmills:
Hi @winiciusallan! Thank you for looking at this! Yes, ideally the OVN network would be created with no logical router when both `ipv4.address=none` and `ipv6.address=none` are provided.

stgraber:
@winiciusallan yeah, we'd need to remove that check and skip at least the `CreateLogicalRouter` in such situations, possibly some other router and subnet related calls, though I'd expect most of the subnet stuff to already handle it.

winiciusallan:
Any thought about when `network` is not provided but `ipv4.address=none` and `ipv6.address=none` are?

e.g.
```bash
$ incus network create foo-net network=none ipv4_address=none ipv6_address=none
or
$ incus network create foo-net ipv4_address=none ipv6_address=none
```

[...]

PR #1533, opened 2024-12-20
internal/server: allow creation of OVN network with no uplink

winiciusallan:
Closes: #1529


stgraber (approved):

stgraber:
I did a few small edits to the logic to still allow for a router to be generated when an ipv4 or ivp6 subnet is configured (so we can send out IPv6 RAs) and I tweaked a bit of the cleanup logic for those changing the configuration of an existing network.

Seems to be working well in my tests.

winiciusallan:
Ack! Thanks for your changes, I got it

PR #1535, opened 2024-12-20
Encode mc alias for storage buckets to match allowed characters

monkz:
Encode and prefix the bucket name as mc has requirements for alias naming.
See @ https://github.com/minio/mc/blob/f4dd5e4a07ffe764f2f489a5f42c4079128e405e/cmd/config.go#L174

This allows - for example - to use incus project names with numbers as first character.

### DIFF
--- a/internal/server/storage/s3/miniod/miniod.go
+++ b/internal/server/storage/s3/miniod/miniod.go
@@ -3,13 +3,13 @@ package miniod
 import (
 	"bytes"
 	"context"
+	"encoding/base64"
 	"fmt"
 	"net"
 	"net/url"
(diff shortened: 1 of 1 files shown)


stgraber:
@monkz what does that do for existing buckets?

monkz:
@stgraber  as far I can see, this change is only relevant for temporary entries in /var/lib/incus/mc/config/config.json and temporary files in /var/lib/incus/mc/policy/ .
It does not modify the bucketVol.Name.

stgraber:
Okay, then that should be fine.

stgraber (approved):

PR #1538, opened 2024-12-21
Ceph refactor

### LINKED ISSUE #1023: CephFS mount uses deprecated mount-string, fails to mount Volume based CephFS
# Required information

 * Distribution: NixOS
 * Distribution version: 24.05
   * Kernel version: 6.6.40
   * LXC version: 6.0.1
   * Incus version: 6.2.0
   * Storage backend in use: CephFS

# Issue description

CephFS has changed its mount string in Quincy, the version that has recently reached its estimated EoL date (current being Reef, Squid is upcoming AFAIK).
This means that any still active release (talking about upstream, not distros) has a mount string that is different from the one Incus is using right now.

This leads to users having a really hard time trying to mount CephFS created via the newer CephFS Volumes/Subvolumes mechanic (at least I haven't gotten it working yet).

[As described in the discussion boards](https://discuss.linuxcontainers.org/t/cephfs-storage-add-gives-error-failed-to-mount-using-ceph-no-route-to-host/20211/20?u=benaryorg) the [old syntax](https://docs.ceph.com/en/pacific/man/8/mount.ceph/) was:

```text
[mon1-addr]:3300,[mon2-addr]:3300,[mon3-addr]:3300:/path/to/thing
```

and a lot of options via the `-o` parameter (or the appropriate field in the mount syscall).
Notably Incus does not rely on the config file for this but manually scrapes the mon addresses out of the config file (which has its own issues because the [used string matching](https://github.com/lxc/incus/blob/0739fff2443ac0c8064f0099761b64021e7c0c4f/internal/server/storage/drivers/utils_ceph.go#L80) is insufficient to catch an initial mon list which then refers to the mons by name and the mons being listed in their own sections with their addresses directly as `mon_addr`, which means that while `mount.ceph` can just mount the volume, Incus fails during the parsing phase of the config file.

The [new syntax](https://docs.ceph.com/en/reef/man/8/mount.ceph/) is:

```
user@fsid.cephfsname=/path/to/thing
```

So with the user, the (optional) fsid, and the cephfs name being encoded into the string there are a few less options, although they do still exist.

# Steps to reproduce

1. run CephFS on ≥Quincy
2. create CephFS volume and subvolume
3. try to mount it

With vaguely correct seeming parameters provided to Incus this will still lead to interesting issues like getting *No Route to Host* errors despite everything being reachable.
Honestly, if you find options that manage to mount that, please tell me because I can't seem to find any.

# Information to attach

<details>
  <summary>Any relevant kernel output (dmesg)</summary>

```text
[ +13.628392] libceph: mon0 (1)[2001:db8::1:0]:3300 socket closed (con state V1_BANNER)
[  +0.271853] libceph: mon0 (1)[2001:db8::1:0]:3300 socket closed (con state V1_BANNER)
[  +0.519922] libceph: mon0 (1)[2001:db8::1:0]:3300 socket closed (con state V1_BANNER)
[  +0.520979] ceph: No mds server is up or the cluster is laggy
```
</details>

<details>
  <summary>Main daemon log (at /var/log/incus/incusd.log)</summary>

```text
Jul 19 20:32:09 lxd2 incusd[10412]: time="2024-07-19T20:32:09Z" level=error msg="Failed mounting storage pool" err="Failed to mount \"[2001:41d0:700:2038::1:0]:3300,[2001:41d0:1004:1a22::1:1]:3300,[2001:41d0:602:2029::1:2]:3300:/\" on \"/var/lib/incus/storage-pools/cephfs\" using \"ceph\": invalid argument" pool=cephfs
```
</details>

 - [ ] ~~Container log (`incus info NAME --show-log`)~~
 - [ ] ~~Container configuration (`incus config show NAME --expanded`)~~
 - [ ] ~~Output of the client with --debug~~
 - [ ] ~~Output of the daemon with --debug (alternatively output of `incus monitor --pretty` while reproducing the issue)~~ (doesn't really log anything about the issue)

MadnessASAP:
In continuation of #1473 it is now approaching completion. It passes the full test suite, in addition incusd now can use ceph on my own cluster where previously it could not. 

Also initial tests seem to indicate it fixes #1023: 
```
user@host ~> incus storage show ceph
config:
  cephfs.cluster_name: ceph
  cephfs.path: cephfs/volumes/test-group/test-subvol/e8459ffc-ac7e-41de-bb96-aaef2ee0e8b5
  cephfs.user.name: admin
  source: cephfs/volumes/test-group/test-subvol/e8459ffc-ac7e-41de-bb96-aaef2ee0e8b5
description: ""
name: ceph
driver: cephfs
used_by: []
status: Created
locations:
- none
```

Remaining is to write a few tests to capture some of the additional Ceph superpowers and to squash the whole mess of commits down.

With that in mind, comments and concerns are now very much welcome

### DIFF
--- a/internal/server/device/device_utils_disk.go
+++ b/internal/server/device/device_utils_disk.go
@@ -240,6 +240,12 @@ again:
 
 // diskCephfsOptions returns the mntSrcPath and fsOptions to use for mounting a cephfs share.
 func diskCephfsOptions(clusterName string, userName string, fsName string, fsPath string) (string, []string, error) {
+	// Get the FSID
+	fsid, err := storageDrivers.CephFsid(clusterName)
+	if err != nil {
+		return "", nil, err
+	}
+
 	// Get the monitor list.
 	monAddresses, err := storageDrivers.CephMonitors(clusterName)
 	if err != nil {
@@ -252,14 +258,15 @@ func diskCephfsOptions(clusterName string, userName string, fsName string, fsPat
 		return "", nil, err
 	}
 
-	// Prepare mount entry.
-	fsOptions := []string{
-		fmt.Sprintf("name=%v", userName),
-		fmt.Sprintf("secret=%v", secret),
-		fmt.Sprintf("mds_namespace=%v", fsName),
-	}
+	srcPath, fsOptions := storageDrivers.CephBuildMount(
+		userName,
+		secret,
+		fsid,
+		monAddresses,
+		fsName,
+		fsPath,
+	)
 
-	srcPath := strings.Join(monAddresses, ",") + ":/" + fsPath
 	return srcPath, fsOptions, nil
 }
 
--- a/internal/server/device/device_utils_disk.go
+++ b/internal/server/device/device_utils_disk.go
@@ -30,25 +30,35 @@ const RBDFormatPrefix = "rbd"
 // RBDFormatSeparator is the field separate used in disk paths for RBD devices.
 const RBDFormatSeparator = " "
 
-// DiskParseRBDFormat parses an rbd formatted string, and returns the pool name, volume name, and list of options.
-func DiskParseRBDFormat(rbd string) (string, string, []string, error) {
-	if !strings.HasPrefix(rbd, fmt.Sprintf("%s%s", RBDFormatPrefix, RBDFormatSeparator)) {
-		return "", "", nil, fmt.Errorf("Invalid rbd format, missing prefix")
+// DiskParseRBDFormat parses an rbd formatted string, and returns the pool name, volume name, and map of options.
+func DiskParseRBDFormat(rbd string) (string, string, map[string]string, error) {
+	// Remove and check the prefix.
+	prefix, rbd, _ := strings.Cut(rbd, RBDFormatSeparator)
+	if prefix != RBDFormatPrefix {
+		return "", "", nil, fmt.Errorf("Invalid rbd format, wrong prefix: %q", prefix)
 	}
 
-	fields := strings.SplitN(rbd, RBDFormatSeparator, 3)
-	if len(fields) != 3 {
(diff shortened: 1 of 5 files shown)


MadnessASAP:
Good news! Not dead! Just slow coming back from the holidays. Happy to hear it compiles on LTS @benaryorg. The Ceph logic is good to go, I'm just not yet satisfied with the testing yet, hopefully get that taken care of in the next few days and this will be ready to be (hopefully) merged.

MadnessASAP:
Upon further reflection, I've come to the conclusion that I don't have the necessary time or knowledge to be rooting around within the Incus/GitHub testing framework. The PR as it stands passes the existing tests and works within my homelab for the past couple months. As such I'm confident in saying it does not cause a regression and it does cure a previous pain point.

I'm happy to address specific points of concern but otherwise I'm going to say this ready for merge.

internal/server/device/device_utils_disk.go
```diff
@@ -29,25 +29,36 @@ const RBDFormatPrefix = "rbd"
...
-	}
-
-	fields := strings.SplitN(rbd, RBDFormatSeparator, 3)
-	if len(fields) != 3 {
-		return "", "", nil, fmt.Errorf("Invalid rbd format, invalid number of fields")
-	}
-
-	opts := fields[2]
+// DiskParseRBDFormat parses an rbd formatted string, and returns the pool name, volume name, and map of options.
+func DiskParseRBDFormat(rbd string) (pool string, volume string, opts map[string]string, err error) {
```
stgraber, line 33: We don't use the named return values syntax within the Incus codebase code style (though we apparently don't have a static analysis test for it since that cleared our tests).

MadnessASAP: Fixed!

stgraber (approved):

PR #1544, opened 2024-12-23
incusd/networks: Tweak dnsmasq startup condition

### LINKED ISSUE #1537: incus starts dnsmasq on bridge even if dns.mode none and ipv[46].dhcp false
In `internal/server/network/driver_bridge.go`, Incus gathers invocation flags to pass to a dnsmasq instance, and then makes a stub empty `dnsmasq.raw` config file so incus will not read system `dnsmasq.conf`.

Nowhere is a network's `dns.mode` setting consulted, except to avoid adding `-s`, `--interface-name` and `-S` options to dnsmasq invocation.  The daemon is still started even with `dns.mode=none` and the values of both `ipv4.dhcp=false` and `ipv6.dhcp=false`, it just doesn't get the above flags added to `dnsmasq` invocation in that case.

The daemon does not need to be started when nothing is using it (and explicitly disabled by configuration).  My use case is complete static host db in /etc/hosts, and static ip address, done by cloud-init.

I could use an unmanaged network, but I like the bridge being created and given an IP address by Incus, and also retain option to use managed nat/firewall later.  In the meantime, dnsmasq should not be started, as long as both dns and dhcp are not in use.

Incus v6.7

``` yaml
config:
  dns.mode: none
  ipv4.address: 10.32.192.1/22
  ipv4.dhcp: "false"
  ipv4.firewall: "false"
  ipv4.nat: "false"
  ipv6.address: none
  ipv6.dhcp: "false"
  ipv6.firewall: "false"
  ipv6.nat: "false"
description: ""
name: br0
type: bridge
used_by:
- /1.0/profiles/default
managed: true
status: Created
locations:
- none
project: default
```

alex14641:
dns.mode == none
    ipv4.address == none || ipv4.dhcp == false
    ipv6.address == none

Fixes #1537

### DIFF
--- a/internal/server/network/driver_bridge.go
+++ b/internal/server/network/driver_bridge.go
@@ -634,7 +634,7 @@ func (n *bridge) setup(oldConfig map[string]string) error {
 	}
 
 	// IPv6 bridge configuration.
-	if !slices.Contains([]string{"", "none"}, n.config["ipv6.address"]) {
+	if !util.IsNoneOrEmpty(n.config["ipv6.address"]) {
 		if !util.PathExists("/proc/sys/net/ipv6") {
 			return fmt.Errorf("Network has ipv6.address but kernel IPv6 support is missing")
 		}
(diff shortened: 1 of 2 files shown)


alex14641:
I'm seeing the error "internal/server/network/driver_bridge.go:2694: File is not `gci`-ed with --skip-generated -s standard -s default -s". Is this something I need to fix? If so how?

stgraber:
The `gci` error is usually an indication that you need to run `gofmt -s -w ./`.

I've now done that for you as well as a quick rebase on the main branch.

stgraber (approved):

PR #1557, opened 2025-01-03
Tweak namespace attach logic in forkproxy and forkfile

stgraber:


hallyn (approved):
Not a huge deal, but normally -1 for error and EINVAL are conceptually different...  it could confuse people or cause trouble later. But most likely not so I'll just meege...

PR #1560, opened 2025-01-06
Expand block reset logic

stgraber:
In Incus 6.8 (and 6.0.3), we started performing block discarding prior to writing the initial volume data. This is primarily useful on thick LVM volumes and in some other situations where an existing volume is being reset (some snapshot operations, volume refreshes, ...) in order to avoid prior data from showing up in the resulting volume.

It's particularly problematic on thick LVM as someone unaware of LVM's design may find themselves getting access to deleted data from another instance, effectively causing a data leakage. That's part of LVM's design for thick volumes as they are just an offset on the underlying PVs and not something unique to Incus, but still something that may catch some users off guard.

Our initial implementation of the reset logic works perfectly for file backed storage and appears to work as expected on the majority of storage options tested.

However we've had reports of some environments, primarily multi-device VGs or devices with broken block discard support where the current approach is insufficient.

With this PR, we end up with a much more thorough implementation of block device resetting as we now:
 - Reset file storage the same way we used to
 - Write marker files throughout the device to confirm proper reset
 - Issue a secure discard request and check if successful by checking that the markers are gone
 - If not, issue a regular discard request and again check for our markers
 - If still present, issue a block level zeroing request and check for the markers once again
 - At last, if still not reset, perform a full zero-ing of the device

This should catch all cases and goes from safest and easiest all the way to safe but extremely slow as the ultimate fallback.

Reported-by: Takero Funaki <flintglass@gmail.com>

### DIFF
--- a/internal/linux/discard.go
+++ b/internal/linux/discard.go
@@ -1,0 +1,222 @@
+package linux
+
+import (
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/lxc/incus/v6/shared/subprocess"
+)
+
+// ClearBlock fully resets a block device or disk file using the most efficient mechanism available.
+// For files, it will truncate them down to zero and back to their original size.
+// For blocks, it will attempt a variety of discard options, validating the result with marker files and eventually fallback to full zero-ing.
+func ClearBlock(blockPath string) error {
+	// Open the block device for checking.
+	fd, err := os.OpenFile(blockPath, os.O_RDWR, 0644)
+	if err != nil {
+		if os.IsNotExist(err) {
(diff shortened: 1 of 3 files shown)


hallyn (approved):

takefun:
It seems that clearBlock() is only executed when a volume is created (lvcreate for LVM driver).
we have to wipe the newly allocated blocks after  lvextend/lvresize.

stgraber:
@takefun hmm, good point. Good thing that blkdiscard support being passed an offset, that'd have made things quite a bit more annoying if it didn't...

PR #1561, opened 2025-01-06
Discard blocks on LVM resize

stgraber:


stgraber:
Still need to do a round of manual testing on this one.

stgraber:
Now updated after some manual tests

stgraber:
Thankfully the LVM backend already rounds up everything to the nearest 512 bytes so we should never hit the slow path in the LVM case.

internal/server/storage/drivers/driver_lvm_utils.go
```diff
@@ -573,6 +573,22 @@ func (d *lvm) resizeLogicalVolume(lvPath string, sizeBytes int64) error {
 		return err
 	}
 
+	// On thick pools, discard the blocks in the additional space when the volume is grown.
+	if !d.usesThinpool() {
+		oldSizeBytes, err := d.logicalVolumeSize(lvPath)
```
takefun, line 578: This returns the new LV size here. It should be called before subprocess.TryRunCommand("lvresize", args...)

stgraber: Fixed

hallyn (approved):

PR #1577, opened 2025-01-16
Add new `config` hook

### LINKED ISSUE #1400: Allow QEMU scriptlets to read VM structs and edit QEMU config
`raw.qemu*` configuration keys override the ones defined in previous profiles. Sometimes, I’d like to write a profile that defines several keys in `raw.qemu.conf`, adds some things in `raw.qemu` and does additional processing in `raw.qemu.qmp.scriptlet`.
And sometimes I’d also like to further modify these QEMU options in my VM settings, without having to repeat everything the profile says.
Could we imagine a mechanism allowing, rather than replacing existing data, appending to them for the `raw.qemu`, `raw.qemu.conf` and `raw.qemu.qmp.{early,{post,pre}-start}` keys? (For `raw.qemu.qmp.scriptlet`, it gets trickier and I’m not sure what to propose…)

bensmrs:
Closes #1400

### DIFF
--- a/internal/server/scriptlet/qemu.go
+++ b/internal/server/scriptlet/qemu.go
@@ -2,6 +2,7 @@ package scriptlet
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
 
@@ -41,6 +42,22 @@ func marshalQEMUConf(conf any) ([]map[string]any, error) {
 	return newConf, nil
 }
 
+// unmarshalQEMUConf unmarshals a configuration into a []qemuCfgSection.
+func unmarshalQEMUConf(conf any) ([]qemuCfgSection, error) {
+	jsonConf, err := json.Marshal(conf)
+	if err != nil {
+		return nil, err
+	}
+
+	var newConf []qemuCfgSection
+	err = json.Unmarshal(jsonConf, &newConf)
+	if err != nil {
+		return nil, err
+	}
+
+	return newConf, nil
+}
+
 // QEMURun runs the QEMU scriptlet.
 func QEMURun(l logger.Logger, instance *api.Instance, cmdArgs *[]string, conf *[]cfg.Section, m *qmp.Monitor, stage string) error {
 	logFunc := log.CreateLogger(l, "QEMU scriptlet ("+stage+")")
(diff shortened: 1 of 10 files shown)


bensmrs:
And obviously I forgot to fix the tests :)

bensmrs:
Phew that linter is very conservative!

internal/server/scriptlet/qemu.go
```diff
@@ -15,10 +16,52 @@ import (
...
+	}
+
+	var newConf []map[string]any
+	err = json.Unmarshal(jsonConf, &newConf)
+	if err != nil {
+		return nil, err
+	}
+
+	return newConf, nil
+}
```
stgraber, line 43: Since this isn't in a qemu-specific package (it's in `scriptlet`), the type and function should probably have a `qemu` prefix so it's clear that it's related to QEMU.

internal/server/scriptlet/qemu.go
```diff
@@ -41,6 +42,22 @@ func marshalConf(conf any) ([]map[string]any, error) {
 	return newConf, nil
 }
 
+// unmarshalConf unmarshals a configuration into a []cfgSection.
+func unmarshalConf(conf any) ([]cfgSection, error) {
```
stgraber, line 46: Same with this function.

Actually would it make sense to have those in the `cfg` package instead?

stgraber:
Looks good and was pretty easy to review thanks to how the commits were split.

Only nit is around those marshal/unmarshal and temporary config type in the `scriptlet` package.
I think it'd be better to them be part of the `cfg` package, but if that's impractical, then they should at least mention qemu in their name so someone looking at what's in the `scriptlet` package doesn't get confused.

internal/server/scriptlet/qemu.go
bensmrs: Yeah you’re right. I could even put the whole scriptlet in its own subpackage. We’ve not really been consistent on this, as the “subpackagification” has been the solution to break dependency cycles. Let’s keep it this way and prefix everything with `qemu` for now.

internal/server/scriptlet/qemu.go
bensmrs: I don’t think so. I see no other place where using this function would make sense, so let’s keep it here to really mean that it’s only supposed to be used here!

bensmrs:
> but if that's impractical

It’s not, but the values `marshalConf` give and `unmarshalConf` take are a bit… wonky.
I’m not a huge fan of `map[string]any` values wandering through code for anything other than serialization. Here, `any` means either `string` or `[]string`, which are very different to type-assert and deal with.
Also I don’t see why anyone would use these functions anywhere else when we have nicely typed structs.

> they should at least mention qemu in their name

`qemuMarshalConf`, `marshalQemuConf`, or `marshalQEMUConf`? I prefer the 3rd.

stgraber:
> `qemuMarshalConf`, `marshalQemuConf`, or `marshalQEMUConf`? I prefer the 3rd.

Yeah, 3rd is fine with me.

stgraber (approved):

PR #1587, opened 2025-01-17
Support for ACLs for bridge NIC device when using nftables driver

stgraber:
Closes #1225


stgraber:
@mikerobski so that's my tweaked version of your PR.
I would normally directly push into the PR to avoid having a PR close a PR but because of the configuration of your clone, I can't do that directly.

Anyway, as mentioned in the other one, this one mostly adds:
 - API extension
 - Check that the instance is on the same system prior to trigerring a device update
 - Adds ReloadDevice so we can call that instead of a full Update call
 - Some minor re-titling and rebasing for consistency

mikerobski:
Thanks for cleaning it up, @stgraber !
The PR looks good to me.

PR #1593, opened 2025-01-20
Added unit for minimum ZFS blocksize

Spitfireap:
It was not really clear that 512 without unit meant bytes. 

Regarding ZFS driver, It seems there is no way to set up compression for volumes/datasets. Is this something that might be worth adding, or it has already been attempted ?


stgraber:
> Regarding ZFS driver, It seems there is no way to set up compression for volumes/datasets. Is this something that might be worth adding, or it has already been attempted ?

We respect the setting set on the zpool. If the zpool has `compress=on`, we won't override it with `compress=off`. Additionally, all zpools that we create ourselves has `compress=on` set.

We've not seen much need for per-instance or per-volume compression configuration. Usually it's more of a system policy, as in, are you happy with using a bit more CPU to save space and get faster I/O, so having the configuration done pool-wide makes sense.

stgraber (approved):

Spitfireap:
> We've not seen much need for per-instance or per-volume compression configuration. Usually it's more of a system policy, as in, are you happy with using a bit more CPU to save space and get faster I/O, so having the configuration done pool-wide makes sense.

Understood, thanks. I was looking for the setting since I come from a Truenas environment. Thanks the explanation.

Issue #1393, opened 2024-11-18
Implement a VM memory dump API


stgraber:
It's occasionally useful to have a way to retrieve a full memory dump of a virtual machine.
QEMU makes that pretty easy with the `dump-guest-memory` QMP command combined with the `vmcoreinfo` device.

We could integrate this with:
 - API: `GET /1.0/instances/NAME/debug/memory`
 - CLI: `incus debug get-instance-memory`

SpiffyEight77:
@stgraber 
Hi, I'm interested in working on this issue. Could you please assign it to me?

stgraber:
Done!

stgraber:
Hey @SpiffyEight77, how's it going with this one?

SpiffyEight77:
Hey @stgraber, the feature is almost ready, I'm running local tests now. Will submit the PR for review before Jan 23rd.

stgraber:
Looking forward to it!

kth5:
I am currently attempting to launch a `ppc64le` VM (on POWER9) using `incus` 6.23 and it seems as if, `qemu-system-ppc64` in version 10.2.1 at least, does not have support for the `vmcoreinfo` device.

Launch command:
```
sudo incus launch archpower/2026.04.03 archpower-builder-powerpc64le-vm --vm -c limits.cpu=4 -c limits.memory=4GiB -c limits.memory.hotplug=4GiB
```

`limits.memory.hotplug` is mandatory as otherwise, `maxmemory` may not be set to an amount divisable by 256M as required by `qemu-system-ppc64` by the way.

The machine does not launch with the following error:
```
Error: Failed to run: forklimits limit=memlock:unlimited:unlimited fd=3 -- /usr/bin/qemu-system-ppc64 -S -name archpower-builder-powerpc64le-vm -uuid b80027f2-6d07-401e-ac3d-bb081f65a919 -daemonize -cpu host -nographic -serial chardev:console -nodefaults -no-user-config -sandbox on,obsolete=deny,elevateprivileges=allow,spawn=allow,resourcecontrol=deny -readconfig /run/incus/archpower-builder-powerpc64le-vm/qemu.conf -spice unix=on,disable-ticketing=on,addr=/run/incus/archpower-builder-powerpc64le-vm/qemu.spice -pidfile /run/incus/archpower-builder-powerpc64le-vm/qemu.pid -D /var/log/incus/archpower-builder-powerpc64le-vm/qemu.log -run-with user=incus: : exit status 1
[...]

PR #1602, opened 2025-01-24
incus: Fix example description for debug dump-memory command

SpiffyEight77:
## Description

- incus: Fix example description for debug dump-memory command
- related to PR #1393


stgraber:
Oops, good catch :)

stgraber (approved):

PR #1606, opened 2025-01-25
incus-user: unify logging, support --verbose and --debug

bboozzoo:
Refactor logging setup so that it matches the rest of the code base. Add support for --verbose and --debug, similarly to other commands.

### DIFF
--- a/cmd/incus-user/main.go
+++ b/cmd/incus-user/main.go
@@ -1,19 +1,28 @@
 package main
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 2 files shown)


stgraber (approved):

stgraber:
Looks like the `PreRun` function needs a boilerplate comment.
I'd probably use something like:
```
// PreRun runs immediately prior to the main Run function.
```

bboozzoo:
> ```
> cmd/incus-user/main.go:57:	if err := run(); err != nil {
> ```
> 
> We don't do one-line define+check, so that should be split to:
> 
> ```
> err := run()
> if err != nil {
> ```

Thanks for pushing a fix.

PR #1607, opened 2025-01-25
incus-user: keep track of socket path used to connect to the server

bboozzoo:
The logic used to figure out the correct Unix socket path for access to
 the demon in `client.ConnectIncusUnix()` was fixed in in
 d8b78d5007c01d07b8e7fba78aceb917fbfcf879 to account for use of
 /run/incus/unix.socket as an alternative socket location. However, the
 proxying code of incus-user was not updated when the fix was
 introduced. As a result, even if initial connection to the server and
 the configuration queries were successful, the actual attempt to proxy
 a client request would fail while trying to connect to
 /var/lib/incus/unix.socket regardless of prior findings.

### DIFF
--- a/cmd/incus-user/main_daemon.go
+++ b/cmd/incus-user/main_daemon.go
@@ -63,6 +63,14 @@ func (c *cmdDaemon) Run(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("Unable to connect to the daemon: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


stgraber (approved):

stgraber:
Looks good, I'll re-try the failing test tomorrow and that will get this one merged, thanks!

PR #1611, opened 2025-01-27
Added LZ4 support for incus import

### LINKED ISSUE #1610: Incus import of lz4 compressed files not working
I am trying to import containers that were exported with incus 6.9 using lz4 compression. When I try to re-import the container , I get an error saying it is using an invalid compression despite exporting with lz4. My work around was to decompress it to a tar file and then import.

For now I have to decompress the lz4 exported backup:
`
lz4 -d CATT-Casting-Server-20250125194402.tar.lz4
`
Then I can import the tar file:
`
incus import CATT-Casting-Server-20250125194402.tar CATT-Casting-Server
`

Image

Spitfireap:
Fixes #1610 

I need to actually test it. I cannot retrieve where, in the import process, the backup_file is decompressed, any hint would be appreciated :).


stgraber:
I fixed the magic number (it was backwards), tweaked the tar extract options and confirmed that I can export an lz4 compressed backup and import it back.

stgraber (approved):

PR #1621, opened 2025-02-01
Add `linstor` storage driver

### LINKED ISSUE #564: Add a `linstor` storage driver
Adding this as its own feature request from discussions in #344.

There's been some interest recently in seeing Linstor added as a remote storage driver alongside Ceph and clustered LVM.

The initial things to look at are:

- [ ] Simplest way to deploy a minimal Linstor setup over 3 machines
- [ ] CLI & API way to create or delete what we call a storage pool
- [ ] CLI & API way to create or delete a basic block volume of a defined size
- [ ] CLI & API way to create or delete a snapshot on top of a volume
- [ ] CLI & API way to create or delete a new volume from a snapshot of another volume
- [ ] CLI & API way to handle moving a volume between systems (if any operation is needed for that, usually things like locking, primary host affinity, that kind of stuff)

Go client for Linstor: https://github.com/LINBIT/golinstor

luissimas:
# Description

This PR introduces a new `linstor` storage driver to enable provisioning of volumes backed by [LINSTOR](https://linbit.com/linstor/), as discussed in #564.

All storage driver features are implemented except for storage buckets. Storage buckets cannot be implemented natively on LINSTOR, and the most logical approach would be to fallback to the generic implementation like it's done for all drivers other than Ceph. We didn't go in that route due to the fact that today Incus considers that any remote storage driver supports an optimized implementation of storage buckets, but that's not the case for LINSTOR ( https://github.com/lxc/incus/blob/main/internal/server/storage/backend.go#L3827).

Although most driver features are implemented, not all of them are optimized using LINSTOR native capabilities. We fallback to the generic implementation in several cases, including:

1. Migration across different storage pools
2. Copying volumes with snapshots
3. Backups
4. Refreshing volumes

Closes #564


luissimas:
We expect to send integration tests validating the scenarios in the next few days. I think we'll probably need some Linstor setup on the testing environment as well.

stgraber:
There's a bunch of static analysis failure around code styling/formatting that should be pretty easy to handle.

cmd/incusd/daemon.go
```diff
@@ -2739,3 +2745,40 @@ func (d *Daemon) getOVS() (*ovs.VSwitch, error) {
...
+
+func (d *Daemon) setupLinstor() error {
+	d.linstorMu.Lock()
+	defer d.linstorMu.Unlock()
+
+	// Clear any existing client
+	d.linstor = nil
+
+	// Get the Linstor controller connection string
+	controllerConnection := d.globalConfig.LinstorControllerConnection()
```
bensmrs, line 2757: Maybe we need to be more strict on this one. Setting a connection string without a scheme seems to break the client. So I guess we should enforce `://` to be present in the string.

cmd/incusd/daemon.go
```diff
@@ -2739,3 +2745,40 @@ func (d *Daemon) getOVS() (*ovs.VSwitch, error) {
 
 	return d.ovs, nil
 }
+
+func (d *Daemon) setupLinstor() error {
```
bensmrs, line 2749: This is fine in my setup, but if someone wants to join several pools (for example `https://foo` for `lpool0` and `https://bar` for `lpool1`), there is no way to do it. Isn’t creating a client for each storage pool using LINSTOR preferable?

cmd/incusd/daemon.go
```diff
@@ -2739,3 +2745,40 @@ func (d *Daemon) getOVS() (*ovs.VSwitch, error) {
...
+
+	return nil
+}
+
+func (d *Daemon) getLinstor() (*linstor.Client, error) {
+	// Setup the client if it does not exist
+	if d.linstor == nil {
+		err := d.setupLinstor()
+		if err != nil {
+			return nil, fmt.Errorf("Failed to connect to Linstor: %w", err)
```
bensmrs, line 2779: This adds the `Failed to connect to Linstor: ` prefix twice

internal/server/storage/drivers/driver_linstor_utils.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// createResourceGroup creates a new resource group for the storage pool.
+func (d *linstor) createResourceGroup() error {
+	d.logger.Debug("Creating Linstor resource group")
+
+	// Retrieve the Linstor client
+	linstor, err := d.state.Linstor()
+	if err != nil {
+		return fmt.Errorf("Could not load Linstor client: %w", err)
```
bensmrs, line 96: This adds a prefix that does not really add any information to an already prefixed error

internal/server/storage/drivers/driver_linstor_utils.go
```diff
@@ -0,0 +1,215 @@
...
+
+	return nil
+}
+
+// deleteResourceGroup deleter the resource group for the storage pool.
+func (d *linstor) deleteResourceGroup() error {
+	// Retrieve the Linstor client
+	linstor, err := d.state.Linstor()
+	if err != nil {
+		return fmt.Errorf("Could not load Linstor client: %w", err)
```
bensmrs, line 129: This adds a prefix that does not really add any information to an already prefixed error

internal/server/storage/drivers/driver_linstor_utils.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	return nil
+}
+
+// getResourceDefinition returns a Resource Definition instance for the given Resource name.
+func (d *linstor) getResourceDefinition(resourceDefinitionName string) (*linstorClient.ResourceDefinition, error) {
+	linstor, err := d.state.Linstor()
+	if err != nil {
+		return nil, fmt.Errorf("Could not load Linstor client: %w", err)
```
bensmrs, line 144: This adds a prefix that does not really add any information to an already prefixed error

internal/server/storage/drivers/driver_linstor_utils.go
```diff
@@ -0,0 +1,215 @@
...
+	volType := vol.contentType
+
+	return fmt.Sprintf("%s-%s_%s", d.config[LinstorResourceGroupNameConfigKey], volName, volType)
+}
+
+// getLinstorDevPath return the device path for a given `vol`.
+func (d *linstor) getLinstorDevPath(vol Volume) (string, error) {
+	linstor, err := d.state.Linstor()
+	if err != nil {
+		return "", fmt.Errorf("Could not load Linstor client: %w", err)
```
bensmrs, line 171: This adds a prefix that does not really add any information to an already prefixed error

internal/server/storage/drivers/driver_linstor_utils.go
```diff
@@ -0,0 +1,215 @@
...
+	return volume.DevicePath, nil
+}
+
+// makeVolumeAvailable makes a volume available on the current node.
+func (d *linstor) makeVolumeAvailable(vol Volume) error {
+	l := d.logger.AddContext(logger.Ctx{"volume": vol.Name()})
+	l.Debug("Making volume available on node", logger.Ctx{"volume": vol.Name()})
+	linstor, err := d.state.Linstor()
+	if err != nil {
+		return fmt.Errorf("Could not load Linstor client: %w", err)
```
bensmrs, line 202: This adds a prefix that does not really add any information to an already prefixed error

internal/server/storage/drivers/driver_linstor_volumes.go
```diff
@@ -0,0 +1,398 @@
...
+// filler function.
+func (d *linstor) CreateVolume(vol Volume, filler *VolumeFiller, op *operations.Operation) error {
+	l := d.logger.AddContext(logger.Ctx{"volume": vol.Name()})
+	l.Debug("Creating a new Linstor volume")
+	rev := revert.New()
+	defer rev.Fail()
+
+	linstor, err := d.state.Linstor()
+	if err != nil {
+		return fmt.Errorf("Unable to get the Linstor client: %w", err)
```
bensmrs, line 96: This adds a prefix that does not really add any information to an already prefixed error

internal/server/storage/drivers/driver_linstor_volumes.go
```diff
@@ -0,0 +1,398 @@
...
+}
+
+// DeleteVolume deletes a volume of the storage device.
+func (d *linstor) DeleteVolume(vol Volume, op *operations.Operation) error {
+	l := d.logger.AddContext(logger.Ctx{"volume": vol.Name()})
+	l.Debug("Deleting Linstor volume")
+
+	linstor, err := d.state.Linstor()
+	if err != nil {
+		return fmt.Errorf("Unable to get the Linstor client: %w", err)
```
bensmrs, line 208: This adds a prefix that does not really add any information to an already prefixed error

internal/server/storage/linstor/linstor.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Failed to create Linstor client: %w", err)
+	}
+
+	// Configure the client HTTP transport
+	httpTransport := &http.Transport{}
+
+	// If a CA cert is provided, use it to validate the server certificates
+	if sslCACert != "" {
+		rootCas := x509.NewCertPool()
```
bensmrs, line 37: Nit: `rootCAs` would be better

cmd/incusd/daemon.go
```diff
@@ -2739,3 +2745,40 @@ func (d *Daemon) getOVS() (*ovs.VSwitch, error) {
...
+	// Set the client
+	d.linstor = client
+
+	return nil
+}
+
+func (d *Daemon) getLinstor() (*linstor.Client, error) {
+	// Setup the client if it does not exist
+	if d.linstor == nil {
+		err := d.setupLinstor()
```
bensmrs, line 2777: I’m seeing a small race condition leading to a double execution of `setupLinstor`:
* `getLinstor` gets called until after the `d.linstor == nil` condition is checked
* `setupLinstor` gets called before `err := d.setupLinstor()` gets called
* `setupLinstor` gets called again by `err := d.setupLinstor()`.

I’d put a synchronization here…

internal/server/storage/drivers/driver_linstor_utils.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	resourceName := d.getResourceDefinitionName(vol)
+	err = linstor.Client.Resources.MakeAvailable(context.TODO(), resourceName, d.state.ServerName, linstorClient.ResourceMakeAvailable{
```
bensmrs, line 206: This implies that the current server node name is a valid LINSTOR node, which may be wrong

luissimas: Yeah, the matching of node names between Incus and LINSTOR is an underlying assumption. I'm not sure if there's a way around that, I personally don't see one right now. Although this was included in the driver documentation, I'd like to enforce that somehow to make it clearer to users if there's any mismatch between LINTOR's and Incus's views of the cluster nodes.

The reason for that is that AFAIK LINSTOR requires a target satellite node name when making volumes available on nodes. The "making volumes available on nodes" part is required because, unlike Ceph, volumes are not automatically available on all cluster nodes.

If we don't find a way to work around that limitation (which AFAIK is a fundamental limitation due to how LINSTOR works), I propose that we at least do our best to make those assumptions clear to users and validate them at runtime. I think Incus should be capable to validate that the following are true:

1. All Incus nodes are running the `linstor-satellite` service and have the DRBD9 kernel module loaded
2. The set of Incus node names is a subset of the set of LINSTOR satellite node names

With that said, it's not clear to me what the behavior should be if any of those are violated. Should we refuse the loading of the `linstor` driver and crash Incus? A warning log message is enough? I think that's very open to discussion.

cmd/incusd/daemon.go
```diff
@@ -2739,3 +2745,40 @@ func (d *Daemon) getOVS() (*ovs.VSwitch, error) {
...
+	d.linstor = nil
+
+	// Get the Linstor controller connection string
+	controllerConnection := d.globalConfig.LinstorControllerConnection()
+
+	// Get the SSL certificates if needed
+	sslCACert, sslClientCert, sslClientKey := d.globalConfig.LinstorSSL()
+
+	// Get Linstor client
+	client, err := linstor.NewClient(controllerConnection, sslCACert, sslClientCert, sslClientKey)
```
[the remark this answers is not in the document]
luissimas: Good catch! I suspect this is from the way that the Go client handles the URL being passed wrapped in a `BaseURL` function. I think we could easily fix that while also adding support for specifying a list of controller URLs separated by commas.

internal/server/storage/drivers/driver_linstor.go
```diff
@@ -0,0 +1,232 @@
...
+	d.config["volatile.initial_source"] = d.config["source"]
+
+	// Fill default config values
+	err := d.FillConfig()
+	if err != nil {
+		return fmt.Errorf("Could not create Linstor storage pool: %w", err)
+	}
+
+	// If a source is provided, override the resource group name
+	if d.config["source"] != "" {
```
bensmrs, line 107: So this part of the code leads to a very tricky bug in the tests: *every* other storage driver sets the `source` key, and it’s tested in the clustering tests. I suggest doing the following to both make the tests work and be consistent with the other drivers:

```go
if d.config["source"] == "" {
    d.config["source"] = d.config[LinstorResourceGroupNameConfigKey]
} else {
    d.config[LinstorResourceGroupNameConfigKey] = d.config["source"]
}
```

To be fair, I’m not convinced we need LinstorResourceGroupNameConfigKey when we already have `source`…

bensmrs: This is a blocking point for my patch of the test suite, so I’m waiting for this comment to be handled

luissimas: I've added an extra check for the consistency between the two config options. Also, if both `LinstorResourceGroupNameConfigKey` and `source` are not set, we fallback to the storage pool name and also set the `source` key to that value. This should match the ceph driver behavior 1:1

cmd/incusd/daemon.go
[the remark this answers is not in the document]
luissimas: We now have support for multiple URLs, and now a missing URL scheme should not crash the client. We now use the `Controllers` function to pass the URLs to the client, which correctly handles missing url schemes and ports. Before that, we were using the `BaseURL` function, which expects an already parsed URL and does not add any default values.

`BaseURL`: https://github.com/LINBIT/golinstor/blob/master/client/client.go#L118
`Controllers`: https://github.com/LINBIT/golinstor/blob/master/client/client.go#L174

cmd/incusd/daemon.go
luissimas: I don't think there's a way to support connections to different controllers simultaneously because LINSTOR itself does not support having a given satellite managed by different controllers. In this example I've deployed an additional `linstor-controller` service with its own database on `server02`, added the satellite nodes and then stopped and started the service to see the impact on `server01`, which had the original `linstor-controller`.

```
root@server01:~# linstor node list
╭─────────────────────────────────────────────────────────────╮
┊ Node     ┊ NodeType  ┊ Addresses                   ┊ State  ┊
╞═════════════════════════════════════════════════════════════╡
┊ server01 ┊ SATELLITE ┊ 10.172.117.216:3366 (PLAIN) ┊ Online ┊
┊ server02 ┊ SATELLITE ┊ 10.172.117.181:3366 (PLAIN) ┊ Online ┊
┊ server03 ┊ SATELLITE ┊ 10.172.117.134:3366 (PLAIN) ┊ Online ┊
╰─────────────────────────────────────────────────────────────╯
root@server02:~# systemctl start linstor-controller
root@server01:~# linstor node list
╭────────────────────────────────────────────────────────────────────────────────╮
┊ Node     ┊ NodeType  ┊ Addresses                   ┊ State                     ┊
╞════════════════════════════════════════════════════════════════════════════════╡
┊ server01 ┊ SATELLITE ┊ 10.172.117.216:3366 (PLAIN) ┊ OFFLINE(OTHER_CONTROLLER) ┊
┊ server02 ┊ SATELLITE ┊ 10.172.117.181:3366 (PLAIN) ┊ OFFLINE(OTHER_CONTROLLER) ┊
┊ server03 ┊ SATELLITE ┊ 10.172.117.134:3366 (PLAIN) ┊ OFFLINE(OTHER_CONTROLLER) ┊
╰────────────────────────────────────────────────────────────────────────────────╯
```

Note that a single controller still allows us to create different storage pools on LINSTOR (one for HDDs and one for NVMEs, for example). We can then use the `linstor.resource_group.storage_pool` to specify that an Incus storage pool should allocate volumes on a given set of LINSTOR storage pools.

bensmrs:
I’ve done a lot of grunt work on the commits to simplify further reviews. Renaming is now working because we no longer store the actual volume names in LINSTOR.

bensmrs:
`incus copy` is now working. The tests now fail on `incus publish`.

bensmrs:
Well well well…
```
Setting up drbd-dkms (9.2.12-1ppa1~jammy1) ...
```
and then
```
DRBD: DRBD version has to be >= 9. Current DRBD version: 8.4.11
```

I’m confused.

bensmrs:
At least the “““clustered””” tests work now :)

bensmrs:
Looks like @luissimas and I can’t reproduce the standalone test error on our machines…

bensmrs:
This is causing problems: https://github.com/lxc/incus/blob/90f62abafc71ea749c3d0a4fba375cf76f815053/test/suites/remote.sh#L255-L272

bensmrs:
Good catch :)

bensmrs:
`rsync` seems to have some trouble here. Attaching `strace` gives the following:
```
> strace -p "$(pgrep -f rsync)" -v -s 9999
strace: Process 63622 attached
pselect6(1, [0], NULL, [0], {tv_sec=59, tv_nsec=968240958}, NULL) = 0 (Timeout)
pselect6(1, [0], NULL, [0], {tv_sec=60, tv_nsec=0}, NULL) = 0 (Timeout)
pselect6(1, [0], NULL, [0], {tv_sec=60, tv_nsec=0}, NULL) = 1 (in [0], left {tv_sec=59, tv_nsec=616914004})
read(0, "", 4)                          = 0
write(2, "rsync: connection unexpectedly closed (0 bytes received so far) [Receiver]\n", 75) = 75
rt_sigaction(SIGUSR1, {sa_handler=SIG_IGN, sa_mask=[], sa_flags=SA_RESTORER, sa_restorer=0x7f9e06c49da0}, NULL, 8) = 0
rt_sigaction(SIGUSR2, {sa_handler=SIG_IGN, sa_mask=[], sa_flags=SA_RESTORER, sa_restorer=0x7f9e06c49da0}, NULL, 8) = 0
write(2, "rsync error: error in rsync protocol data stream (code 12) at io.c(232) [Receiver=3.4.0]\n", 89) = 89
clock_nanosleep(CLOCK_REALTIME, 0, {tv_sec=0, tv_nsec=100000000}, 0x7ffd125ec6e0) = 0
exit_group(12)                          = ?
+++ exited with 12 +++
```

bensmrs:
I may be totally wrong, but I think we’re failing the condition at https://github.com/lxc/incus/blob/e9e4aca5dd1db0791f2f11402498077e6be37cfb/internal/server/instance/drivers/driver_lxc.go#L5623.
I think the negotiated migration type is LINSTOR, so the LXC driver does not enable `MultiSync` when sending the container. In the end, we have a first round of `Recv`/`Send`, but during the final step, only `Recv` is run.

bensmrs:
There’s still quite some rebasing work left to do, but we’re very close now :)
Let’s hope for a ready-to-review version in a few hours.

luissimas:
@stgraber marking this as ready for review. As mentioned in the description the implementation covers most of the features. The docs also include a guide on how to set up LINSTOR and integrate it with Incus, as well as a few notes on the internal implementation that should be useful for operators.

We're still investigating an intermittent error that occurred in two CI jobs, but since it's a bit hard to reproduce and the fix should not result in a big diff we don't think it's a block for starting the review. 

https://github.com/lxc/incus/actions/runs/13883608381/job/38845351623
https://github.com/lxc/incus/actions/runs/13900201207/job/38889870211

stgraber:
Nice. This is one that's going to take me a little while to review especially as I've got the address set PR that's going to near a bit of quality time from me for both review and some changes.

But I'm hoping to get to this one later this week so we have it merged quite a few days before the next release.

bensmrs:
Great! As you’ll see, the PR touches the codebase at plenty of unexpected places; don’t hesitate to ask for our motivations if it isn’t obvious.
Oh, and @luissimas did some aggressive tests with LINSTOR and just confirmed that the problem we encountered with tests hanging on snapshot creation is reproducible using only LINSTOR tools, so it’s a good sign for our driver. We’ll tell the LINBIT people as we get more debugging material.

bensmrs:
Oh we switched to gofumpt…

stgraber:
I'm expecting to be done with the address set review and merge tomorrow at which point I'll turn my focus on this one. I'm hopeful that everything will look good and only minor tweaks will be needed.

I'd really like to have this in 6.11 which is due for tagging late on Thursday.
Not helping matters, I'm at a conference Monday through Wednesday, so will work with quite limited quiet time for complex reviews :(

stgraber:
So far I did a quick rebase on this one mostly to pull out all the changes that are not directly from this branch and re-title some of the other commits. I'll start by reviewing the changes that aren't specific to linstor and upstream those separately to shrink this branch a bit and make the DCO check happy.

bensmrs:
Oof the linter is very aggressive now :)
Please ping me when {you want / are ready for} me to fix the linter warnings and rebase, so that we don’t lose too much time on it.

stgraber:
I now opened https://github.com/lxc/incus/pull/1842 which I'll let go through tests and merge. Then I can rebase this branch on it, getting us rid of a bunch of commits. After that I'll start reviewing this commit by commit.

I'd recommend refraining from pushing anything while I work my way through this as I usually just push the smaller tweaks / typo fixes and the like directly into your branch rather than go through review comments + waiting for you to fix it.

stgraber:
Rebased, re-ordered and cleaned up. Ready for commit by commit review. I'll get on that this evening once I'm back from a conference dinner here (ironically enough, at a storage related conference ;))

bensmrs:
> Rebased, re-ordered and cleaned up. Ready for commit by commit review.

That’s very nice to read!

> I'll get on that this evening once I'm back from a conference dinner here (ironically enough, at a storage related conference ;))

Have fun!

bensmrs:
Btw, not sure why you removed the LINSTOR migration line from the protobuf generator

stgraber:
> Btw, not sure why you removed the LINSTOR migration line from the protobuf generator

I split it out into its own commit. I always keep the generated commits separate from manual changes.

stgraber:
Oh, I see, there's a bad rebase on it, let me fix that now.

doc/reference/storage_ceph.md
```diff
@@ -101,7 +101,7 @@ Key                     | Type      | Condition                 | Default
 `block.filesystem`      | string    | block-based volume with content type `filesystem` | same as `volume.block.filesystem`              | {{block_filesystem}}
 `block.mount_options`   | string    | block-based volume with content type `filesystem` | same as `volume.block.mount_options`           | Mount options for block-backed file system volumes
 `initial.gid`           | int       | custom volume with content type `filesystem`  | same as `volume.initial.uid` or `0`           | GID of the volume owner in the instance
-`initial.mode`          | int       | custom volume with content type `filesystem`  | same as `volume.initial.mode` or `711`        | Mode  of the volume in the instance
+`initial.mode`          | int       | custom volume with content type `filesystem`  | same as `volume.initial.mode` or `711`        | Mode of the volume in the instance
```
stgraber, line 104: Should be split out as separate commit for backporting.
I'll send that one through my branch now and remove it from this one.

stgraber:
I started the commit by commit review, hoping to get through that over the next few hours.

internal/server/storage/linstor/linstor.go
```diff
@@ -0,0 +1,69 @@
...
+// Client represents an HTTP Linstor client.
+type Client struct {
+	Client *linstorClient.Client
+}
+
+// NewClient initializes a new Linstor client.
+func NewClient(controllerConnection, sslCACert, sslClientCert, sslClientKey string) (*Client, error) {
+	logger.Info("Creating new Linstor client", logger.Ctx{"controllerConnection": controllerConnection})
+
+	// Configure the client HTTP transport
```
stgraber, line 54: Nit, comments should end with a period. I'll fix those.

test/suites/incremental_copy.sh
```diff
@@ -30,7 +30,7 @@ do_copy() {
   incus storage volume set "${source_pool}" container/c1 user.foo=main
 
   # Set size to check this is supported during copy.
-  incus config device set c1 root size=50MiB
+  incus config device set c1 root size=200MiB
```
stgraber, line 33: We should get the non-linstor changes out of here so they can be pushed to LTS and reduce the delta (even if only really needed under Linstor).

stgraber:
Done with my review and edits.
I'm going to wait for #1846 to land, then rebase this one and we can land it.

Thanks everyone who worked on this, this looks great!

stgraber (approved):

PR #1626, opened 2025-02-03
incusd/instance/drivers: Improve NUMA balancing

lnutimura:
This PR fixes the assignment of `volatile.cpu.nodes` when `limits.cpu` is used alongside `limits.cpu.nodes=balanced`, and `limits.cpu` exceeds the number of CPUs available per NUMA node.

Previously, `volatile.cpu.nodes` was always set to a single NUMA node, even when more were required. With this fix, it can now span multiple NUMA nodes when necessary.

### DIFF
--- a/internal/server/instance/drivers/driver_lxc.go
+++ b/internal/server/instance/drivers/driver_lxc.go
@@ -1913,9 +1913,9 @@ func (d *lxc) startCommon() (string, []func() error, error) {
 	revert := revert.New()
 	defer revert.Fail()
 
-	// Assign a NUMA node if needed.
+	// Assign NUMA node(s) if needed.
 	if d.expandedConfig["limits.cpu.nodes"] == "balanced" {
-		err := d.setNUMANode()
+		err := d.balanceNUMANodes()
 		if err != nil {
 			return "", nil, err
 		}
--- a/internal/server/instance/drivers/driver_qemu.go
+++ b/internal/server/instance/drivers/driver_qemu.go
@@ -1205,9 +1205,9 @@ func (d *qemu) start(stateful bool, op *operationlock.InstanceOperation) error {
 
 	defer op.Done(err)
 
-	// Assign a NUMA node if needed.
+	// Assign NUMA node(s) if needed.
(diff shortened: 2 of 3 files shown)


internal/server/instance/drivers/driver_lxc.go
```diff
@@ -1913,7 +1913,7 @@ func (d *lxc) startCommon() (string, []func() error, error) {
 	revert := revert.New()
 	defer revert.Fail()
 
-	// Assign a NUMA node if needed.
+	// Assign NUMA node(s) if needed.
 	if d.expandedConfig["limits.cpu.nodes"] == "balanced" {
 		err := d.setNUMANode()
```
lnutimura, line 1918: Right now, `setNUMANode()` is only called when `limits.cpu.nodes` is set to `balanced`.

In the future, could this function be called with `limits.cpu.nodes` having a different value?
If not, then I think we can remove some of the redundant checks within `setNUMANode`.

internal/server/instance/drivers/driver_qemu.go
```diff
@@ -1205,7 +1205,7 @@ func (d *qemu) start(stateful bool, op *operationlock.InstanceOperation) error {
 
 	defer op.Done(err)
 
-	// Assign a NUMA node if needed.
+	// Assign NUMA node(s) if needed.
 	if d.expandedConfig["limits.cpu.nodes"] == "balanced" {
 		err := d.setNUMANode()
```
lnutimura, line 1210: Same here.

internal/server/instance/drivers/driver_lxc.go
stgraber: Looks like it's indeed limitation to balanced for now.
If simplifying it, we'd probably want to rename it to make that clearer.

stgraber:
I've renamed the function to `balanceNUMANodes` and removed the `balanced` check from within the function. That allows flattening it a bit. Otherwise I've fixed the static analysis issue so we should be good to go now.

stgraber (approved):

Issue #1625, opened 2025-02-03
Isolated OVN Networks unable to query using info


nsanity2097:
When leveraging network: none and removing dhcp/nat - incus no longer provides info on the network.

```
C:\Users\Geoff_Killen\Downloads>incus version
Client version: 6.9
Server version: 6.9
```

```
C:\Users\gk\Downloads>incus network show ovn-isolated
config:
  bridge.mtu: "1442"
  ipv4.address: 10.181.150.1/24
  ipv4.dhcp: "false"
  ipv4.nat: "false"
  ipv6.address: fd42:be6f:42b1:121::1/64
  ipv6.dhcp: "false"
  ipv6.nat: "false"
  network: none
description: ""
name: ovn-isolated
type: ovn
used_by:
- /1.0/instances/o1
- /1.0/instances/o2
- /1.0/instances/o3
- /1.0/instances/top-elephant
- /1.0/instances/win2025vm
- /1.0/profiles/new
- /1.0/profiles/opnsense
- /1.0/profiles/win2025
managed: true
status: Created
locations:
- ubu-node-0
- ubu-node-1
project: default

C:\Users\gk\Downloads>incus network info ovn-isolated
Error: object not found
```
Regular networks seem fine

```
C:\Users\gk\Downloads>incus network info ovn1
Name: ovn1
MAC address: 00:16:3e:2d:ef:3b
MTU: 1442
State: up
Type: broadcast

IP addresses:
  inet  10.253.13.1/24 (link)
  inet6 fd42:d916:dbfb:e7f3::1/64 (link)

Network usage:
  Bytes received: 0B
  Bytes sent: 0B
  Packets received: 0
  Packets sent: 0

OVN:
  Chassis: ubu-node-0
  Logical router: incus-net14-lr
  IPv4 uplink address: 10.116.146.2
  IPv6 uplink address: fd42:7824:c152:4341:216:3eff:fe2d:ef3b

C:\Users\gk\Downloads>incus network show ovn1
config:
  bridge.mtu: "1442"
  ipv4.address: 10.253.13.1/24
  ipv4.dhcp: "false"
  ipv4.l3only: "false"
  ipv4.nat: "true"
  ipv6.address: fd42:d916:dbfb:e7f3::1/64
  ipv6.nat: "true"
  network: incusbr1
  volatile.network.ipv4.address: 10.116.146.2
  volatile.network.ipv6.address: fd42:7824:c152:4341:216:3eff:fe2d:ef3b
description: ""
[...]

PR #1628, opened 2025-02-03
incus/network/info (ovn): Fix object not found.

rxtom:
Fix for lxc/incus issue #1625


stgraber:
I did a minor change to the approach so we can retain most of the error handling.

stgraber (approved):

PR #1655, opened 2025-02-13
Implement `smbios11` config keys

stgraber:


stgraber:
@poettering

internal/server/instance/drivers/driver_qemu.go
```diff
@@ -1627,6 +1627,14 @@ func (d *qemu) start(stateful bool, op *operationlock.InstanceOperation) error {
 	// SMBIOS only on x86_64 and aarch64.
 	if d.architectureSupportsUEFI(d.architecture) {
 		qemuArgs = append(qemuArgs, "-smbios", "type=2,manufacturer=LinuxContainers,product=Incus")
+
+		for k, v := range d.expandedConfig {
+			if !strings.HasPrefix(k, "smbios.") {
+				continue
+			}
+
+			qemuArgs = append(qemuArgs, "-smbios", fmt.Sprintf("type=11,value=%s=%s", strings.TrimPrefix(k, "smbios."), v))
```
poettering, line 1636: i am not a go person, but this probably needs some escaping, see the comment on https://www.qemu.org/docs/master/system/qemu-manpage.html#options → you need to replace `,` by `,,` in the strings you insert here.

stgraber: Ah yeah, good point. Looking at the QEMU doc, comma seems to be the only real problem here so that's an easy one.

poettering: (btw, i think there might be other places you need to add this, unrelated to the smbios stuff. glancing over the code there is at least one `-drive` and one `-chardev` expression added to the qemu cmdline further down in the code where it might be wise to escape the `,` to be fully correct. Didn't do a comprehensive scan though, might be more.)

poettering:
so i am neither a go hacker nor an incus hacker, hence my review ain't carry weight, but fwitw the relevant bits look fine to me! thank you for working on this, and so quickly!

internal/server/instance/drivers/driver_qemu.go
stgraber: I've added a small helper that does it now and also found only those two extra calls (-drive and -chardev) that may be affected (though they're passed paths that are under our control so unlikely to contain a comma to begin with).

stgraber:
Per @hallyn comment elsewhere, I've changed this to use `smbios11` as the config key space in case we ever want to directly allow users to mess with other tables (we internally mess with some already).

hallyn (approved):

Issue #1629, opened 2025-02-03
incus copy --refresh fails with "Name already in use" when using LVM storage backend


dennisoderwald:
I encountered an issue when trying to use the --refresh flag with incus copy. The command fails with the error:

```
incus copy container-name container-backup --target node2 --refresh
Error: Name "container-backup" already in use
```

**Steps to reproduce:**

Create an instance container-name on the node1.

Perform an initial copy to the target with:
```
incus copy container-name container-backup --target node2
```

(I also tried it with --refresh initial)

Attempt to refresh the instance with:
```
incus copy container-name container-backup --target node2 --refresh
```

→ This fails with "Name already in use".

**System Information**

Client Version: 6.9
Server Version: 6.9
Storage Backend: LVM

**Expected Behavior:**

The --refresh flag should update the existing instance rather than treating it as a duplicate.

Does the LVM storage backend support the --refresh functionality for incus copy? If not, is there any workaround available?

I could not find any restriction regarding LVM on the [official documentation](https://linuxcontainers.org/incus/docs/main/reference/manpages/incus/copy/), which might explain the problem.

Would appreciate any insights or confirmation on whether this is a known limitation or a bug.

Thanks!

PatrickHuetter:
I can confirm this issue. Same problem here.

PR #1658, opened 2025-02-15
Fix instance copy error when using '--refresh' flag

presztak:
Copying an instance with the `--refresh` flag sometimes fails due to instance name validation errors.

Do not set the `Name` field in `api.InstancePost` or `api.InstanceSnapshotPost` structs, as this field is meant only for renaming instances.

Fixes: #1629

### DIFF
--- a/cmd/incusd/instance_snapshot.go
+++ b/cmd/incusd/instance_snapshot.go
@@ -649,25 +649,21 @@ func snapshotPost(s *state.State, r *http.Request, snapInst instance.Instance) r
 
 	parentName, snapName, _ := api.GetParentAndSnapshotName(snapInst.Name())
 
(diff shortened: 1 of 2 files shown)


presztak:
@stgraber some tests related to copying snapshots are still failing. One option is to keep the changes as they are but only perform this check https://github.com/lxc/incus/blob/64e9144a9d5baa1df8da1e05ae4b4adf40d17f19/cmd/incusd/instance_snapshot.go#L669-L671 when `migration=false`. Alternatively, since `--refresh` can only be used with instances, we could keep the Name field in the `api.InstanceSnapshotPost` struct. What do you think?

stgraber (approved):

PR #1661, opened 2025-02-16
Fix docs for load balancer create backend

gwenya:
The option "listen port" does not exist on backends, only on ports. The sentence about what happens if no target port is specified doesn't make any sense I think, but I also don't know what the actual behaviour in that case is.

### DIFF
--- a/doc/howto/network_load_balancers.md
+++ b/doc/howto/network_load_balancers.md
@@ -65,11 +65,11 @@ The backend target address must be within the same subnet as the network that th
 Use the following command to add a backend specification:
 
 ```bash
-incus network load-balancer backend add <network_name> <listen_address> <backend_name> <listen_ports> <target_address> [<target_ports>]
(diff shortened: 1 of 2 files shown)


doc/howto/network_load_balancers.md
```diff
@@ -65,11 +65,10 @@ The backend target address must be within the same subnet as the network that th
 Use the following command to add a backend specification:
 
 ```bash
-incus network load-balancer backend add <network_name> <listen_address> <backend_name> <listen_ports> <target_address> [<target_ports>]
+incus network load-balancer backend add <network_name> <listen_address> <backend_name> <target_address> [<target_ports>]
 ```
 
 The target ports are optional.
-If not specified, the load balancer will use the listen ports for the backend for the backend target ports.
```
stgraber, line 72: If not specified, the load balancer defaults to using the frontend port as the backend target port.

stgraber (approved):

PR #1665, opened 2025-02-18
Replace ast.Package with types.Package

breml:
Replace the deprecated `ast.Package` with `types.Package`.

Change the approach, how `go:generate` is used reducing the number of calls from n to 1 per package, which gives a significant speedup of the whole processing. With this change, the additional time for the loading of the `types.Package` information is more than compensated. On my machine, generating the database statements and the methods in `internal/server/db/cluster` is roughly twice as fast as without this PR.


stgraber:
Just made some minor edits to the commit structure and titles.

PR #1668, opened 2025-02-20
Add provider for DNS-01 ACME challenge

### LINKED ISSUE #1364: ACME - Add support for DNS-01 validation
Hi there! New to Incus, and having a great time setting up my first lab cluster. Most recently, I was able to get `incus-ui-canonical` working, and after getting that all sorted, I started to look into getting certificates cut for it. Found the built in ACME support (awesome!) but discovered that it only supported HTTP-01 validation. I generally prefer to use DNS-01 validation, as I often run services that are not exposed to the Internet, and the Incus management interface would definitely be one of those interfaces I'd prefer to not expose. 

Adding support for DNS-01 validation I imagine would be a relatively easy lift, as the `lego` module already has support to do this. I'm not a well versed Go programmer by any stretch, but I might be willing to give this a shot myself.

accuser:
Preliminary PR for addition of DNS-01 challenge provider.

Example configuration:

```yaml
config:
  acme.agree_tos: true
  acme.domain: foo.example.net
  acme.email: admin@foo.example.net
  acme.challenge: DNS-01
  acme.provider: cloudflare
  acme.provider.environment: |-
    CLOUDFLARE_EMAIL=admin@foo.example.net
    CLOUDFLARE_API_KEY=XYZ
```

Closes #1364


belthesar:
This looks pretty good! Thank you for doing this! I do have some feedback: 

Per @stgraber's feedback here: https://github.com/lxc/incus/issues/1364#issuecomment-2670125307 , we should probably consider nestling config in keys specific to the challenge. My intention was to pattern this off of Traefik's config, an example found here: https://doc.traefik.io/traefik/https/acme/#configuration-examples 
Given what you've done, I'd probably create a map something like: 
```yaml
acme:
  agree_tos: true
  challenge: "DNS-01"
  dnsChallenge:
    provider: "ProviderName"
    config: |-
      ENVVAR_ONE="value"
      ENVVAR_TWO="value"
    resolvers: "1.1.1.1,1.0.0.1"
```
This would set up a solid pattern if additional configuration options for challenges become available as well.

accuser:
I think you're right, but I'd probably fall towards:

```yaml
config:
  acme.challenge: DNS-01
  acme.dns:
    provider: cloudflare
    resolvers: 
    - 1.1.1.1
    - 1.0.0.1
    env: |-
      CLOUDFLARE_EMAIL=admin@foo.example.net
      CLOUDFLARE_API_KEY=XYZ
```

I'm favouring `env` for the environment variables, rather than `config`, and a list of strings for `resolvers`.

stgraber:
Going to take a quick look at this one, for one thing, do a rebase and tweak the commit list :)

stgraber:
Waiting for #1679 to merge, then this one can get rebased and will be good to merge.
I've just tested it here with a couple of DNS providers and it seems to be working as expected.

stgraber (approved):

Issue #1657, opened 2025-02-13
incus-migrate import failure when profile has a custom block volume attached.


Taverius:
With a profile like so:
```yaml
config: {}
description: Test Profile
devices:
  openwrt_opt:
    pool: incus-zfs
    source: openwrt_opt
    type: disk
name: test
used_by: []
project: default
```

Where `openwrt_opt` is a 4GiB block device with an ext4 fs, if that matters.

`incus-migrate` state at import:
```yaml
Instance to be created:
  Name: OpenWRT-Test
  Project: default
  Type: virtual-machine
  Source: openwrt-24.10.0-x86-64-generic-ext4-combined-efi.img
  Source format: raw
  Profiles:
  - test
  Storage pool: incus-zfs
  Config:
    security.secureboot: "false"
```

Fails like so:

```
Please pick one of the options above [default=1]:
Transferring instance: OpenWRT-Test: 113.91MB (28.47MB/s)Error: Failed to prepare device "openwrt_opt" for migration: Not supported
```

Instance is not in `incus list`, but when exists in storage pool:

`❯ incus storage show incus-zfs`
```yaml
config:
  source: incus
  volatile.initial_source: /dev/disk/by-partlabel/disk-fast-incus
  zfs.pool_name: incus
description: ZFS on Micron 7450 800GB PCIe4.0x4
name: incus-zfs
driver: zfs
used_by:
- /1.0/images/7cc853ea6eff448c2ed87172fabde4f1b2614b3c4f68205a30b34a321dd4c2f5
- /1.0/instances/OpenWRT-Test
- /1.0/instances/alpine-vm
- /1.0/profiles/default
- /1.0/profiles/test
- /1.0/storage-pools/incus-zfs/volumes/custom/backups
- /1.0/storage-pools/incus-zfs/volumes/custom/images
- /1.0/storage-pools/incus-zfs/volumes/custom/openwrt_opt
status: Created
locations:
- none
```

`❯ incus storage volume list incus-zfs`
```
+-----------------+------------------------------------------------------------------+----------------------------+--------------+---------+
|      TYPE       |                               NAME                               |        DESCRIPTION         | CONTENT-TYPE | USED BY |
+-----------------+------------------------------------------------------------------+----------------------------+--------------+---------+
| custom          | backups                                                          | Backup Storage on ZFS      | filesystem   | 0       |
+-----------------+------------------------------------------------------------------+----------------------------+--------------+---------+
[...]

PR #1669, opened 2025-02-20
Check if disk is remote when migrating with an extra disk

presztak:
This additional check ensures that migrations are not performed on local disks. On remote disks, the migration effectively only mounts the disk on the target node.

Fixes: #1657


presztak:
@stgraber This still doesn't resolve the issue of importing an instance with `incus-migrate` when a remote disk is attached via a profile. Should we add an additional check, or is this sufficient for now?

stgraber:
I "think" this should be sufficient.

PR #1679, opened 2025-02-21
Add filtering to all API collections

### LINKED ISSUE #1674: Filtering for other object collections
I would like to filter networks, acls etc with the odata-like query filter in the API.

I took a look at the source code and it seems to me like the filtering is very generic, so it should be trivial to support it on other types as well. Is there anything I should be aware of before I take a stab at implementing this?

gwenya:
This PR aims to make odata-like filtering available for all collection endpoints (see #1674). 

So far I have only implemented it for `/1.0/networks`, could someone take a quick look and tell me if the way I'm doing this is alright, before I work on the other endpoints?

Closes #1674

### DIFF
--- a/cmd/incusd/api_project.go
+++ b/cmd/incusd/api_project.go
@@ -15,6 +15,7 @@ import (
 	"github.com/gorilla/mux"
 
 	incus "github.com/lxc/incus/v6/client"
+	"github.com/lxc/incus/v6/internal/filter"
 	"github.com/lxc/incus/v6/internal/jmap"
 	"github.com/lxc/incus/v6/internal/server/auth"
 	"github.com/lxc/incus/v6/internal/server/db"
@@ -73,6 +74,12 @@ var projectAccessCmd = APIEndpoint{
 //  ---
 //  produces:
 //    - application/json
+//  parameters:
+//    - in: query
+//      name: filter
+//      description: Collection filter
+//      type: string
+//      example: default
 //  responses:
 //    "200":
 //      description: API endpoints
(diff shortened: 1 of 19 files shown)


stgraber:
You're missing the Signed-off-by line and the extra import needs to be moved down with the other internal imports.

stgraber:
@gwenya is that the whole set?

If so, I can take a look at the remaining test failures and get this merged.

stgraber:
I've fixed the issues reported by `make static-analysis` and did a quick rebase with minor re-titling of the commits, added the output of `make update-api` and added an API extension so we can detect this change.

Assuming this passes all tests and that there isn't anything missing, we should be good to go with this.

stgraber:
Looking good, just need to retry that random failure

gwenya:
@stgraber it is the whole set unless I missed something, but I am not yet actually certain that it works :D

There aren't any tests for it and I'm not sure how or where to write any, and I haven't gotten around yet to manually testing all of them. 

There's also something weird with the formatting of the swagger specs, something (probably my IDE) kept indenting them with mixed tabs and spaces when I committed, and I only noticed afterwards and didn't yet get around to fix that.

I would also like to add the filtering support to the client like we have it for the existing filter-capable endpoints, but maybe that should be a separate PR.

stgraber:
> There aren't any tests for it and I'm not sure how or where to write any, and I haven't gotten around yet to manually testing all of them.

I'll take a quick look at that. We should be able to do a few basic tests with `incus query` until the various CLI commands get updated to use the server-side filtering logic.

> There's also something weird with the formatting of the swagger specs, something (probably my IDE) kept indenting them with mixed tabs and spaces when I committed, and I only noticed afterwards and didn't yet get around to fix that.

Yeah, it's a bit of a mess and definitely inconsistent. I've been tempted a few times to try and sort that mess out but always ran out of time ;)

> I would also like to add the filtering support to the client like we have it for the existing filter-capable endpoints, but maybe that should be a separate PR.

Incus 6.10 is getting tagged later today so I think it'd make sense to merge the API side today (after I do some manual tests and add a few automatic ones), then 6.11 can start making use of it from the CLI.

That basically allows for API users to start kicking the tires on this without the new logic being used for everyone, allowing for a slower roll out.

stgraber:
I edited all the swagger chunks in those files to remove the tabs. That took some effort as gofmt kept wanting to put them back in.

stgraber:
I still need to fix a crash in the buckets code and validate certificates, profiles and projects, but otherwise this is starting to behave.

I had to fix a few cases where a pointer was being passed to the filter which then leads to a crash when processing the request, that's been the most common issue so far.

stgraber (approved):

gwenya:
thank you, that is amazing!
Sorry for letting you do half the work, I got pretty busy the past couple days and couldn't give it the attention I wanted to :/

stgraber:
There really wasn't much left.

Most of the time was spent fighting `gofmt` trying to mess with the swagger text and then running into a limitation of the filters on structs with multiple in-line yaml statements. The rest was pretty smooth sailing, just testing everything and occasionally having to fix the exact struct being passed to the filter, but that just took a few minutes to work out.

PR #1692, opened 2025-02-24
Rework virtiofsd uid/gid map handling

stgraber:
virtiofsd has built-in uid/gid mapping these days, so use that rather than rely on our own user namespace. This fixes issues with the most recent virtiofsd which now fails to drop some privileges due to our user namespace wrapper.


internal/server/device/device_utils_disk.go
```diff
@@ -344,15 +345,43 @@ func DiskVMVirtiofsdStart(execPath string, inst instance.Instance, socketPath st
...
+
+			if entry.IsGID {
+				args = append(args, fmt.Sprintf("--translate-gid=map:%d:%d:%d", entry.NSID, entry.HostID, entry.MapRange))
+
+				args = append(args, fmt.Sprintf("--translate-gid=forbid-guest:%d:%d", lastGID, entry.NSID-1))
+				lastGID = entry.NSID
+			}
+		}
+
+		if lastUID < 4294967295 {
```
hallyn, line 373: Hm, but if there was only a uid, and no gid, then
lastGid will still be its default value and therefore
pass the < 4294967295 test, and you'll add an entry
for that.  Is that what you wanted?

stgraber: If no gid is passed (very bad idea), we should ban all GIDs, so I'd expect a 0-4294967295 reject range to be put in place.

stgraber: But that actually brings in a good point that I should initialize at -1 as with current logic, we'd end up only forbidding 1-4294967295, allowing the most dangerous one...

stgraber: For extra fun: https://gitlab.com/virtio-fs/virtiofsd/-/issues/198

hallyn (approved):

PR #1695, opened 2025-02-25
generate-database: Use deferred func to map errors & make generated code self-sufficient

breml:
Extend generate-database such that mapping of errors can be done outside of the generated code.

Embed all the boilerplate and make the generated code self-sufficient.

Remove the dependency of the generated code on package github.com/lxc/incus/v6/internal/server/db/query


internal/server/db/cluster/certificate_projects.mapper.go
```diff
@@ -10,11 +10,8 @@ import (
 	"fmt"
```
stgraber, line 10: Can you split the re-generation of `internal/server/db` to a separate commit?

internal/server/db/cluster/mapper_errors.go
```diff
@@ -0,0 +1,24 @@
+package cluster
```
stgraber, line 1: And the addition of the boilerplate and mapper also to a separate commit

internal/server/db/cluster/stmt.go
```diff
@@ -2,7 +2,7 @@
 
 package cluster
 
-//go:generate generate-database db mapper generate
+//go:generate generate-database db mapper generate -b mapper_boilerplate.go
```
stgraber: And that to its own commit too

breml:
@stgraber Updated the commits as requested.

stgraber (approved):

PR #1698, opened 2025-02-25
incusd/network/bridge: Fix deletion of tunnels and dummy devices

montag451:
In a previous PR (https://github.com/lxc/incus/pull/1627) I forgot to update one path in the code.

### DIFF
--- a/internal/server/network/driver_bridge.go
+++ b/internal/server/network/driver_bridge.go
@@ -665,27 +665,9 @@ func (n *bridge) setup(oldConfig map[string]string) error {
 		}
 	}
 
-	// Get a list of interfaces.
-	ifaces, err := net.Interfaces()
+	err = n.deleteChildren()
 	if err != nil {
-		return err
-	}
-
-	// Cleanup any existing tunnel and dummy devices.
-	for _, iface := range ifaces {
-		l, err := ip.LinkFromName(iface.Name)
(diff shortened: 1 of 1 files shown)


internal/server/network/driver_bridge.go
```diff
@@ -1509,6 +1509,29 @@ func (n *bridge) Stop() error {
...
+	}
+
+	// Cleanup any existing tunnel and dummy devices.
+	for _, iface := range ifaces {
+		l, err := ip.LinkFromName(iface.Name)
+		if err != nil {
+			return err
+		}
+
+		if l.Master != n.name || l.Kind != "vxlan" && l.Kind != "gretap" && l.Kind != "dummy" {
```
stgraber, line 1525: Would be easier to read if using `slices.Contains` for the different kinds I think.

stgraber:
So the tests failing bring up a good point.

If the user passes a gretap, vxlan or dummy device through `bridge.external_interfaces`, we definitely don't want to delete those.

It would probably be a good idea to go over the interfaces in `bridge.external_interfaces` and remove those from the bridge before nuking everything else.

internal/server/network/driver_bridge.go
montag451: Done!

stgraber (approved):

PR #1705, opened 2025-02-28
allow unprivileged users in application containers to bind to all ports

### LINKED ISSUE #1704: Allow non-root users in OCI to listen to low port numbers
In docker it is possible since some time ago for non-root users to listen to service ports, which is generally considered to be safe and makes sense for application containers (see https://github.com/moby/moby/pull/41030).

Would it make sense for Incus to also set `net.ipv4.ip_unprivileged_port_start` to `0` in OCI containers by default?

gwenya:
closes #1704


gwenya:
Oops, sorry forgot. I’ll do that tomorrow first thing in the morning.

PR #1706, opened 2025-02-28
Allow ICMP and low ports for unprivileged users in OCI containers

gwenya:
Docker uses the range `0 2147483647` for the ping group IDs, but that fails if not that many groups are available, and the default incus configuration only allocates 1000000000 IDs, so I opted for a more conservative `0 65536`.

I suppose ideally we would check what ID range is available and use that, but I couldn't figure out how.

### DIFF
--- a/internal/server/instance/drivers/driver_lxc.go
+++ b/internal/server/instance/drivers/driver_lxc.go
@@ -2306,6 +2306,28 @@ func (d *lxc) startCommon() (string, []func() error, error) {
 			volatileSet["volatile.container.oci"] = "true"
 		}
 
+		// Allow unprivileged users to use ping.
+		maxGid := int64(4294967295)
+
+		if !d.IsPrivileged() {
+			maxGid = 0
+			idMap, err := d.CurrentIdmap()
+			if err != nil {
+				return "", nil, err
+			}
+
+			for _, entry := range idMap.Entries {
+				if entry.NSID+entry.MapRange-1 > maxGid {
+					maxGid = entry.NSID + entry.MapRange - 1
+				}
+			}
+		}
+
+		err = lxcSetConfigItem(cc, "lxc.sysctl.net.ipv4.ping_group_range", fmt.Sprintf("0 %d", maxGid))
+		if err != nil {
+			return "", nil, err
+		}
+
 		// Configure the entry point.
 		if len(config.Process.Args) > 0 && slices.Contains([]string{"/init", "/sbin/init", "/s6-init"}, config.Process.Args[0]) {
 			// For regular init systems, call them directly as PID1.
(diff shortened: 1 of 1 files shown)


internal/server/instance/drivers/driver_lxc.go
```diff
@@ -2336,6 +2336,11 @@ func (d *lxc) startCommon() (string, []func() error, error) {
 			return "", nil, err
 		}
 
+		err = lxcSetConfigItem(cc, "lxc.sysctl.net.ipv4.ping_group_range", "0 65536")
```
stgraber, line 2339: Why limit it to 65536?
The max value these days is 4294967295.

stgraber: Oh, you commented about that already.

stgraber: Okay, so we'll need a bit of trickery to get the maximum gid from the idmap then.

stgraber: @gwenya Can you try something like this maybe?
```diff
diff --git a/internal/server/instance/drivers/driver_lxc.go b/internal/server/instance/drivers/driver_lxc.go
index 7bbc5e4610..462e4ce9a7 100644
--- a/internal/server/instance/drivers/driver_lxc.go
+++ b/internal/server/instance/drivers/driver_lxc.go
@@ -2321,6 +2321,27 @@ func (d *lxc) startCommon() (string, []func() error, error) {
                        }
                }
 
+               maxGid := int64(4294967295)
+               if !d.IsPrivileged() {
+                       maxGid = 0
+
+                       idmap, err := d.CurrentIdmap()
+                       if err != nil {
+                               return "", nil, err
+                       }
+
+                       for _, entry := range idmap.Entries {
+                               if entry.NSID + entry.MapRange - 1 > maxGid {
+                                       maxGid = entry.NSID + entry.MapRange - 1
+                               }
+                       }
+               }
+
+               err = lxcSetConfigItem(cc, "lxc.sysctl.net.ipv4.ping_group_range", fmt.Sprintf("0 %d", maxGid))
+               if err != nil {
+                       return "", nil, err
+               }
+
                err = lxcSetConfigItem(cc, "lxc.init.cwd", config.Process.Cwd)
                if err != nil {
                        return "", nil, err
```

gwenya: What would I have to do to get more than one idmap on a container? This code works with just one, but I'd like to test it with multiple as well.

stgraber: Using `raw.idmap` should let you pass in more entries

stgraber: Say something like `raw.idmap: both 9000000 9000000`  would create a uid and gid map entry for that uid/gid to be passed through to the container, which would then raise the max gid to 9000000

gwenya: perfect, that works

stgraber:
Moved the commits from both PRs into this one, re-titled the commits a little and added comments to the code.

stgraber (approved):

PR #1716, opened 2025-03-02
incusd/storage/ceph: Re-introduce keyring parsing

### LINKED ISSUE #1711: New ceph implementation required permissions unclear without admin user
With the 6.10 update, the ceph integration in our cluster broke (with the fix from today already applied):

When starting a VM, this error occurs:
`Failed setting up disk device "root": Failed to query ceph config for auth_service_required: Failed to run: ceph --cluster ceph config get client.incus auth_service_required: exit status 1 (2025-03-01T13:53:07.250+0100 7efc748886c0 -1 monclient(hunting): handle_auth_bad_method server allowed_methods [2] but i only support [2] [errno 1] RADOS permission error (error connecting to the cluster))`

This is a result of this call:

https://github.com/lxc/incus/blob/4905664364bea3cb57b08bc56b20879739a2471c/internal/server/storage/drivers/utils_ceph.go#L190-L200

If I understand it correctly, incus tries to directly obtain the keyring from ceph instead of using the provided keyring in `/etc/ceph`.
As the call is made without `--id` or `--name`, the call is tried with the ceph admin user by default.

However, this does not work, as `/etc/ceph` only contains the incus ceph user in our deployment.
We don't want incus to act as the admin user, as the ceph cluster is also used for other applications.

While we theoretically could set the user with `CEPH_ARGS='-n client.incus'` env var, this still would not work, as I am not really aware which permissions the ceph user needs in order to run the required commands.
So this still results in `Error EACCES: access denied` (User has not the permissions to access the config)

Incus ignoring the given username for the ceph pool seems to be problematic.
And I could not find documentation that states which permissions the ceph user needs.

With 6.9 it worked fine with 
- `mon profile rbd`
- `osd profile rbd pool=incus`

stgraber:
Closes #1711

### DIFF
--- a/internal/server/storage/drivers/utils_ceph.go
+++ b/internal/server/storage/drivers/utils_ceph.go
@@ -1,13 +1,16 @@
 package drivers
 
 import (
+	"bufio"
 	"encoding/json"
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/lxc/incus/v6/shared/api"
 	"github.com/lxc/incus/v6/shared/logger"
 	"github.com/lxc/incus/v6/shared/subprocess"
+	"github.com/lxc/incus/v6/shared/util"
 )
 
 // CephGetRBDImageName returns the RBD image name as it is used in ceph.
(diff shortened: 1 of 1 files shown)


internal/server/storage/drivers/utils_ceph.go
```diff
@@ -208,20 +211,119 @@ func CephKeyring(cluster string, client string) (string, error) {
...
+	for scan.Scan() {
+		line := scan.Text()
+		line = strings.TrimSpace(line)
+
+		if line == "" {
+			continue
+		}
+
+		if strings.HasPrefix(line, "key") {
+			fields := strings.SplitN(line, "=", 2)
```
hallyn, line 240: here and below, it seems like you should first drop anything after a # or ;. Else a line like

```
key = value # used to be key = other
```

will wrongly be skipped.

stgraber: I'm need to review the logic again, for now what's in there is a straight up revert to the logic we've had for the past 5+ years so probably best to go with that for the regression fix and then look at it again later

hallyn: Sounds good.

hallyn (approved):

PR #1726, opened 2025-03-05
Fixed incus edk2 path overwrite issue

nanjj:
Current INCUS_EDK2_PATH once set, no matter it's directory exists or not, it will overwrite the buildin EDK2 Path. The fix added two conditions to overwrite:
1. buildit EDK2 Path directory not exist, and
2. INCUS_EDK2_PATH directory exists.

And a function naming issue is fixed too(GetAchitectureFirmwarePairs -> GetArchitectureFirmwarePairs).

### DIFF
--- a/internal/server/instance/drivers/driver_qemu.go
+++ b/internal/server/instance/drivers/driver_qemu.go
@@ -2159,7 +2159,7 @@ func (d *qemu) setupNvram() error {
 	d.logger.Debug("Generating NVRAM")
 
 	// Cleanup existing variables.
-	for _, firmwarePair := range edk2.GetAchitectureFirmwarePairs(d.architecture) {
+	for _, firmwarePair := range edk2.GetArchitectureFirmwarePairs(d.architecture) {
 		err := os.Remove(filepath.Join(d.Path(), filepath.Base(firmwarePair.Vars)))
 		if err != nil && !errors.Is(err, fs.ErrNotExist) {
 			return err
--- a/internal/server/instance/drivers/driver_qemu.go
+++ b/internal/server/instance/drivers/driver_qemu.go
@@ -2159,7 +2159,12 @@ func (d *qemu) setupNvram() error {
 	d.logger.Debug("Generating NVRAM")
 
 	// Cleanup existing variables.
-	for _, firmwarePair := range edk2.GetArchitectureFirmwarePairs(d.architecture) {
+	firmwares, err := edk2.GetArchitectureFirmwarePairs(d.architecture)
+	if err != nil {
+		return err
+	}
+
+	for _, firmwarePair := range firmwares {
 		err := os.Remove(filepath.Join(d.Path(), filepath.Base(firmwarePair.Vars)))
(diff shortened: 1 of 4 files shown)


internal/server/instance/drivers/driver_qemu.go
```diff
@@ -2159,7 +2159,7 @@ func (d *qemu) setupNvram() error {
 	d.logger.Debug("Generating NVRAM")
 
 	// Cleanup existing variables.
-	for _, firmwarePair := range edk2.GetAchitectureFirmwarePairs(d.architecture) {
+	for _, firmwarePair := range edk2.GetArchitectureFirmwarePairs(d.architecture) {
```
stgraber, line 2162: That's not a typo, it's a function which returns pairs of architecture and firmware.

internal/server/instance/drivers/edk2/driver_edk2.go
```diff
@@ -207,5 +212,9 @@ func GetArchitectureFirmwarePairsForUsage(hostArch int, usage FirmwareUsage) []F
 
 // GetenvEdk2Path returns the environment variable for overriding the path to use for EDK2 installations.
 func GetenvEdk2Path() string {
-	return os.Getenv("INCUS_EDK2_PATH")
+	if iep := os.Getenv("INCUS_EDK2_PATH"); iep != "" && util.PathExists(iep) {
```
stgraber, line 215: Silently ignoring an invalid path seems like a bad idea

stgraber: Would probably be best to turn this into a `(string, error)` and having it fail if the path is specified but doesn't exist.

internal/server/instance/drivers/edk2/driver_edk2.go
```diff
@@ -178,16 +178,21 @@ func GetAchitectureFirmwarePairs(hostArch int) []FirmwarePair {
...
-					searchPath = installation.Path
-				}
+			searchPath := installation.Path
+			if !util.PathExists(searchPath) {
+				searchPath = incusEdk2Path
+			}
+
+			if searchPath == "" {
+				continue
+			}
```
stgraber, line 193: Should add a comment on top of that code to explain what's going on here, basically that since we have a user-specified firmware path, we'll check in there too.

stgraber (changes requested):
Just a few tweaks needed to get this through static analysis and line up with our general code style and pattern.

Note that the new test file is triggering a bunch of static analysis failure (see Github Actions), so those will need to be addressed too.

We also do not use the define-and-check syntax within the Incus codebase, so `if foo := bar(); foo == "baz" {` should be written as two lines, one `foo := bar()` and one `if foo == "baz" {`

internal/server/instance/drivers/edk2/driver_edk2.go
nanjj: done

stgraber (approved):

PR #1728, opened 2025-03-05
Address sets for nftables and OVN

### LINKED ISSUE #1450: Implement network address sets
One current missing piece of our networking puzzle is support for address sets.
That is, a named set of address and/or subnets which can be used as source/destination selector within ACLs.

There is a native OVN object for this feature, so it would make sense to initially implement this for OVN networking.

Address sets should be scoped the same way as ACLs as they'll be very closely linked. This means address sets should be part of a project when `features.networks=true`

The expected changes should roughly be:
 - Allocate API extension
 - New database table (schema change)
 - New database functions (using the generator)
 - Add REST API structs for network address sets
 - Add OVN NB bindings for managing address sets
 - Add REST API to manipulate address sets
 - Add integration with OVN ACLs
 - Implement CLI command
 - Refresh translations
 - Add tests

irhndt:
Closes #1450
This pull request add address sets support for incus, more precisely it allows the use of named sets for nftables and address sets in OVN.
This is my first 'real' contribution and first go project. You may want to check important logic parts in nftables and ovn drivers. 
 
Notes:

- I wasn't able to get patch test work in github action but worked on my local env
- I did not add ovn tests in main shell test suite because I did not see any test related to ovn in scripts (perhaps did not look well) but I could add them quickly
- I made a lot of 'bad' commits because I wasn't able to run test in my local env and did a lot of back and forth with github actions 
- Support for OpenFGA has been added, filtering too
- Surely missed or forgot something and would be happy to enhance the feature !


irhndt:
Hi, is there anything I can do to help process the PR?

stgraber:
Nope, it's next on my list to review, just taking a bit as it's a sizable PR :)

stgraber:
I spent a couple of hours on cleaning up the branch, mostly rebasing everything on current main and re-slicing the code in our usual commit chunks. I also did a few minor code style tweaks here and there, but nothing major.

I did notice a few major changes that I'll want to make to this branch, but I'll need to do that later, maybe tomorrow evening, maybe Thursday. It's mostly around the database, changing the name of the tables to line up with our usual pattern, using the DB generator instead of manual SQL functions and a few other similar tweaks.

Anyway, I also need to do an actual review of the logic and manual testing of all of this.
Today was really a superficial scan through things as I put all the changes into commits.

stgraber:
Don't worry about the test failures, it's expected at this stage :)

irhndt:
Good to know, looking forward to learn from your modifications !

stgraber:
Haven't forgotten about this but have had to deal with some priority bugs.

stgraber:
Been making some progress on this one, but working locally so haven't pushed anything yet.

I've renamed all the ExternalIDs stuff to Config for consistency, renamed the tables to our usual pattern too and put the DB generator in place to generate all the DB access code.

I'm now replacing all the DB function calls with the equivalent for the generated code.

After that's done, I'll want to make sure we can get the tests behaving again before I actually review the content of the commits.

irhndt:
Yes, at first glance I used ExternalIDs while reading OVN man pages however after I have seen that ACL for example had the same field under config... Forgot to change to comply but I should have done it. 

Concerning DB function I want to see how you do that as I don't get exactly what is the 'equivalent'.

When I took a look at failed tests last week I saw that we have failure on the PATCH request test. I was able to get it to work on my computer but once I ran it in github action it wouldn't, so as everything else was ok in github actions I commented it out. So we may want to check other tests are behaving before tackling the patch issue.

Thanks for the heads up!

stgraber:
Just pushing an intermediate state. I've moved everything to the DB generator, code builds, static analysis looks clean, so I want to get an idea of what I broke before I clean things up and review the rest.

irhndt:
HI, I took a look on the error:

`Error: Failed to fetch from "network_address_set_config" table: Failed to fetch from "network_address_set_config" table: no such table: networks_addresss_sets_config`

I think the issue lies in the GetConfig function in `internal/server/db/cluster/config.mapper.go`, I will take a look later if you dont do it in the meantime. I am off for most of the week-end.

stgraber:
> HI, I took a look at the error:
> 
> `Error: Failed to fetch from "network_address_set_config" table: Failed to fetch from "network_address_set_config" table: no such table: networks_addresss_sets_config`
> 
> I think the issue lies in the GetConfig function in `internal/server/db/cluster/config.mapper.go`, I will take a look later if you dont do it in the meantime. I am off for most of the week-end.

Yeah, I know, but this is generated code. You can't change that file as it will get overwritten.

I reached out to @masnax about the database generator and what we need to do to make it behave given the name of the tables involved.

stgraber:
Did a bunch of cleanup and minor fixes here and there, added gendoc for the config keys and did a few more tweaks. I've now manually reviewed all commits except for the two most complex ones:

 - incusd/firewall: Add address set support
 - incusd/network/address-set: Add new API

I'm hoping to get to review those later tomorrow and then can get this one merged and start reviewing the #1621 PR.

stgraber:
@irhndt for #1802, you can probably look at this PR now for a more relevant example of moving from handwritten functions to using the DB generator though there will likely be a few more tweaks as I review the main API commit.

irhndt:
> @irhndt for #1802, you can probably look at this PR now for a more relevant example of moving from handwritten functions to using the DB generator though there will likely be a few more tweaks as I review the main API commit.

Yes thank you this will be a good example.

irhndt:
Nftables firewall was a bit tricky hope it goes well.

stgraber:
Quick update on this one. I've gone through the API commit and that one is now good to go.
So that just leaves the firewall commit left to review.

Then I want to do some manual testing to validate some of the logic that's a bit harder to follow and that I'd just rather see work for myself :)

I mostly want to see it:
 - On OVN, creating all the expected DB records and cleaning everything up when removed
 - On nftables, creating all the expected records and cleaning everything up when removed
 - Test situation where address set is used by multiple ACLs and across multiple instances and networks. Then try to delete some (but not all) of the ACLs and networks, confirm that the address set is still in place, then delete the rest and confirm everything is clean.
 - Confirm that having the same address set name and acl name across multiple projects doesn't lead to conflicts (not too worried about OVN but I want to make sure for nft)

irhndt:
> Quick update on this one. I've gone through the API commit and that one is now good to go. So that just leaves the firewall commit left to review.
> 
> Then I want to do some manual testing to validate some of the logic that's a bit harder to follow and that I'd just rather see work for myself :)
> 
> I mostly want to see it:
> 
> * On OVN, creating all the expected DB records and cleaning everything up when removed
> * On nftables, creating all the expected records and cleaning everything up when removed
> * Test situation where address set is used by multiple ACLs and across multiple instances and networks. Then try to delete some (but not all) of the ACLs and networks, confirm that the address set is still in place, then delete the rest and confirm everything is clean.
> * Confirm that having the same address set name and acl name across multiple projects doesn't lead to conflicts (not too worried about OVN but I want to make sure for nft)

I think the first two points should be ok.
For the third I think too, but I don't remember how extensively I have tested this.

And the last I think it won't work, because I designed the address set name thinking that likely it would be unique across projects. 

Nftables driver will create as_ipv4 and as_ipv6 when requested without taking in account a specific project. 
Everyone in the incus nft namespace can use and see the address set and it is a design of nftables.

Named sets defined are available for all the nft incus namespace so currently we may have mixes between different projects / networks. 

Good to know each time network is reloaded (many events can bring this) if an address set we want to set is already defined it is flushed before being re-applied.

Not sure if it will be easy to bound it to a project or a network maybe we want to prefix the address set with the network name or something.

One work-around I see would be to prefix nft named sets with incus `networkName`, in this case when we deleted a network we only delete named sets that use `networkName` as a prefix.


Tell me if you want me to work on something later to adjust if needed.

stgraber:
> > Quick update on this one. I've gone through the API commit and that one is now good to go. So that just leaves the firewall commit left to review.
> > Then I want to do some manual testing to validate some of the logic that's a bit harder to follow and that I'd just rather see work for myself :)
> > I mostly want to see it:
> > 
> > * On OVN, creating all the expected DB records and cleaning everything up when removed
> > * On nftables, creating all the expected records and cleaning everything up when removed
> > * Test situation where address set is used by multiple ACLs and across multiple instances and networks. Then try to delete some (but not all) of the ACLs and networks, confirm that the address set is still in place, then delete the rest and confirm everything is clean.
> > * Confirm that having the same address set name and acl name across multiple projects doesn't lead to conflicts (not too worried about OVN but I want to make sure for nft)
> 
> I think the first two points should be ok. For the third I think too, but I don't remember how extensively I have tested this.
> 
> And the last I think it won't work, because I designed the address set name thinking that likely it would be unique across projects.
> 
> Nftables driver will create as_ipv4 and as_ipv6 when requested without taking in account a specific project. Everyone in the incus nft namespace can use and see the address set and it is a design of nftables.
> 
> Named sets defined are available for all the nft incus namespace so currently we may have mixes between different projects / networks.
> 
> Good to know each time network is reloaded (many events can bring this) if an address set we want to set is already defined it is flushed before being re-applied.
> 
> Not sure if it will be easy to bound it to a project or a network maybe we want to prefix the address set with the network name or something.
> 
> One work-around I see would be to prefix nft named sets with incus `networkName`, in this case when we deleted a network we only delete named sets that use `networkName` as a prefix.
> 
> Tell me if you want me to work on something later to adjust if needed.


So for the nft case, I think that's fine. nft is only possible on bridges, bridges are only allowed in the default project.

The OVN case is where we end up with sets inside of projects and so being able to have the same set name multiple times as well as the same network name multiple times.

That's why we generally keep track of things based on database id for networks, acls, ... when feeding stuff into OVN.

Anyway, we may be fine here but I'll check during testing.

irhndt:
> > > Quick update on this one. I've gone through the API commit and that one is now good to go. So that just leaves the firewall commit left to review.
> > > Then I want to do some manual testing to validate some of the logic that's a bit harder to follow and that I'd just rather see work for myself :)
> > > I mostly want to see it:
> > > 
> > > * On OVN, creating all the expected DB records and cleaning everything up when removed
> > > * On nftables, creating all the expected records and cleaning everything up when removed
> > > * Test situation where address set is used by multiple ACLs and across multiple instances and networks. Then try to delete some (but not all) of the ACLs and networks, confirm that the address set is still in place, then delete the rest and confirm everything is clean.
> > > * Confirm that having the same address set name and acl name across multiple projects doesn't lead to conflicts (not too worried about OVN but I want to make sure for nft)
> > 
> > 
> > I think the first two points should be ok. For the third I think too, but I don't remember how extensively I have tested this.
> > And the last I think it won't work, because I designed the address set name thinking that likely it would be unique across projects.
> > Nftables driver will create as_ipv4 and as_ipv6 when requested without taking in account a specific project. Everyone in the incus nft namespace can use and see the address set and it is a design of nftables.
> > Named sets defined are available for all the nft incus namespace so currently we may have mixes between different projects / networks.
> > Good to know each time network is reloaded (many events can bring this) if an address set we want to set is already defined it is flushed before being re-applied.
> > Not sure if it will be easy to bound it to a project or a network maybe we want to prefix the address set with the network name or something.
> > One work-around I see would be to prefix nft named sets with incus `networkName`, in this case when we deleted a network we only delete named sets that use `networkName` as a prefix.
> > Tell me if you want me to work on something later to adjust if needed.
> 
> So for the nft case, I think that's fine. nft is only possible on bridges, bridges are only allowed in the default project.
> 
> The OVN case is where we end up with sets inside of projects and so being able to have the same set name multiple times as well as the same network name multiple times.
> 
> That's why we generally keep track of things based on database id for networks, acls, ... when feeding stuff into OVN.
> 
> Anyway, we may be fine here but I'll check during testing.

Okay, if needed I think we can simply update calls to : 
`client.CreateAddressSet
client.DeleteAddressSet
client.UpdateAddressSetAdd
client.UpdateAddressSetRemove` 
Example:
`client.DeleteAddressSet(context.TODO(), ovn.OVNAddressSet(setName))`

We can swap `setName` to something like `projectName+setName`, and it should be transparent for OVN.  Correct me if I am wrong, this could be done now so we don't step in name mangling errors in the future.

stgraber:
Looked through the firewall commit, seems fine. I'll look at manual tests next before merging this.

stgraber:
Right, so the OVN naming needs changes indeed.

We currently have `foo_ip6` and `foo_ip4` when all other OVN things are typically named like `incus_OBJECT$ID_ip4`.

So my `foo_ip6` should become `incus_set1_ip4` given database ID 1 for the set.

stgraber:
@irhndt can you push an extra commit to change the OVN address set names to follow the `incus-set<ID>_ip4` and `incus-set<ID>_ip6` where the `ID` is the database ID for the record.

That would line up with what we do for networks, load-balancers, ... basically anything we put in OVN where we need to provide a name.

stgraber:
On OVN, I also just did:
 - incus network address-set create foo
 - incus network acl create bar
 - incus network acl edit bar (and added an ingress rule using `$foo` as source)
 - incus config device set test eth0 security.acls=bar

This cause the address set to appear in the OVN database.

But then none of those actions caused it to go away:
 - incus config device set test eth0 security.acls=
 - incus network acl delete bar
 - incus network address-set delete foo

I would have expected it to get away as soon as its ACL ref count hit zero which would have been as soon as the `set security.acls=` as that caused the ACL itself to go away since nothing was referencing it.

stgraber:
I'm basically playing with address sets and ACLs and doing `ovn-nbctl list acl` and `ovn-nbctl list address_set` and doing diffs of those before/after various actions to confirm that only what I'm expected is appearing in there and that it goes away when performing the reverse action.

stgraber:
Thanks! 

That's fine, that's what reviews are for and also why I do reviews in that order.

I usually start by making sure I'm happy with the API, CLI and doc as that's user facing and not something we can easily fix after the fact. 

Then once I'm happy with that, I do a quick check of all of the remaining code, partly for style, partly for risky spots like security checks and the like.

Then once everything checks out, I finally deploy it on both a test standalone and a test cluster and try to use it in most ways I can see a user using it. 

That last step is usually what catches most of the issues as it's hard to actually imagine how everything fits together during code review. But that's useful context for the actual manual test as now I know mostly where to look at for whatever issue happened :)

So anyway, we're getting in very good shape now, hopefully just that one more fix and then manual tests will all clear and we'll be good to go.

irhndt:
Okay I have made something up that worked for me, however I messed up commits when pulling before push, some drivers_nftables got pushed and golangcgi.yml. Trying to fix this as I did not wanted to do it.

stgraber:
Let me know when you're happy with things on your end, then I'll look at the changes, then rebase and go back to manually testing.

irhndt:
Okay I have a simple issue I have not figured out hwo to solve it for now.
In acl ovn when we parse rules and stumble accross an address set we need to do a conversion to ovn statements:

`ip4.src == incus-set<ID>_ip4 ...`

However previously we used address set name that allowed us to use subjectCriterion:

`ip4.src == <subjectCriterion>_ip4 ...`
 
 
 That avoided us to query the database to get the ID of relevant address set, however now I don't have neither state object to query db nor projectName.
 
 The concerned function is `ovnRuleSubjectToOVNACLMatch`, @stgraber  do you have an idea to workaround ?
 
 Also for static checks and commit signing I do not get why it won't pass anymore ? 
 
 After that I'll do a bit more of testing and we should be ok though.

stgraber:
I'll take a look at the issue you mentioned shortly.

stgraber:
@irhndt I've now added an extra commit (first one in the branch) which adds plumbing for the state struct through to the OVN ACL functions. Can you check if that helps you out?

irhndt:
> @irhndt I've now added an extra commit (first one in the branch) which adds plumbing for the state struct through to the OVN ACL functions. Can you check if that helps you out?

I made changes, I'm still doing some manual testing but logic should be ok.

stgraber:
@irhndt I'm not seeing any new changes in the branch since I last pushed it.
Let me know when you've pushed the change and I'll do another round of cleanup and rebase then get back to reviewing and testing.

irhndt:
Sorry I made a mistake I thought I had everything pushed and I got some problems with my VM I need to reset my env. Still I made a push that should be good (haven't did a static-analysis) I will rebase with formatting when I reset my env.

stgraber:
```
root@d13:~# incus list
+------+---------+----------------------+------------------------------------------------+-----------+-----------+
| NAME |  STATE  |         IPV4         |                      IPV6                      |   TYPE    | SNAPSHOTS |
+------+---------+----------------------+------------------------------------------------+-----------+-----------+
| d13  | RUNNING | 10.69.232.120 (eth0) | fd42:b0e2:ef9f:b81c:1266:6aff:fe91:faac (eth0) | CONTAINER | 0         |
+------+---------+----------------------+------------------------------------------------+-----------+-----------+
root@d13:~# incus network acl create foo
Network ACL foo created
root@d13:~# incus config device set d13 eth0 security.acls=foo
root@d13:~# incus network address-set create bar
Network address set bar created
root@d13:~# incus network acl rule add foo egress action=allow state=enabled destination='$bar'
Error: Failed updating bridge NIC ACL: Failed to trigger device update for device "eth0" of instance "d13" in project "default": Failed adding bridge filter rules for instance device "d13.eth0" (bridge): Failed apply nftables config: Failed to run: nft -f -: exit status 1 (/dev/stdin:10:35-43: Error: No such file or directory
	iifname "vethe1dce946" ip daddr  @bar_ipv4 accept
	                                ^^^^^^^^^
/dev/stdin:11:36-44: Error: No such file or directory
	iifname "vethe1dce946" ip6 daddr  @bar_ipv6 accept
	                                 ^^^^^^^^^
/dev/stdin:25:35-43: Error: No such file or directory
	iifname "vethe1dce946" ip daddr  @bar_ipv4 accept
	                                ^^^^^^^^^
/dev/stdin:26:36-44: Error: No such file or directory
	iifname "vethe1dce946" ip6 daddr  @bar_ipv6 accept
	                                 ^^^^^^^^^)
root@d13:~# 
```

irhndt:
I'll check that as soon as I can with the same scenario.

irhndt:
Okay I crafted something up, but I still feel that a cleanup of the code after reviews would do great. 
I added support for ACLs attached to an instance using nftables as backend. Is there other kind of techniques to attach an ACLs to an instance / network that I should look in it see if the address sets comply?

Sorry again to have forgotten all that part.

stgraber:
> Okay I crafted something up, but I still feel that a cleanup of the code after reviews would do great. I added support for ACLs attached to an instance using nftables as backend. Is there other kind of techniques to attach an ACLs to an instance / network that I should look in it see if the address sets comply?
> 
> Sorry again to have forgotten all that part.

Nope, that should be it, ACLs are either on a NIC or on a network, or a combination of the two.

stgraber:
I merged the changes and will do some testing in a little bit.

stgraber:
Things seem to behave on local test. I tried with a few containers, adding/removing stuff and it seems to be behaving.

stgraber:
OVN testing is next

stgraber:
@irhndt so I found a problem on the OVN side with my first test :(

What I did is:
 - incus network acl create test
 - incus config device set a1 eth0 security.acls=test
 - incus network address-set create foo
 - incus network address-set add foo 1.1.1.1
 - incus network acl rule add test egress action=reject state=enabled destination='$foo'

I can see the expected rule in the ACL table:
```
_uuid               : f45b0526-9f38-42ba-a7eb-1608fc6340af
action              : reject
direction           : to-lport
external_ids        : {incus_port_group=incus_acl4}
label               : 0
log                 : false
match               : "(inport == @incus_acl4) && (ip6.dst == $incus_set2_ip6 || ip4.dst == $incus_set2_ip4)"
meter               : []
name                : []
options             : {}
priority            : 400
severity            : []
tier                : 0
```

But `incus_set2_ip4` and `incus_set2_ip6` don't exist:
```
root@server01:~# ovn-nbctl list address_set | grep incus_set
root@server01:~# 
```

So that's leading to an invalid ACL.

irhndt:
> * incus network acl rule add test egress action=reject state=enabled destination='$foo'

The problem is that the trigger for reloading address-set for OVN are at network / interface load, in fact trying to solve it I came into a circular dependency issue as I call some of ACL package func in address set while I wanted to call EnsureAddressSets during Update method from ACL.

 I only tested -why be so dumb- loading/reloading in the following order:

incus network acl create test
incus network address-set create foo
incus network address-set add foo 1.1.1.1
incus network acl rule add test egress action=reject state=enabled destination='$foo'
incus config device set a1 eth0 security.acls=test // Always ending by security.acls config or address set update


I think with all those things I forgot I might want to rebuild it properly and take time to do it well now that I get a grasp on this will be used. As I had a small experience in managing incus I forgot to test all of the above cases, I should have thinked it twice. I might have some time to finish it properly this weekend but I don't want rush it now (sadly missing 6.11 ) unless you think it might be done until ?

stgraber:
I moved the issue to 6.12 now. That's fine. Makes sense to spend some time to think through the different cases for both bridged and OVN when it comes to dealing with references and when to add/update/remove the set.


We ideally want the address sets to be handled as a dependency of an ACL, so when our logic would kick in the creation of ACL rules in the backend (nft or OVN), then the set would first get loaded if it's not there yet. When ACL rules get removed, we'd then need to check whether the address set is still in use by any active ACL rule and if not, it would get removed.

Then address set updates basically just check if the address set is loaded into the backend and if it is, updates it there. This one can be done in one of two ways:
 - The correct way would be to track down all the ACLs that have rules that make use of that address set, then check on what driver they're used and refresh the driver
 - The easier shotgun approach, of just calling the update logic on both nft and OVN regardless and ignore failures related to the set not existing

irhndt:
@stgraber 
First of all thank you for your advice. 
I reviewed "I think" all cases of address set update but I'm sure you'll find something 😆. 
I solved circular dependency by copying some of ACLs usage function over address set package. This is because to compute address set network usage I do the following:

- Find all ACLs using the address set
- For each of those ACLs compute their networkUsage
- Then we have address network usage

So I needed the ACL NetworkUsage function.

I added calls to load / unload address sets in nft or OVN backend inside the following files:

`internal/server/network/ACL/dirver_common.go` in Update method

`internal/server/network/address-set/dirver_common.go`

`internal/server/network/driver_ovn.go`

`internal/server/network/driver_bridge.go`

`internal/server/device/nic_bridge.go`

`internal/server/device/nic_ovn.go`

I am hopping it will be okay. In any case I will review at least one last time to be sure I didn't do any wrong.

I used the following manual test script:

https://github.com/irhndt/incus/blob/test-devs/tests_as.sh
 
Hope I did not forget anything. Have a nice week.

stgraber:
The OVN behavior seems fine now. Going to quickly recheck the nftables one.

stgraber:
nftables implementation seems to still work as expected here, so I think we're good to go!

stgraber:
Thanks for all the work @irhndt !

PR #1731, opened 2025-03-06
generate-database: Add support for marshal to JSON

breml:


stgraber:
Just a couple of things:
 - To me testing IsTruthy will always require me to go look up the function to see what we mean whereas !IsFalseOrEmpty is pretty clearly "is not false or empty", so I think keeping the long form will be overall easier to follow. 
 - Incus code style has error messages starting with a capital letter. Not something that's always been consistently done, but we at least fix the ones we notice and try not to introduce more error messages following a different style.

breml:
> * To me testing IsTruthy will always require me to go look up the function to see what we mean whereas !IsFalseOrEmpty is pretty clearly "is not false or empty", so I think keeping the long form will be overall easier to follow.

Sure, I can change this, but there is a [lint-script](https://github.com/lxc/incus/blob/main/test/lint/negated-is-bool.sh) prohibiting `!IsFalseOrEmpty`.

stgraber:
> > * To me testing IsTruthy will always require me to go look up the function to see what we mean whereas !IsFalseOrEmpty is pretty clearly "is not false or empty", so I think keeping the long form will be overall easier to follow.
> 
> Sure, I can change this, but there is a [lint-script](https://github.com/lxc/incus/blob/main/test/lint/negated-is-bool.sh) prohibiting `!IsFalseOrEmpty`.
> 
> So I see the following options:
> 
> * Rename `IsTruthy` to `IsNotFalseOrEmpty`
> * Change the linter to allow an exception for `!IsFalseOrEmpty`
> * Keep `IsTruthy`
> 
> What do you prefer @stgraber?
> 

Ah, I forgot about that check. 

Let's do IsNorFalseOrEmpty then.

PR #1732, opened 2025-03-07
Do not download instance types if cache loadable

nanjj:
and environment INCUS_SKIP_INSTANCE_TYPES set to true, because most enterprise
environment has no internet access, so they can put instance types in cache and
set `INCUS_SKIP_INSTANCE_TYPES = true`.


stgraber:
I'd probably name it `INCUS_SKIP_INSTANCE_TYPES` instead so it can be `true` when enabled.
You'd also want to add the new variable to doc/environment.md so people can actually find this

stgraber (approved):

PR #1734, opened 2025-03-07
Allow announcing extra routes through DHCPv4

gwenya:
This PR adds a new option `ipv4.dhcp.static-routes` for both bridge and ovn networks. The value of the option is directly passed through to dnsmasq and ovsdb respectively, to be used as DHCP option 121.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,6 +38,8 @@ jobs:
       - id: ShellCheck
         name: Differential ShellCheck
         uses: redhat-plumbers-in-action/differential-shellcheck@v5
+        env:
+          SHELLCHECK_OPTS: --shell sh
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
         if: github.event_name == 'pull_request' && matrix.go == 'stable'
--- a/cmd/incusd/main_forknet.go
+++ b/cmd/incusd/main_forknet.go
@@ -391,6 +391,21 @@ func (c *cmdForknet) RunDHCP(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
+	for _, staticRoute := range lease.Offer.ClasslessStaticRoute() {
+		route := &ip.Route{
+			DevName: iface,
+			Route:   staticRoute.Dest.String(),
+			Via:     staticRoute.Router.String(),
+			Family:  ip.FamilyV4,
+		}
+
+		err = route.Add()
+		if err != nil {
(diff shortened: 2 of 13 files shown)


gwenya:
something is wrong with the tests I wrote but I can't figure out how to run them locally to properly figure out the issue :/

stgraber:
I think it can probably just be `ipv4.dhcp.routes`, no need for `static-` in the name

stgraber:
> something is wrong with the tests I wrote but I can't figure out how to run them locally to properly figure out the issue :/

That's fine, we don't pay for the Github runners ;)

Just push some debug commits to dump the network config, instance config, full output of the routing table in the container, ...

It may be as simple as the tiny busybox image on the Ubuntu 22.04 runners not supporting option 121.

gwenya:
I think we might not be running a dhcp4 client (or server?) at all in the tests, is that possible? I just did a run with default settings, and not only does the machine not have an ipv4 on its eth0, `incus network list-leases` shows that there aren't any ipv4 leases. 

The only other test that looks at the leases is network.sh, and that one only tests static leases.

I'm not sure if and how it is possible to test this in the github runners, and right now I'm having trouble testing it on my laptop because any bridge networks I create don't get dhcp at all for some reason (works fine on ovn though, so probably a dnsmasq issue)

gwenya:
oh wait nvm the issue with my bridges locally is just firewall >.<

gwenya:
looking into `udhcpc`, it appears that it doesn't configure anything itself, and instead just passes the dhcp information it receives to a script (`/usr/share/udhcpc/default.script` by default). So even after running `udhcpc` in the test image, no addresses or routes are installed because there is no script for it.
I'm thinking of either providing a script that just prints out the information this test cares about so that it can be grep'ed, or just try to grep the output of `udhcpc`.

gwenya:
I now have the tests with bridge network passing.

OCI get skipped because offline mode, and OVN isn't available on the test runners so I made them skip if the ovn network creation fails (I think we don't have a way of detecting if OVN is available).

gwenya:
I think this probably needs an API extension as well, should I just add one?

gwenya:
I think there is still a bug with the ovn networks, it seems these routes now override the route to the uplink dns server, and I'm gonna have to dig a bit deeper to figure out what is going on there, and why we are routing the uplink DNS in the first place.

stgraber (approved):

PR #1738, opened 2025-03-09
Fix DNS for isolated OVN networks

gwenya:
Currently DHCP in OVN networks is configured to advertise the uplink DNS address, if an uplink exists. OVN networks with the `network=none` option don't get any DNS server over DHCP. 

Since OVN has its own DNS server, we can just use the OVN router IP. 

I don't understand why this isn't used in general, and how the uplink DNS even gets the correct information, but at least for isolated networks I think it makes sense to use OVN for the DNS server.

### DIFF
--- a/internal/server/network/driver_ovn.go
+++ b/internal/server/network/driver_ovn.go
@@ -2660,6 +2660,8 @@ func (n *ovn) setup(update bool) error {
 
 		if uplinkNet != nil {
 			opts.RecursiveDNSServer = uplinkNet.dnsIPv4
+		} else {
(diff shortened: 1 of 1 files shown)


stgraber:
DCO check failed (missing signed-off-by).

OVN doesn't really have a DNS server, what it has is the ability to respond to DNS queries through flow rules. So it doesn't actually matter at all what the IP address is set to, if OVN sees it's a DNS request and it has a reply for that request, it will just send it back to the client rather than hitting the actual DNS server.

stgraber:
> Does that mean when it's using the uplink dns server it also gets responded to by OVN?

Indeed, it does.

stgraber (approved):

PR #1739, opened 2025-03-09
Add option to configure DNS server for bridge and OVN networks

gwenya:
This PR adds two options, `ipv4.dhcp.dns` and `ipv6.dhcp.dns` for both bridge and OVN networks. The values of these options are used as option 6 in dhcp4 and option 23 in dhcp6 respectively, i.e. advertising DNS servers to DHCP clients. The values in `ipv6.dhcp.dns` are also sent as RDNSS in router advertisements by dnsmasq on bridge interfaces, and the first value is configured for router advertisements in OVN (it only supports a single RDNSS value for RA).

If the options are not set, then the current behaviour (dnsmasq advertises itself and OVN networks use whatever their uplink provides) is used.

### DIFF
--- a/internal/server/network/ovn/ovn_nb_actions.go
+++ b/internal/server/network/ovn/ovn_nb_actions.go
@@ -1262,14 +1262,18 @@ func (o *NB) UpdateLogicalSwitchDHCPv4Options(ctx context.Context, switchName OV
 
 	if opts.Router != nil {
 		dhcpOption.Options["router"] = opts.Router.String()
+	} else {
+		delete(dhcpOption.Options, "router")
 	}
 
 	if len(opts.DNSSearchList) > 0 {
 		// Special quoting to allow domain names.
 		dhcpOption.Options["domain_search_list"] = fmt.Sprintf(`"%s"`, strings.Join(opts.DNSSearchList, ","))
+	} else {
+		delete(dhcpOption.Options, "domain_search_list")
 	}
 
-	if opts.RecursiveDNSServer != nil {
+	if len(opts.RecursiveDNSServer) > 0 {
 		nsIPs := make([]string, 0, len(opts.RecursiveDNSServer))
 		for _, nsIP := range opts.RecursiveDNSServer {
 			if nsIP.To4() == nil {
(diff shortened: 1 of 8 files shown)


gwenya:
if #1738 gets merged then this PR needs to be manually rebased and fixed, automatic merge of the if/else condition breaks the logic

gwenya:
I think that test failure is a flake

gwenya:
api extension is still missing, I can write that if you'd like me to

stgraber:
And yeah, on the commit side, what I would have done before merging is:
 - Add a commit at the beginning of the set for `api: network_dns_servers` which adds the extension to both `doc/api-extensions.md` and `internal/version/api.go`
 - Rename the first commit to `incusd/network/bridge: Add dns.servers`
 - Rename the second commit to `incusd/network/ovn: Add dns.servers`
 - Rename the third commit to `doc/network: Add dns.servers to bridge and OVN`

Anyway, that's pretty nitpicky stuff and very subjective on my part, so that's why I usually just do it myself before merging :)

It's mostly following our usual pattern which makes it slightly easier to look things up in a git log or when I cherry-pick changes into the LTS branch.

gwenya:
> I feel that `ipv6.dhcp.dns` working when `ipv6.dhcp` is `false` is going to be a bit confusing :)

oh oops I completely failed to realise that, yeah that's rather confusing.

> What do you think of just making it `dns.servers` instead and having it override the default if at least one server is listed for the family.

We already have `dns.nameservers` for physical networks, should we just reuse that name? Or would that also be confusing since on a physical network it has a somewhat different effect?

Should also probably write a bit of extra documentation that the dns servers only get pushed via DHCP except  that the first of the IPv6 ones gets pushed via RA as well. Would that belong in the table with the options (might get a bit long) or somewhere outside?

stgraber:
> We already have `dns.nameservers` for physical networks, should we just reuse that name? Or would that also be confusing since on a physical network it has a somewhat different effect?

Yep, would be good to reuse the name

stgraber:
> Should also probably write a bit of extra documentation that the dns servers only get pushed via DHCP except that the first of the IPv6 ones gets pushed via RA as well. Would that belong in the table with the options (might get a bit long) or somewhere outside?

Yeah, we can put it in the table.

This is one piece of documentation that still needs to be ported over to `gendoc`.
Once it's moved to `gendoc` then we'll have the ability to put in both short and long comments on options which will make this a lot cleaner.

gwenya:
> So if you have `dns.servers=1.1.1.1`, then that overrides the IPv4 DNS server, if you have `dns.servers=1.1.1.1, 2606:4700:4700::1111`, then it overrides both IPv4 and IPv6.

@stgraber I thought about this some more and I believe we should have it always override both if set, i.e. with `dns.nameservers=1.1.1.1` there would be `1.1.1.1` for ipv4 and no nameservers for ipv6 (dnsmasq *should* support that according to the documentation, but I haven't tried it yet). My reasoning for this is that it is less surprising and in line with the `dns.nameservers` setting on physical networks, which when set always overrides both ipv6 and ipv6 nameservers for attached OVN networks. I am not strongly against your suggestion though.

stgraber:
> @stgraber I thought about this some more and I believe we should have it always override both if set, i.e. with `dns.nameservers=1.1.1.1` there would be `1.1.1.1` for ipv4 and no nameservers for ipv6 (dnsmasq _should_ support that according to the documentation, but I haven't tried it yet). My reasoning for this is that it is less surprising and in line with the `dns.nameservers` setting on physical networks, which when set always overrides both ipv6 and ipv6 nameservers for attached OVN networks. I am not strongly against your suggestion though.

Fine with me. I like consistency :)

internal/server/network/ovn/ovn_nb_actions.go
```diff
@@ -1255,14 +1255,18 @@ func (o *NB) UpdateLogicalSwitchDHCPv4Options(ctx context.Context, switchName OV
 
 	if opts.Router != nil {
 		dhcpOption.Options["router"] = opts.Router.String()
+	} else {
```
gwenya, line 1258: I added these else branches for all the options for consistency, but they're probably not strictly in scope for this PR, I can move them to a new one if desired.

The delete is only really required for the dns servers (and the routes, to which I already added it in the PR for those) because otherwise they would never get set to empty if so configured.

The other options are always set anyways, but in case that ever changes it's probably nice if they all behave in the same way.

stgraber (approved):

PR #1753, opened 2025-03-12
Use lego binary for DNS-01 challenge

### LINKED ISSUE #1722: Use external `lego` binary rather than built-in Go package
Lego with DNS-01 pulls a lot of dependencies which results in a pretty significant increase in size to the binary and ultimately our base memory footprint.

To fix that, the easiest is to interact with the external `lego` binary.

accuser:
This PR replaces the recently added DNS-01 challenge provider with an invocation of the [LEGO](https://github.com/go-acme/lego) binary to address #1722, removing the need for bundling the entirety of the supported LEGO 3rd-party DNS clients.

ACME configuration remains the same.

Closes #1722


accuser:
I've tested with my own DNS (Gandi) and it seems to be working. I forgot to use LE's [staging environment](https://letsencrypt.org/docs/staging-environment/) and managed to go over rate, so please don't make the same mistake... 🤣 

But please review and let me know if this approach is suitable.

accuser:
One thought I have had, but probably not for this PR: the HTTP-01 challenge could also be implemented using the `lego` binary.

stgraber:
@accuser yeah, since we'll be shipping the lego binary anyway, we may as well move all of it to the external binary.

internal/server/acme/acme.go
```diff
@@ -99,96 +101,147 @@ func UpdateCertificate(s *state.State, provider ChallengeProvider, clustered boo
...
-	}
+		cmd := exec.Command("lego", args...)
+		cmd.Env = append(os.Environ(), environment...)
 
-	err = provider.RegisterWithSolver(client.Challenge)
-	if err != nil {
-		return nil, fmt.Errorf("Failed setting challenge provider: %w", err)
-	}
+		logger.Info("Running lego command", logger.Ctx{"cmd": cmd.String()})
+		err = cmd.Run()
```
stgraber, line 144: Should be able to use our `subprocess.RunCommandSplit` for this instead.

`_, _, err := subprocess.RunCommandSplit(ctx, append(os.Environ(), environment), nil, "lego", args...)`

internal/server/acme/acme.go
```diff
@@ -99,96 +101,147 @@ func UpdateCertificate(s *state.State, provider ChallengeProvider, clustered boo
...
 
-	if err != nil {
-		return nil, fmt.Errorf("Failed to register user: %w", err)
-	}
+		if caURL != "" {
+			config.CADirURL = caURL
+		} else {
+			// Default URL for Let's Encrypt
+			config.CADirURL = "https://acme-v02.api.letsencrypt.org/directory"
+		}
```
stgraber: Shouldn't be needed now that you set the default value in the config package

cmd/incusd/acme.go
```diff
@@ -78,7 +78,7 @@ func autoRenewCertificate(ctx context.Context, d *Daemon, force bool) error {
 
 	domain, email, caURL, agreeToS, challengeType := s.GlobalConfig.ACME()
 
-	if domain == "" || email == "" || !agreeToS || challengeType == "" {
+	if domain == "" || email == "" || !agreeToS {
```
stgraber, line 81: We should still keep this one. challengeType is still returned and it has a default value so it's fine to ensure it's not empty.

cmd/incusd/acme.go
```diff
@@ -97,26 +97,8 @@ func autoRenewCertificate(ctx context.Context, d *Daemon, force bool) error {
...
-		if challengeProvider == nil {
-			return nil
-		}
-	} else {
-		challengeProvider = d.http01Provider
-	}
-
 	opRun := func(op *operations.Operation) error {
-		newCert, err := acme.UpdateCertificate(s, challengeProvider, s.ServerClustered, domain, email, caURL, force)
+		newCert, err := acme.UpdateCertificate(s, d.http01Provider, s.ServerClustered, domain, email, caURL, challengeType, force)
```
stgraber, line 101: I'd probably move `challengeType` to be the second argument here

stgraber:
Looks good, just a few minor tweaks needed. I'll take care of those myself now as that should be rather quick.

stgraber (approved):

PR #1756, opened 2025-03-12
Add project support to storage pools in preseed init

megheaiulian:
### Add project support to storage pools in preseed init.

This allow to initialize `incus` with storage pools for different projects.

### DIFF
--- a/client/incus_server.go
+++ b/client/incus_server.go
@@ -210,73 +210,66 @@ func (r *ProtocolIncus) ApplyServerPreseed(config api.InitPreseed) error {
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


stgraber:
I'm very confused by this one.

Storage pools are one of the very few global entities in Incus.
Only a global administrator can create them and they are never part of projects.

megheaiulian:
I am sorry I think misunderstood how this works currently.
i thought that since networks and storage volumes can be set per project that it makes sense to have the storage pools work the same way. I can see now there is no feature flag for projects to have their own storage pools.

PR #1759, opened 2025-03-13
Use the correct path for ingesting DNS-01 challenge certificate outputs

accuser:
The path used in #1753 is incorrect, and certificates should be being loaded from the `certificates` subfolder. This PR fixes this issue.


accuser:
@stgraber not sure how I missed this on the previous PR as I had originally coded it this way, but it was likely when I moved the implementation internally. Sorry!

stgraber (approved):

Issue #1763, opened 2025-03-14
Use `lego` binary to handle HTTP-01 challenge


accuser:
Since #1753 the DNS-01 challenge is handled with the `lego` binary. HTTP-01 challenge can also now be handled with the `lego` binary.

Proposed changes:

- Add `core.http_address` to config.
- Add well-known HTTP server to `incusd` that is enabled when `core.http_address` is valid. This new server simply serves files from the `.well-known` subfolder of var root (i.e., `/var/lib/incus`).
- When updating certificates using HTTP-01 challenge, call `lego` with the `--http.port [port] --http.webroot [webroot]` settings so that the challenge token is written to the file system folder `[webroot]/.well-known/acme-challenge/` and available via `http://[domain]:[port]/.well-known/acme-challenge/[token]` 

I have this as a PoC in a local branch, but want to sense check this approach first before submitting a PR @stgraber.

stgraber:
SO in general, we don't want Incus to have an HTTP listener, we only do HTTPS.

Can `lego` itself run the http server during validation?

accuser:
It can - I'm just testing now. If this works, I'll submit a PR shortly.

PR #1770, opened 2025-03-14
Use lego binary for http 01 challenge

accuser:
This addresses #1763. I've used `lego` binary to solve HTTP-01 challenges.

New config keys:

`acme.http.port` - the interface and port to use with `lego` (maps to `--http.port`) and defaults to `:80` (defined by `lego`).


internal/ports/ports.go
accuser, line 1: This got refactored when I added HTTPDefaultPort, which is no longer required since using `lego` as the server for solving HTTP-01 challenge

internal/server/acme/acme.go
```diff
@@ -102,148 +95,66 @@ func UpdateCertificate(s *state.State, challengeType string, provider ChallengeP
...
+	certInfo, err = localtls.KeyPairAndCA(tmpDir+"/certificates", domain, localtls.CertServer, true)
 	if err != nil {
-		return nil, fmt.Errorf("Failed to obtain certificate: %w", err)
+		return nil, fmt.Errorf("Failed to load certificate and key file: %w", err)
 	}
 
-	l.Info("Finished issuing certificate")
-
-	return certificates, nil
+	return &certificate.Resource{
```
accuser, line 156: This is from `lego`, but is only used here for a convenient return type and to match the previous HTTP01Provider.

internal/ports/ports.go
accuser: I've added a comment to resolve.

stgraber:
I'm working on this one now. I think we're going to still support the `.well-known/acme-challenge/<token>` route to avoid breaking existing environments that rely on Incus' ability to respond to the challenge over HTTPS and having an external reverse proxy handle the HTTP to HTTPS part.

My plan is to simply keep the route alive and then forward any requests on there over to the `lego` listener.

accuser:
Re-adding the `--http.webroot` option will sort that, but yes, the endpoint would need to be revived.

stgraber (approved):

stgraber (approved):

PR #1771, opened 2025-03-14
incusd/network/ovn: Keep getting router name when network none

diegofernandes:


stgraber:
@diegofernandes I'll take it from there and go a bit further as there are a few more cases to think of here:
 - network=none => Gets you a Logical Router but no IPv4/IPv6 uplink addresses
 - network=none + ipv4.address=none + ipv6.address=none => Get you no Logical Router at all

So I think I'll take your change and then expand things so we also expose the Logical switch name.

stgraber (approved):

PR #1777, opened 2025-03-16
client: Fix spelling errors found by codespell

cjwatson:
My plan is to fix all the reported errors and then add `codespell` to `static-analysis`; like any such tool it's imperfect, but I've generally found its maintainers to have pretty good judgement.  This is just a small initial PR to check that you're OK with using it in principle; if so then I'll send a larger PR (~2500 lines) to fix all the remaining errors and enable it as part of `make static-analysis`.  (If you'd rather not have it as an explicit lint step, then I guess most of the typos will be worth fixing anyway, so I can still send those.)

### DIFF
--- a/client/incus_instances.go
+++ b/client/incus_instances.go
@@ -466,7 +466,7 @@ func (r *ProtocolIncus) GetInstanceFull(name string) (*api.InstanceFull, string,
 	instance := api.InstanceFull{}
 
 	if !r.HasExtension("instance_get_full") {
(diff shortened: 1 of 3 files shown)


stgraber (approved):

stgraber:
Thanks! I believe we have `golangci-lint` running `misspell` already but that's definitely missed a whole bunch of typoes over the years, so likely not as comprehensive as codespell :)

PR #1778, opened 2025-03-16
Fix spelling errors and run codespell automatically

cjwatson:
See #1777.

I skipped a few things: the main one was "requestor" (curated dictionaries only seem to have "requester"), which looked like too much trouble to change.


cjwatson:
Ugh, looks like the ignore mechanism doesn't work quite the same way with older versions.  I'll try to figure that out.

stgraber:
@cjwatson yeah, same as with the distrobuilder PR. I'd just pull the latest upstream version with pip/pipx when run in CI. That way we get the latest upstream logic and avoid the 22.04 issue.

cjwatson:
OK, it took me some time to pacify all the various CI steps, but it should work now.

internal/server/apparmor/instance_lxc.go
```diff
@@ -27,7 +27,7 @@ profile "{{ .name }}" flags=(attach_disconnected,mediate_deleted) {
   signal peer=@{profile_name},
 
   # Allow normal process handling
-  ptrace (readby),
+  ptrace (readby),  # codespell:ignore readby
```
stgraber, line 30: Those hints in the apparmor generated profiles seem less than ideal as that ends up on the actual end user system and gets fed to the apparmor parser.

Is there a way to put the comment above the entire template block rather than in-line within the profile?

stgraber: Doesn't seem to be possible. I'm going to just split the actual profiles into separate files that we can then just ignore.

stgraber: Done

stgraber (approved):

PR #1781, opened 2025-03-17
Add ipv4.dhcp.expiry option for ovn networks

gwenya:


stgraber:
Odd that they don't support a matching setting on DHCPv6.

stgraber (approved):

gwenya:
> Odd that they don't support a matching setting on DHCPv6.

@stgraber yeah I thought that too, but the OVN DHCP is completely stateless and, apart from when one manually configures a specific address for an instance, gives out exactly the same address that the instance would have gotten via SLAAC. So, again apart from the manually configured addresses, there isn't a need for lease times.

Issue #1730, opened 2025-03-06
Report CPU count in `/state` endpoint


bensmrs:
Currently, only CPU time is reported in the `/state` API endpoint for instances. While the total CPU time can be interesting to some, I think most users are actually interested in a CPU usage in percents (which we can calculate easily) relative to the CPU count (which we don’t have access to).

stgraber:
I'm not sure how you'd get a percentage value by knowing the current CPU count.

The data we get from the kernel is the total CPU time since start of the instance, this doesn't account for any change in the number of CPUs available to the instance during its lifetime or changes in CPU availability on the host such as isolated CPUs or similar dedicated allocations to other processes on the system.

So knowing that your container used 300s of CPU time, has been running for a week and is on a 64 core system would have your report 0.000775% of CPU usage. That doesn't seem terribly useful to me.

I suspect what most folks are actually after is an instant CPU consumption in percentage but that we definitely don't have the data for.

bensmrs:
> I'm not sure how you'd get a percentage value by knowing the current CPU count.

No, but doing $100\times\frac{dm\left(t\right)}{dt}$ (with $m$ the measure of CPU time), you get the used CPU time as a percentage of elapsed time. Divide it by the CPU count and you get a percentage relative to all the CPUs.

> The data we get from the kernel is the total CPU time since start of the instance, this doesn't account for any change in the number of CPUs available to the instance during its lifetime or changes in CPU availability on the host such as isolated CPUs or similar dedicated allocations to other processes on the system.

I’m not sure that invalidates my point.

> So knowing that your container used 300s of CPU time, has been running for a week and is on a 64 core system would have your report 0.000775% of CPU usage. That doesn't seem terribly useful to me.

It’s not, so I’m not suggesting exposing the CPU percentage, only CPU time and CPU count. The idea is to allow a WebUI (to be honest, I happen to have started developing one a few months ago) to show immediate (well, after at least 2 measurements) CPU usage in a meter bar, as you currently do on the branded UI with memory and disk, and as most other hypervisor UIs do. On a meter bar, 100% does not mean 100% CPU time, but 100% CPU time on all the CPUs, as 100% CPU time means **one** CPU being at 100% capacity.

> I suspect what most folks are actually after is an instant CPU consumption in percentage but that we definitely don't have the data for.

Well, we do, if we have the CPU count.

stgraber:
[...]

PR #1807, opened 2025-03-18
Add allocated CPU time to instance state

bensmrs:
Closes: #1730

### DIFF
--- a/shared/api/instance_state.go
+++ b/shared/api/instance_state.go
@@ -98,6 +98,12 @@ type InstanceStateCPU struct {
 	// CPU usage in nanoseconds
 	// Example: 3637691016
 	Usage int64 `json:"usage" yaml:"usage"`
+
+	// CPU time available per second, in nanoseconds
+	// Example: 4000000000
+	//
+	// API extension: instance_state_cpu_time
+	AllocatedTime int64 `json:"allocated_time" yaml:"allocated_time"`
 }
 
 // InstanceStateMemory represents the memory information section of an instance's state.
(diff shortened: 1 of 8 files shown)


bensmrs:
Oh I forgot adding an API extension…

stgraber:
Looking pretty good, though as I just mentioned in the issue, it'd probably be best to entirely rely on cgroups in the container case rather than parse the limits as the underlying OS may have applied something stricter (particularly when no limits are set).

Obviously for the VM case, that doesn't really apply, so on that front, we can just assume full CPUs based on the limits.cpu as that controls the number of VM threads directly (and also has a fixed default when not set).

bensmrs:
Alright, I’ll rewire stuff tomorrow to take your suggestions into account

bensmrs:
The more I think about it, the more I believe that percentage-based allocation doesn’t really impact the time allocation. I’ll modify the PR and only read the CFS limits.

stgraber:
Yeah, percentage based allocation only turns into CPU shares which are only meaningful when the machine is overloaded, so rarely useful. CPU sets and CFS quotas are the main game here.

shared/api/instance_state.go
```diff
@@ -98,6 +98,10 @@ type InstanceStateCPU struct {
 	// CPU usage in nanoseconds
 	// Example: 3637691016
 	Usage int64 `json:"usage" yaml:"usage"`
+
+	// CPU time available per second, in nanoseconds
+	// Example: 4000000000
```
stgraber, line 103: Missing API extension comment here

stgraber:
Looks good, just that one tiny nit but I'll take care of it as well as re-order the commits.

stgraber (approved):

Issue #1697, opened 2025-02-25
Add new disk type: usb-storage


foxtrotcz:
Hi,
this issue basically solves my previous one: https://github.com/lxc/incus/issues/1688

I propose implementing `usb-storage` as new disk type.
https://qemu-project.gitlab.io/qemu/system/devices/usb.html#connecting-usb-devices

This device connects to `qemu-xhci` which is already implemented in Incus for USB passthrough. It also supports hotplugging.
Drivers for USB device are also preloaded in Windows installer so it can be used as media for mounting additional drivers during Windows install. It can be also used for mounting Windows install ISO which will then allow to install easily on NVMe disk.

I tested it via qemu monitor and it looks to me it works nicely.
```
drive_add 0 if=none,id=stick,media=cdrom,format=raw,file=/home/user/drivers.iso,file.locking=off
device_add usb-storage,bus=qemu_usb.0,drive=stick,id=usbstorage
```

bensmrs:
Well, if we do this, can we do the same with network adapters?
Would solve my macOS problems :)

foxtrotcz:
@bensmrs I will post here my chat conversation with @stgraber which basically answers this.

> FoxtrotCZ
> [10:05 pm](https://discuss.linuxcontainers.org/chat/c/-/7/247)
> [@stgraber](https://discuss.linuxcontainers.org/u/stgraber) Hi, just a question about network driver in Incus. Currently there is only virtio-net. Would you consider implementing also e1000 as alternative?
> 
> I found this thread, do I understand correctly that you will not add any other alternatives or was that meant only for Snap version? [First class support for alternative NIC drivers?](https://discuss.linuxcontainers.org/t/first-class-support-for-alternative-nic-drivers/16206)
> Thanks.
> 
> (edited)
> stgraber
> [10:28 pm](https://discuss.linuxcontainers.org/chat/c/-/7/248)
> We are trying to avoid fully virtualized drivers as much as possible as they are the main source of security issues. So as it stands, we do not support any non-paravirtualized drivers, your only way to use them is through raw.qemu which makes that instance unsupported.
> 
> FoxtrotCZ
> [10:30 pm](https://discuss.linuxcontainers.org/chat/c/-/7/249)
> Thanks, I just ask so I dont uselessly create feature request on github that will be denied :slight_smile:
> 
> In storage you dont mind non-virtio drivers as long as they are hotpluggable.
> 
> stgraber
[...]

PR #1814, opened 2025-03-19
Add support for USB NICs

bensmrs:
This PR allows to use USB emulation for NICs through an `io.bus` device option.
It introduces a `UseUSBBus` field to the `RunConfig` struct, which may also help for #1697

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2752,3 +2752,7 @@ Introduces `ipv4.dhcp.expiry` for OVN networks.
 ## `instance_state_cpu_time`
 
 This adds an `allocated_time` field below `CPU` in the instance state API.
+
+## `network_io_bus`
+
+This introduces a new `io.bus` property for compatible network devices allowing to choose between `virtio` (default) and `usb`.
--- a/doc/reference/devices_nic.md
+++ b/doc/reference/devices_nic.md
@@ -10,6 +10,11 @@ NICs support hotplugging for both containers and VMs (with the exception of the 
 Network devices, also referred to as *Network Interface Controllers* or *NICs*, supply a connection to a network.
 Incus supports several different types of network devices (*NIC types*).
 
+```{note}
+When using a USB network adapter with a VM, mainline QEMU will replace the leading two bytes of a MAC address with `40:`.
+Those affected by this may want to manually set the `hwaddr` property to a MAC address starting with `40:` to align the host and guest reporting of the MAC.
+```
+
 ## `nictype` vs. `network`
 
 When adding a network device to an instance, there are two methods to specify the type of device that you want to add: through the `nictype` device option or the `network` device option.
@@ -75,6 +80,7 @@ Key                                   | Type    | Default           | Managed | 
 `boot.priority`                       | integer | -                 | no      | Boot priority for VMs (higher value boots first)
 `host_name`                           | string  | randomly assigned | no      | The name of the interface inside the host
 `hwaddr`                              | string  | randomly assigned | no      | The MAC address of the new interface
+`io.bus`                              | string  | `virtio`          | no      | Override the bus for the device (can be `virtio` or `usb`) (VM only)
 `ipv4.address`                        | string  | -                 | no      | An IPv4 address to assign to the instance through DHCP (can be `none` to restrict all IPv4 traffic when `security.ipv4_filtering` is set)
 `ipv4.routes`                         | string  | -                 | no      | Comma-delimited list of IPv4 static routes to add on host to NIC
 `ipv4.routes.external`                | string  | -                 | no      | Comma-delimited list of IPv4 static routes to route to the NIC and publish on uplink network (BGP)
@@ -123,6 +129,7 @@ Key                     | Type    | Default           | Managed | Description
 `boot.priority`         | integer | -                 | no      | Boot priority for VMs (higher value boots first)
 `gvrp`                  | bool    | `false`           | no      | Register VLAN using GARP VLAN Registration Protocol
 `hwaddr`                | string  | randomly assigned | no      | The MAC address of the new interface
+`io.bus`                | string  | `virtio`          | no      | Override the bus for the device (can be `virtio` or `usb`) (VM only)
 `mode`                  | string  | `bridge`          | no      | Macvlan mode (one of `bridge`, `vepa`, `passthru` or `private`)
 `mtu`                   | integer | parent MTU        | yes     | The MTU of the new interface
 `name`                  | string  | kernel assigned   | no      | The name of the interface inside the instance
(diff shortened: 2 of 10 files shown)


internal/server/instance/instance_utils.go
```diff
@@ -464,10 +464,18 @@ func LoadFromBackup(s *state.State, projectName string, instancePath string, app
...
-func DeviceNextInterfaceHWAddr() (string, error) {
-	// Generate a new random MAC address using the usual prefix
+func DeviceNextInterfaceHWAddr(usb bool) (string, error) {
+	var prefix string
+	if usb {
+		// Use a prefix that suits QEMU for USB NICs
+		prefix = "40:xx:xx:xx:xx:xx"
+	} else {
+		// Use the usual prefix
+		prefix = "10:66:6a:xx:xx:xx"
```
stgraber, line 474: Nit but we like comments to end with a period these days ;)

stgraber:
So I've been looking into the QEMU side of things.

As far as I can tell, there are effectively two distinct fields, one that holds the full provided MAC and once that's derived from that MAC minus the first two bytes being switched to `0x40`.

When looking in the VM, I don't actually see the MAC address that's set by Incus at all.
Do you have the same behavior in your case?

Stéphane

stgraber:
Oh, testing this stuff is made a bit hardware due to the fact that this branch doesn't seem to implement hotplug correctly yet. So adding a `io.bus=usb` NIC after the instance has started has it land on virtio-net instead.

bensmrs:
> When looking in the VM, I don't actually see the MAC address that's set by Incus at all.
Do you have the same behavior in your case?

You see a NIC with another address? On my machine at least I have the correct address.

> this branch doesn't seem to implement hotplug correctly yet

Oh that’s something I haven’t tested yet. I’ll have a look.

stgraber:
> You see a NIC with another address? On my machine at least I have the correct address.

I'm pretty sure I got a completely random MAC earlier here, but trying it again now, I do see the "expected" behavior of the MAC address starting with "40:xx".

The problem is that we can't really generate such a MAC. "40:" is not a reserved prefix by the IEEE and I see a lot of entries in "40:xx", so we're effectively violating MAC allocation rules here.


Honestly what I'd do at this point is just keep our MAC addresses the way we do it normally, so feed it a valid MAC and let QEMU mangle the first two bytes to "0x40". We can then leave a note in the doc that this is happening and is out of our control.

I'll ask on QEMU's IRC about what's up with that, see if anyone knows.
I tried a physical USB network adapter and it has a normal MAC, no "40:" prefix.
I also failed to find any standard requiring the 0x40 prefix.

stgraber:
I asked on IRC now. Personally I'm very tempted to just push a one-line patch to QEMU in the Zabbly package that removes that limitation and see what happens ;)

Looking at the libvirt code, they're passing a normal MAC and actually show an example with a MAC starting with ("00:"), I guess they never noticed the rewriting happening.

bensmrs:
Ok I got hotplugging working; I’ll run more tests soon.

> I'm pretty sure I got a completely random MAC earlier here, but trying it again now, I do see the "expected" behavior of the MAC address starting with "40:xx".

*Ahem* now that you’re mentioning it, I just saw this behavior. And it disappeared… Weird.

> The problem is that we can't really generate such a MAC. "40:" is not a reserved prefix by the IEEE and I see a lot of entries in "40:xx", so we're effectively violating MAC allocation rules here.

Well it’s a free allocation range, so we can define a range for Incus use and it’ll be fine. What problems do you see?

> Honestly what I'd do at this point is just keep our MAC addresses the way we do it normally, so feed it a valid MAC and let QEMU mangle the first two bytes to "0x40". We can then leave a note in the doc that this is happening and is out of our control.

Sure, but if you have your network manager configured to get a DHCP lease for all its interfaces, if you don’t explicitly define an `hwaddr`, you won’t get a lease from Incus. That’s why I modified the address generation logic.

> I'll ask on QEMU's IRC about what's up with that, see if anyone knows.

I asked a few months ago when I started tinkering with macOS, but no answer…

> I also failed to find any standard requiring the 0x40 prefix.

Yeah I’ve done my research on that and… well I haven’t found anything. But also the commit introducing the `40:` prefix is from December 2009…
It’s a pretty old, rarely-touched driver.

> Personally I'm very tempted to just push a one-line patch to QEMU in the Zabbly package that removes that limitation and see what happens

Oh my.

stgraber:
> Oh my.

I pushed that change to the Zabbly daily package now. Will be easy to test and see if anything obviously bad happens from it :)

bensmrs:
So do you want me to remove all the logic dealing with `40:`? I can document that MACs not starting with `40:` only work with the Zabbly flavor of QEMU. It’ll simplify the code quite a bit. I’ll do some debugging tomorrow to try to reproduce the strange behavior that we noticed with wrong random MACs appearing.

stgraber:
Yeah, we should remove the check because otherwise we're forcing people into using MAC addresses that they're not supposed to be using...

I haven't yet tested the new QEMU but if it works and nobody sees anything obviously wrong with it, I guess we should just send the patch upstream and see if someone finally tells us what's going on with that thing :)

bensmrs:
Oh btw, do we agree that [that line](https://github.com/lxc/incus/pull/1814/files#diff-992fb4e9f1a365b2b538e69f02f33d68f836de902083a76706f5c64a89ac8f3fL4484) is a leftover from an old logic that doesn’t make sense anymore?

stgraber:
> Oh btw, do we agree that [that line](https://github.com/lxc/incus/pull/1814/files#diff-992fb4e9f1a365b2b538e69f02f33d68f836de902083a76706f5c64a89ac8f3fL4484) is a leftover from an old logic that doesn’t make sense anymore?

Github seems a bit confused about what's being linked here.
If it's about the vectors math, then yeah, that code looks a bit odd.

We have a minimum of two queues which then means a minimum of 6 vectors, so the value can never be 0 and should just always enable multi-queue (mq) and expose the number of vectors when on PCI.

bensmrs:
> Github seems a bit confused about what's being linked here.
If it's about the vectors math, then yeah, that code looks a bit odd.

The auto-scroll is broken but the highlight is still there. It was indeed about the vector math.

> I haven't yet tested the new QEMU but if it works and nobody sees anything obviously wrong with it, I guess we should just send the patch upstream and see if someone finally tells us what's going on with that thing :)

For now at least it works.

I’ll try to wrap the issue up soon, and possibly track the MAC problem we encountered, although I’m afraid I’ll end up debugging QEMU…

bensmrs:
A few questions/remarks.

1. I could only get the random MAC twice, and twice the same: ca:89:78:d0:7f:76, dev enp1s0f6u4i1.

By any chance, is it the same MAC you get when it goes wrong?

2. Changing `io.type` (hot or cold) leads to the device losing its IP, even after a reboot. Reverting `io.type` works.

I’m pretty sure it’s Linux doing its thing. The sheer idea of doing that feels cursed enough for me to not even consider documenting the behavior. Wdyt?

3. Pretty often, when I plug 2 `usb-net` devices, only one of them gets an IP.

Is it something you can reproduce on your side?

bensmrs:
Just now, I got the same wrong MAC. The QMP command is the correct one, so I’m quite afraid to dig deeper.

bensmrs:
Couldn’t find anything in the QEMU driver for `usb-net`.
The fact that we experienced such a bug on an unpatched QEMU makes me think QEMU may not be the culprit but rather the kernel, although I couldn’t find anything inspiring in the code dealing with RNDIS and CDC devices.

stgraber:
Yeah, I agree that it's probably some USB probing issue or something in the kernel.
I believe I actually saw that before with a physical USB network adapter.

So anyway, our side of things is correct so we can move on with that.

stgraber (approved):

bensmrs:
For the record, when the interface is bugged, I get in dmesg:
```
[    6.864508] cdc_subset 1-4:1.0: probe with driver cdc_subset failed with error -22
[    6.983942] cdc_subset 1-4:1.1 usb0: register 'cdc_subset' at usb-0000:01:00.6-4, Linux Device, 22:3b:42:e7:ff:eb
[    6.984958] usbcore: registered new interface driver cdc_subset
[    6.985127] cdc_ether 1-4:1.0: probe with driver cdc_ether failed with error -16
```

PR #1835, opened 2025-03-24
Add `io.bus=usb` for disks

bensmrs:
Closes: #1697

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2756,3 +2756,7 @@ This adds an `allocated_time` field below `CPU` in the instance state API.
 ## `network_io_bus`
 
 This introduces a new `io.bus` property for compatible network devices allowing to choose between `virtio` (default) and `usb`.
+
+## `disk_io_bus_usb`
+
+Adds a new `usb` value for `io.bus` on `disk` devices.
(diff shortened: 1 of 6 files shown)


stgraber:
Needs a `make update-metadata`  run.

bensmrs:
Note: I have no idea whether live migration works out of the box or if we have to disable it, like with NVMe (btw, could you point me to where QEMU says it doesn’t work? I couldn’t find it).
Also, the `UseUSBBus` field introduced in #1814 isn’t used here, so I think we should try to remove the `RunConfigItem` type (I really don’t get why it exists) so we can simplify the logic and easily remove `UseUSBBus`.

stgraber:
> Note: I have no idea whether live migration works out of the box or if we have to disable it, like with NVMe (btw, could you point me to where QEMU says it doesn’t work? I couldn’t find it).

QEMU fails with an error telling you it's not supported. I don't remember seeing it in documentation but we confirmed that there's missing logic in the code to handle it.


To test if live migration would work, you don't need a cluster or even multiple systems.
You can try with:
 - incus config set NAME migration.stateful=true
 - incus config device override NAME root size.state=1GiB
 - incus start NAME
 - incus stop --stateful NAME
 - incus start NAME

That effectively live-migrates to a file and then restores it.

stgraber:
No problem with stateful stop/start here, so I think this is fine for live migration.

stgraber (approved):

PR #1871, opened 2025-03-31
instances/drivers/qemu: update user parameter for QEMU v9.1+

dnegreira:
Replace deprecated `-runas` with supported `-run-with user` parameter for QEMU versions 9.1 and above.

Closes: #1870

### DIFF
--- a/internal/server/instance/drivers/driver_qemu.go
+++ b/internal/server/instance/drivers/driver_qemu.go
@@ -1653,7 +1653,15 @@ func (d *qemu) start(stateful bool, op *operationlock.InstanceOperation) error {
 
 	// Attempt to drop privileges (doesn't work when restoring state).
 	if !stateful && d.state.OS.UnprivUser != "" {
-		qemuArgs = append(qemuArgs, "-runas", d.state.OS.UnprivUser)
+		qemuVer, _ := d.version()
+		qemuVer91, _ := version.NewDottedVersion("9.1.0")
+
+		// Since QEMU 9.1 the parameter `runas` has been marked as deprecated.
+		if qemuVer != nil && qemuVer.Compare(qemuVer91) >= 0 {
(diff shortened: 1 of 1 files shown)


bensmrs:
`-runas` has been marked for deprecation since 9.1 (https://github.com/qemu/qemu/commit/95e0fb0afac678b13a70b476b4ad564569d6df8c), so you may want to update your check accordingly

internal/server/instance/drivers/driver_qemu.go
```diff
@@ -1653,7 +1653,16 @@ func (d *qemu) start(stateful bool, op *operationlock.InstanceOperation) error {
...
+		qemuVer91, _ := version.NewDottedVersion("9.1.0")
+		qemuVer, _ := d.version()
+		if qemuVer != nil {
+			// Since QEMU 9.1 the parameter `runas` has been marked as deprecated.
+			if qemuVer.Compare(qemuVer91) >= 0 {
+				qemuArgs = append(qemuArgs, "-run-with", fmt.Sprintf("user=%s", d.state.OS.UnprivUser))
+			} else {
+				qemuArgs = append(qemuArgs, "-runas", d.state.OS.UnprivUser)
+			}
+		}
```
stgraber, line 1665: In this case, if the version doesn't parse we end up with nothing.
I think we should instead be doing:

```go
if qemuVer != nil && qemuVer.Compare(qemuVer91) >= 0 {
// use run-with
} else {
// use runas
}
```

stgraber: That also has the benefit of making things a bit more compact and easier to read

dnegreira: Fixed on https://github.com/lxc/incus/pull/1871/commits/6eea371007bdea422d493b16e7ce0677aa709c36

stgraber (approved):

PR #1896, opened 2025-04-05
docs: mDNS setup for cluster HA

MOZGIII:
This PR proposes an mDNS based setup for cluster high availability docs.

Ref: https://datatracker.ietf.org/doc/html/rfc6762

### DIFF
--- a/doc/.wordlist.txt
+++ b/doc/.wordlist.txt
@@ -164,6 +164,7 @@ macOS
 macvlan
 Makefile
 manpages
+mDNS
 Mbit
 MiB
 Mibit
--- a/doc/howto/cluster_access.md
+++ b/doc/howto/cluster_access.md
@@ -89,3 +89,11 @@ Traffic will then get balanced between all servers and as soon as a server goes 
 ```{note}
 To minimize fallback delay, one can make use of BFD alongside BGP to get sub-1s fallback time.
 ```
+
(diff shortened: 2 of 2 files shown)


stgraber:
Hi @MOZGIII, looks like your commit message is missing the required `Signed-off-by:` line, see CONTRIBUTING.md for details

MOZGIII:
Yes, I'll need to clone locally to amend the commit. Does it look good otherwise?

stgraber:
The Signed-off-by line is there to confirm that you own the work and are allowed to contribute it to the project. Until that particular check passes, we don't actually look at the proposed changes to avoid being "tainted" in the event that you cannot sign the DCO and have to retract the PR.


That's more of a problem for larger code contributions than documentation changes, but the same rationale still applies. We need to know that the code we're looking at is legally safe for inclusion and only once we know that this is the case, do we look at it.

Otherwise we face the issue of having looked at code that we can't use, and have that prevent us from coming up with our own fix for the issue as we may intentionally or not plagiarize that code we've seen but couldn't include.


Anyway, that's the reason for the automatic DCO check and our general behavior around reviews of new PRs.

MOZGIII:
Reasonable, ok then. I'll need some time before I can sign off - Github UI doesn't let me do it after the fact, so gtg to the desk.

PR #1898, opened 2025-04-06
Usability improvements to incus-migrate

stgraber:


cmd/incus-migrate/main_migrate.go
```diff
@@ -586,6 +590,12 @@ func (c *cmdMigrate) Run(cmd *cobra.Command, args []string) error {
 		}
 	} else {
 		_, ext, convCmd, _ := archive.DetectCompression(config.SourcePath)
+
+		_, err := exec.LookPath(convCmd[0])
```
hallyn, line 594: ConvCmd coupd be nil, leading to a panic, right?

stgraber: Oops, indeed. I fixed it now by moving it into the if statement below.

stgraber: @hallyn should be good to go now

brauner (approved):

PR #1900, opened 2025-04-06
Fix cli action format string

c4t3l:
Fixes "non-constant format string" errors in fmt calls.


c4t3l:
After re-syncing my main branch and checking again I can see that you already addressed the fixes presented here in https://github.com/lxc/incus/pull/1883.  This PR should be closed.  I will continue to check static analysis on my end and open any PR that seems appropriate.

stgraber:
Yeah, I've made a bunch of the packages be `golangci-lint run` clean which covers some of the issues you've run into.

PR #1901, opened 2025-04-06
Fix cli admin format string

c4t3l:
Fixes "non-constant format string" errors in fmt calls.


c4t3l:
Closing this as these errors were already addressed in #1883

PR #1903, opened 2025-04-07
Add support for configurable logging targets 

### LINKED ISSUE #1420: Configurable syslog targets
Not everyone likes using Grafana Loki for logging. A common alternative which works as a protocol with most log management systems is good old syslog.

It would be good to extend our internal logging/event forwarding mechanism to also work over syslog.

I suspect we may want something like this:
 - logging.syslog.NAME.target (address of the remote syslog server)
 - logging.syslog.NAME.facility (syslog facility to use for the forwarded events)
 - logging.syslog.NAME.severity (syslog severity to use for the forwarded events, if not specified, `info` is used unless we have an internal severity already)
 - logging.syslog.NAME.type (logging, lifecycle or network-acl)
 - logging.syslog.NAME.loglevel (minimum log level when forwarding logging events)

(Once this lands, we'll want to standardize LOKI and move it to `logging.loki.NAME.XYZ` instead of its current single-server top-level at `loki.*`)

presztak:
This PR introduces support for configurable logging targets.
The first implemented targets are syslog and loki.

Closes #1420


stgraber (approved):

valyala:
How about adding [VictoriaLogs](https://docs.victoriametrics.com/victorialogs/)  logging target? Though VictoriaLogs [supports Loki data ingestion format](https://docs.victoriametrics.com/victorialogs/data-ingestion/promtail/), it is better to use JSON line data ingestion format instead, since it is easier to debug and reason about - every log entry is represented as a single JSON line with arbitrary log fields. See [these docs](https://docs.victoriametrics.com/victorialogs/data-ingestion/#json-stream-api).

An additional benefit of JSON line data ingestion format over Loki format is that it allows storing log fields with arbitrary number of unique values (aka high-cardinality log fields). Contrary to Loki, VictoriaLogs supports and encourages using such fields, and provides fast full-text search over these fields. See [data model docs](https://docs.victoriametrics.com/victorialogs/keyconcepts/) for details.

stgraber:
Should be pretty straightforward for someone to contribute a new logging driver now, provided there's either a good lightweight Go client available or the API is simple enough to just drive it directly.

PR #1920, opened 2025-04-10
Add generated documentation for network bridge

### LINKED ISSUE #1792: Port `bridge` network documentation to `gendoc`
The bridge network documentation (`doc/reference/network_bridge.md`) currently uses a static table of valid options. This should be moved to an auto-generated table using `gendoc`.

This is typically done in two steps:

1) Add the required `gendoc` comments where the configuration options are defined (usually alongside their validator)
2) Generate the new config metadata (`make update-metadata`)
3) Update the documentation to use the generated metadata

See https://github.com/lxc/incus/pull/847 for an example.

NathanChase22:
Resolves #1792 

Ran `make doc-spellcheck` and it dinged me for having "vxlan", "dns", and "openvswitch". However I assume this is okay given that these words are also found in the documentation on the site.

Signed-off-by: Nathan Chase <ntc477@utexas.edu> & Noor Ali <noorali05@utexas.edu>

### DIFF
--- a/doc/config_options.txt
+++ b/doc/config_options.txt
@@ -1600,6 +1600,432 @@ User keys can be used in search.
 ```
 
 <!-- config group network_address_set-common end -->
+<!-- config group network_bridge-common start -->
+```{config:option} bgp.ipv4.nexthop network_bridge-common
+:condition: "BGP server"
+:default: "local address"
+:shortdesc: "Override the next-hop for advertised prefixes"
+:type: "string"
+
+```
+
+```{config:option} bgp.ipv6.nexthop network_bridge-common
+:condition: "BGP server"
+:default: "local address"
+:shortdesc: "Override the next-hop for advertised prefixes"
+:type: "string"
+
+```
+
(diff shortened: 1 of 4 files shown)


NathanChase22:
Not sure if the systems test failures are due to me, but I can definitely work on resolving failures with the documentation and commit tests. 

1. Including a `Signed-off-by: ` in my commits
2.  Remove hjard tabs included in `network_bridge.md`

stgraber:
Change looks great!

I only had a few minor things to tweaked:
 - Re-titled the commits and made sure they all individually contain the correct author field and Signed-off-by line
 - Ran `gofumpt -w ./` to fix the issues reported by the Code checks
 - Went through the list of reported documentation failures. There was a case where we had "`vlan``" so the double "`" was causing an issue. There was a cased of "openvswitch" not being escaped as well as a "dns.domain" which was similarly not quoted. That accounts for the three failures we were getting.

stgraber (approved):

PR #1927, opened 2025-04-12
doc: update reliability status of Btrfs & ZFS

Integral-Tech:
Stability status of Btrfs: https://btrfs.readthedocs.io/en/latest/Status.html
The license incompatibility of ZFS can be troublesome, as kernel updates will break the kernel API that ZFS uses from time to time. This means there will be a time period where ZFS does not work on the latest mainline kernel release.

### DIFF
--- a/doc/reference/storage_drivers.md
+++ b/doc/reference/storage_drivers.md
@@ -82,7 +82,7 @@ Therefore, Incus uses `rsync` instead of the optimized transfer for refreshes wi
 ## Recommended setup
 
 The two best options for use with Incus are ZFS and Btrfs.
-They have similar functionalities, but ZFS is more reliable.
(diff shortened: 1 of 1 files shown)


stgraber:
I'm still happy with the current wording. ZFS being out of tree is unfortunate for maintenance but that's generally handled by the distributions.

From a user's point of view, when ZFS is available, it is more reliable than btrfs.
btrfs to this day still has issues with:
 - Corruption on RAID (excluding RAID 0/1)
 - Handling of VM disks (CoW must be disabled)
 - Handling of quotas (qgroups can be bypassed)

Good progress is being made on those and in some cases, those concerns may be addressed in current mainline. But something being fixed in current mainline and something being available to our primarily LTS users are two different stories.

Based on the image server statistics we're getting, the most used kernel version by Incus users is Linux 6.1, so looking at something a bit over 2 years old.

Integral-Tech:
@stgraber Thanks for your reply. Maybe when most Incus users update their kernel to Linux 6.12 (the latest LTS version), the status of Btrfs will be different.

Issue #1932, opened 2025-04-13
Support passing bearer authentication token through access_token parameter


stgraber:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

It was recently reported (https://discuss.linuxcontainers.org/t/tls2jwt-command/23458/6) that unlike basically every other programming language, javascript doesn't support passing through HTTP headers during a websocket request...

To still allow the use of our TLS signed bearer tokens with Websocket connections from a javascript client, we'd want to implement the `access_token` argument which then takes the Bearer token value directly.

This should be easy enough to detect at the same spot in the code that we currently handle the `Authorization` header.

https://self-issued.info/docs/draft-ietf-oauth-v2-bearer.html#query-param

irtaza9:
Where do we have to add the changes? Do we need to add Checkjwttoken [function](https://github.com/lxc/incus/blob/main/internal/server/util/http.go)  in this [file](https://github.com/lxc/incus/blob/main/internal/server/operations/websocket.go) or is the existing approach will handle the case?

stgraber:
I'd just modify the beginning of `CheckJwtToken` and make it so that if we can't find a bearer token in the `Authorization` header, we then go and look for an `access_token` parameter in the request, if we find that, then we continue with the function the same way we would have the bearer token from the header.

irtaza9:
http calls get successful response with authentication by using `access_token` but I tested `web-socket` and it is not connecting with the server. 

![Image](https://github.com/user-attachments/assets/5f4414fc-1987-4356-9518-cc614e66d0ae)

stgraber:
Websocket endpoints like this one which include a secret= argument do not need authentication, so you may have a problem elsewhere.

(Though passing the access token for those URLs shouldn't cause any issue)

Are you getting an actual authentication/authorization failure on the client or server side?

irtaza9:
it's a long running call which never gets timeout in case of websocket and doesn't throws any error. What is the log file path on the server which give more clarity on this?

stgraber:
`incus monitor --pretty` would report untrusted connections.

stgraber:
Just to confirm that you do not need to authenticate for those websockets:
```
stgraber@dakara:~$ incus query -X POST /1.0/instances/kernel-test/console -d '{}'
{
	"class": "websocket",
	"created_at": "2025-04-22T11:47:59.336975758-04:00",
	"description": "Showing console",
	"err": "",
[...]

PR #1940, opened 2025-04-15
feat: support access_token query parameter as JWT fallback

irtaza9:
This update enhances the `CheckJwtToken` function to support scenarios where the JWT token is provided via the `access_token` query parameter instead of the Authorization header.

🔧 Fixes: #1932 

Changes:

- Attempts to extract the token from the Authorization header first (Bearer <token>)
- Falls back to `access_token` query parameter if the header is missing or invalid
- Maintains existing JWT validation: expiration, not-before, and signature verification
- This fix provides compatibility for clients that cannot set headers, such as certain browser-based tools or embedded clients.


stgraber:
@irtaza9 commits need a `Signed-off-by` line before we can review them (see CONTRIBUTING.md)

Issue #986, opened 2024-07-12
Memory hotplug support for VMs


nanjj:
Is it possible for incus to support qemu memory hotplug feature defined [here](https://github.com/qemu/qemu/blob/master/docs/memory-hotplug.txt)? 
The usage is similar with CPU hotplug, launch qemu instance via 
```
-m [size=]megs[,slots=n,maxmem=size]
```
and change memory via QMP like:
```
 (qemu) object_add memory-backend-ram,id=mem1,size=1G
 (qemu) device_add pc-dimm,id=dimm1,memdev=mem1
 (qemu) device_del dimm1
 (qemu) object_del mem1
```
From incus source code we can see CPU hotplug feature has been supported so I am asking why QEMU memory hotplug not supported, very confusing.

And the usage is obviously valuable. 

Memory hotplug has been supported broadly by most guest OS(even the windows, almost all the versions have support for this as you may know, CPU hotplug has been supported by only windows server edition), and qemu has the support, too.

For incus we may need to add check for memory hotplug feature, give an initial slots value(maybe 2) and maxmem size(maybe 32G) , and when user set limits.memory (or maybe limits.memory.size and limits.slots) I dont know clearly, incusd using the qmp client to handle the config change.

stgraber:
Yeah, that's been something we've been meaning to add for a while, but it's also a very complex one to handle right as you need to decide on the right granularity, consider NUMA nodes, handle hugepages, ...

We already have 3-4 different code paths for memory as it stands today and all of those will need to handle DIMM hotplug. The other side of this will be to know how well the OS will handle this.

For CPU we can very easily hotplug/hotremove CPUs and the OS usually handles that pretty well. For memory, hotplug should be okay, hotremove likely to be more problematic, so we may need to use ballooning for hotremove.

For now the trick you can use is start the VM with a higher allocation than needed and the reduce limits.memory which will use the memory balloon driver to shrink things.

nanjj:
>As for `how well the OS will handle this`
I am using vwmare guest OS compatibility guide to check this， for linux memory hotplug as an example：https://www.vmware.com/resources/compatibility/search.php?deviceCategory=software&details=1&osFamily=2&virtualHardware=23&page=1&display_interval=10&sortColumn=Partner&sortOrder=Asc&testConfig=16

srd424:
> For now the trick you can use is start the VM with a higher allocation than needed and the reduce limits.memory which will use the memory balloon driver to shrink things.

[...]

PR #1945, opened 2025-04-15
Memory hotplug support for VMs

presztak:
This PR adds support for hot-plugging memory during VM runtime.
It reuses the same approach and set of functions that are used for generating QEMU CPU and memory configurations. The generated configuration is then translated into a format compatible with QMP. This approach respects both hugepages and NUMA node topology.

Closes: #986

### DIFF
--- a/internal/server/instance/drivers/qmp/commands.go
+++ b/internal/server/instance/drivers/qmp/commands.go
@@ -74,6 +74,35 @@ type CPUModel struct {
 	Flags map[string]any `json:"props"`
 }
 
+// MemoryDevice contains information about a memory device.
+type MemoryDevice struct {
+	Type string          `json:"type"`
+	Data json.RawMessage `json:"data"`
+}
+
+// PCDimmDevice contains information about a memory device of type pc-dimm.
+type PCDimmDevice struct {
+	ID           string `json:"id"`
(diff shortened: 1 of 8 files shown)


internal/server/instance/drivers/qmp/commands.go
```diff
@@ -74,6 +74,35 @@ type CPUModel struct {
...
 }
 
+// MemoryDevice contains information about a memory device.
+type MemoryDevice struct {
+	Type string          `json:"type"`
+	Data json.RawMessage `json:"data"`
+}
+
+// PcDimmDevice contains information about a memory device of type pc-dimm.
+type PcDimmDevice struct {
```
stgraber, line 84: Nit, should be capitalized as `PCIDimmDevice` I think.

stgraber:
We also need to add an API extension.

stgraber:
And it'd be nice to split the second commit in two, first half being the function split which doesn't introduce anything user visible, and then second half will be the hotplug.

internal/server/instance/drivers/qmp/commands.go
presztak: Done.

presztak:
All comments have been addressed.

PR #1955, opened 2025-04-20
Add common aliases for add/create remove/delete/rm in the CLI

### LINKED ISSUE #1797: Add common aliases for add/create remove/delete/rm in the CLI
The CLI is designed to be pretty consistent in the naming of its main actions.

Objects where user data is important will normally be created through `create` and deleted through `delete`. Delete actions usually imply data being lost.

Lighter weight objects, such as relations between objects or entries within a larger object are usually added through `add` and removed through `remove`.

We don't really want to change that behavior, but it would still be quite useful to those without a full mapping of all our commands in their mind to find what they want without having to rely on the help :)

For that reason, we should make sure that:
 - All `create` commands have a `add` alias
 - All `add` commands have a `create` alias
 - All `remove` commands have both `rm` and `delete` aliases
 - All `delete` commands have both `rm` and `remove` aliases

Aliases aren't listed in the help, so this will be basically invisible other than reducing the number of errors that our users run into.

joecwilson:
Finishes adding all aliases for add / create and remove/delete/rm in the incus CLI

closes #1797

### DIFF
--- a/cmd/incus/alias.go
+++ b/cmd/incus/alias.go
@@ -224,7 +224,7 @@ type cmdAliasRemove struct {
 func (c *cmdAliasRemove) Command() *cobra.Command {
 	cmd := &cobra.Command{}
 	cmd.Use = usage("remove", i18n.G("<alias>"))
-	cmd.Aliases = []string{"rm"}
+	cmd.Aliases = []string{"delete", "rm"}
 	cmd.Short = i18n.G("Remove aliases")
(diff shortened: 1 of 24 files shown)


joecwilson:
Marking as draft to investigate which aliases are causing issues with the tests

stgraber:
Oops, I missed that this one got out of draft. I'll look into this now.

stgraber (approved):

PR #1958, opened 2025-04-21
Remove arping dependency

ahmetfturhan:
j‑keck/arping has been archived since 2019. Although it’s been stable (no commits since May 2022), it is still the most‑maintained standalone ARP library available and removes an unmaintained dependency.


stgraber:
Thanks! I did a few minor changes:
 - Split the change from the gomod update
 - Tweaked an error message to capitalize it the way we do within the Incus codebase
 - Aligned Author and Signed-off-by values

stgraber (approved):

ahmetfturhan:
Thank you for the review! This was my first open source contribution, I will apply the error message and splitting commits things in the future :)

PR #1959, opened 2025-04-21
Replace rebfig/cron/v3 with adhocore/gronx

nanjj:


stgraber:
@gibmat heads up that this change makes sense to avoid a rather unmaintained Go package but I don't see this new Go package existing in Debian. So you'll most likely have to revert this commit when looking at update to 6.0.5, hopefully that's not going to cause too much trouble.

gibmat:
> I don't see this new Go package existing in Debian.

This is actually probably the best time to be changing Incus dependencies. :grinning:  The trixie freeze has begun, so we're not expecting any significant changes to packages for the upcoming release, and I've already got a couple of other dependencies that need to be packaged once forky development opens up to successfully build the latest feature release of Incus. I'll add this one to my list, then whenever 6.0.5 drops it shouldn't be any issue to build with the new library.

PR #1960, opened 2025-04-21
Update yaml.v2 to yaml.v3

ahmetfturhan:
Switched to yaml.v3 to enable strict field checking via KnownFields.
Updated decoding logic accordingly.

shared/util: add YamlUnmarshalStrict function with yaml.v3

Adds a utility for strict YAML unmarshalling using KnownFields from yaml.v3.
Improves configuration validation by rejecting unknown fields.


stgraber:
Thanks for the change, I did a few small adjustments:
 - Renamed the new function to capitalize YAML
 - Fixed import order in various files
 - Split commits to be per area of the Incus project
 - Aligned Signed-off-by with Author field
 - Did a full gomod update
 - Adjusted our list of imports in the testsuite

stgraber (approved):

stgraber:
Hmm, this seems to be causing a rather severe regression, going to have to track it down on my end.

stgraber:
After pulling my hair out for a few hours, I've decided that we will not be moving to go-yaml.v3.

The main reason is that go-yaml.v3 doesn't provide compatible output with go-yaml.v2 and also decided to start indenting everything with four spaces. Given our sometimes rather deep structures and the fact that we show YAML to our users, that becomes quite annoying.

As go-yaml is now unmaintained, we're likely to look at moving away from it in the next year or so. So rather than migrating to go-yaml.v3, breaking a bunch of our users in the process and then do it again when we move to something else, let's just skip this migration altogether and deal with an hopefully more reasonable yaml package once the dust has settled and there's a clear winner from the go-yaml deprecation.

ahmetfturhan:
Thank you for the feedback! Thanks for taking the time to investigate all the incompatibilities in go‑yaml.v3. It totally makes sense to skip this migration and avoid piling on more breaking changes for now. I honestly didn't noticed the four space indentation part :) Glad you caught it!

stgraber:
It was causing a bunch of our tests to fail :)

Then when working on fixing the tests I looked at some of my more complex instances and it certainly looked a lot worse. Apparently the YAML standard recommends two spaces indent, no idea why go-yaml.v3 decided to go to 4...

ahmetfturhan:
I saw Kustomize had the same issue when trying to upgrade to v3. 
https://github.com/kubernetes-sigs/kustomize/issues/3946

And just saw this change article; 
https://ubuntu.com/blog/api-v3-of-the-yaml-package-for-go-is-available

This article says that we can customize the custom indentation, but since it's still unmaintained, I think no reason for upgrading to v3 --and cause trouble for bunch of users :)

image

stgraber:
The customization must be done by using SetIndent on an encoder, so every single Marshal can would need to be changed, that didn't seem like fun to do :)

PR #1975, opened 2025-04-23
Replace yaml.v2 with goccy/go-yaml

nanjj:
`github.com/goccy/go-yaml` is active, stable, fast and no additional dependencies on others except go std.

### DIFF
--- a/cmd/incus-agent/templates.go
+++ b/cmd/incus-agent/templates.go
@@ -8,7 +8,7 @@ import (
 	"path/filepath"
 	"strconv"
 
-	"gopkg.in/yaml.v2"
+	"github.com/goccy/go-yaml"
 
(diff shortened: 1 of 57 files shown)


stgraber:
As mentioned in the other issue, we're going to be waiting a few months before we make any move and are likely to start doing it on other less critical projects first before doing Incus itself.

This isn't a trivial low risk change so we want to do it carefully.

nanjj:
> As mentioned in the other issue, we're going to be waiting a few months before we make any move and are likely to start doing it on other less critical projects first before doing Incus itself.
> 
> This isn't a trivial low risk change so we want to do it carefully.

Agree. After the risk handled in https://github.com/lxc/incus/pull/1984, it will be a low risk change.

PR #1989, opened 2025-04-25
Add server side filtering for `incus network list`

### LINKED ISSUE #1909: Use server-side filtering for `incus network list`

We have recently added server-side record filtering on all API endpoints.
We're now looking to roll that out across all commands in the client with a first batch of those being prioritized now.

In general the expected work is:
 - Add `WithFilter` and `WithFilterAllProjects` functions to the client (`client/`)
 - Change the CLI List command to use those functions and provide a filter.

We currently have some examples available through:
 - `incus list`
 - `incus image list`
 - `incus storage volume list`

_No response_

Abdomash:
Resolves #1909

Adds server-side filtering for the `incus network list` command.

Filters are either:
- **a string**, which filters the names of networks
(e.g. `incus network list foo` lists all network names that start with `foo`)
- **a list of  `<key>=<value>` pairs** delimited by a comma `,`, which filters by network properties.
(e,g, `incus network list type=bridge` lists all networks of the type `bridge`)


stgraber:
Does config based filtering also work?

For example `incus network list ipv4.address=10.44.241.1/24` or `incus network list ipv4.nat=true`?

Abdomash:
> Does config based filtering also work?
> 
> For example `incus network list ipv4.address=10.44.241.1/24` or `incus network list ipv4.nat=true`?

No that doesn't work unfortunately. Would like me to add config-based filtering to this PR or as a separate PR?

stgraber:
Would be good to have in here, basically aligning the behavior with what we have in `incus list`

Abdomash:
Sounds good 👍, working on it

Abdomash:
Added config based filtering

`incus network list ipv4.address=10.44.241.1/24` and `incus network list ipv4.nat=true` should work now.

stgraber:
Excellent, looks great!

Issue #1978, opened 2025-04-23
Use netlink directly instead of iproute2


gwenya:
We are currently calling the `ip` command for configuring addresses, routes and neighbour objects. 
This is unnecessary additional overhead and requires parsing the output in some cases, like here: https://github.com/lxc/incus/blob/faa488098f0d4ae41c3480b7e75473b51368a336/internal/server/ip/neigh.go#L61-L87

By using netlink directly we can avoid that.
We already have a dependency on [github.com/vishvananda/netlink](https://github.com/vishvananda/netlink), so we could use that without adding any extra dependencies. 

If there is interest in this I will make a PR.

stgraber:
Absolutely, it was actually the whole goal of the `ip` to separate all that logic out of random spots in the codebase ahead of moving them to direct netlink calls, so I'd definitely be happy to see us use netlink directly.

gwenya:
@stgraber would you prefer one big PR for everything or smaller ones?

I'm also going to rework the API a bit, I can keep that separate as well to make code review easier.

Currently we're using strings for almost everything, which have to be parsed again to pass them to netlink, and I think there's something weird with the way the `tc` things are modeled, I'm gonna have to figure out what any of this even does.

stgraber:
I'm fine either way. We could convert it bit by bit if that's easier or do it in one chunk if that's easier because of inter-dependencies.

Probably just try to split things at least on the commit front, basically one commit to convert to netlink, one commit to update whatever used it, or something like that.

PR #1990, opened 2025-04-25
Use netlink instead of calling iproute2 commands

gwenya:
See #1978 

This is still WIP, the PR at this time is mainly for running the CI. Feedback is very welcome though, especially on the non-trivial TODOs.

### DIFF
--- a/internal/server/device/device_utils_network.go
+++ b/internal/server/device/device_utils_network.go
@@ -216,13 +216,13 @@ func networkCreateVethPair(hostName string, m deviceConfig.Device) (string, uint
 
 	veth := &ip.Veth{
 		Link: ip.Link{
-			Name: hostName,
-			Up:   true,
+			Name:   hostName,
+			Up:     true,
+			Master: m["vrf"],
 		},
 		Peer: ip.Link{
 			Name: network.RandomDevName("veth"),
 		},
-		Master: m["vrf"],
 	}
 
 	// Set the MTU on both ends.
@@ -690,7 +690,7 @@ func bgpRemovePrefix(d *deviceCommon, config map[string]string) error {
 	return nil
 }
 
-// networkSRIOVParentVFInfo returns info about an SR-IOV virtual function from the parent NIC using the ip tool.
+// networkSRIOVParentVFInfo returns info about an SR-IOV virtual function from the parent NIC.
 func networkSRIOVParentVFInfo(vfParent string, vfID int) (ip.VirtFuncInfo, error) {
 	link := &ip.Link{Name: vfParent}
 	vfi, err := link.GetVFInfo(vfID)
@@ -716,9 +716,9 @@ func networkSRIOVSetupVF(d deviceCommon, vfParent string, vfDevice string, vfID 
 
 	// Record properties of VF settings on the parent device.
 	volatile["last_state.vf.parent"] = vfParent
-	volatile["last_state.vf.hwaddr"] = vfInfo.Address
+	volatile["last_state.vf.hwaddr"] = vfInfo.Address.String()
 	volatile["last_state.vf.id"] = fmt.Sprintf("%d", vfID)
-	volatile["last_state.vf.vlan"] = fmt.Sprintf("%d", vfInfo.VLANs[0]["vlan"])
+	volatile["last_state.vf.vlan"] = fmt.Sprintf("%d", vfInfo.VLAN)
 	volatile["last_state.vf.spoofcheck"] = fmt.Sprintf("%t", vfInfo.SpoofCheck)
 
 	// Record the host interface we represents the VF device which we will move into instance.
@@ -772,7 +772,7 @@ func networkSRIOVSetupVF(d deviceCommon, vfParent string, vfDevice string, vfID 
 		}
 
 		// Now that MAC is set on VF, we can enable spoof checking.
-		err = link.SetVfSpoofchk(volatile["last_state.vf.id"], "on")
+		err = link.SetVfSpoofchk(volatile["last_state.vf.id"], true)
 		if err != nil {
 			return vfPCIDev, 0, fmt.Errorf("Failed enabling spoof check for VF %q: %w", volatile["last_state.vf.id"], err)
 		}
@@ -784,7 +784,7 @@ func networkSRIOVSetupVF(d deviceCommon, vfParent string, vfDevice string, vfID 
 		_ = link.SetVfAddress(volatile["last_state.vf.id"], "00:00:00:00:00:00")
 
 		// Ensure spoof checking is disabled if not enabled in instance (only for real VF).
-		err = link.SetVfSpoofchk(volatile["last_state.vf.id"], "off")
+		err = link.SetVfSpoofchk(volatile["last_state.vf.id"], false)
 		if err != nil && d.config["security.mac_filtering"] != "" {
 			return vfPCIDev, 0, fmt.Errorf("Failed disabling spoof check for VF %q: %w", volatile["last_state.vf.id"], err)
 		}
@@ -897,11 +897,7 @@ func networkSRIOVRestoreVF(d deviceCommon, useSpoofCheck bool, volatile map[stri
 	// Reset VF MAC spoofing protection if recorded. Do this first before resetting the MAC
 	// to avoid any issues with zero MACs refusing to be set whilst spoof check is on.
 	if volatile["last_state.vf.spoofcheck"] != "" {
-		mode := "off"
-		if util.IsTrue(volatile["last_state.vf.spoofcheck"]) {
-			mode = "on"
-		}
-
+		mode := util.IsTrue(volatile["last_state.vf.spoofcheck"])
 		link := &ip.Link{Name: parent}
 		err := link.SetVfSpoofchk(volatile["last_state.vf.id"], mode)
 		if err != nil && d.config["security.mac_filtering"] != "" {
--- a/internal/server/device/device_utils_network.go
+++ b/internal/server/device/device_utils_network.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/mdlayher/arp"
 	"github.com/mdlayher/ndp"
+	"golang.org/x/sys/unix"
 
 	deviceConfig "github.com/lxc/incus/v6/internal/server/device/config"
 	pcidev "github.com/lxc/incus/v6/internal/server/device/pci"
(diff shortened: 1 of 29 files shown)


gwenya:
@stgraber we currently don't have any tests specifically for this package, if any of the functions are tested then only incidentally.
Unit tests don't really work since this interacts with the system and there's no point in mocking that. 
I'm thinking maybe a test program that can run any of these functions via reflection, and then a test script that calls said program and uses iproute2 tools to check if it does things correctly. 
And ideally I'd like to be able to run these tests locally without messing up my network setup, so in a VM maybe. I'm not sure if that fits into our current testing system somehow?
Some of these might not even be possible to test in the CI, for example vrf is a kernel module that may or may not be loaded.

stgraber:
Yeah, it's not the kind of code that's particularly directly testable, it makes more sense to have it be indirectly tested through or normal system tests and the daily test runs we have on Jenkins too.

gwenya:
Hmh can we generate coverage data from the existing tests to see if they hit all the code paths in this package?

stgraber:
Hmm, not particularly easily as we typically run our tests across a LOT of `incusd` instances, so can't centrally keep track of everything that was hit. And some of the more complex bits (SR-IOV and the like) are tested externally, some manually, some automatically through Jenkins.

gwenya:
Hmh I see
The tests already caught some bugs that I'm fixing now but without knowing if everything is tested it's hard having confidence in the end result

stgraber:
@gwenya yeah, the best we can do is land this one early in the release to maximize the time we have for all our various tests to catch issues.

internal/server/device/device_utils_network.go
```diff
@@ -528,14 +530,21 @@ func networkSetupHostVethLimits(d *deviceCommon, oldConfig deviceConfig.Device,
 	}
 
 	// Clean any existing entry
-	qdisc := &ip.Qdisc{Dev: veth, Root: true}
-	_ = qdisc.Delete()
-	qdisc = &ip.Qdisc{Dev: veth, Ingress: true}
-	_ = qdisc.Delete()
+	qdiscIngress := &ip.QdiscIngress{Qdisc: ip.Qdisc{Dev: veth, Handle: "ffff:0"}}
```
gwenya, line 533: @stgraber should I put these kinds of changes into their own commit or leave them together with the corresponding change in the ip package?

internal/server/ip/link.go
```diff
@@ -95,415 +84,344 @@ func (l *Link) add(linkType string, additionalArgs []string) error {
...
 	}
 
-	// Search VFs returned for match.
-	found := false
-	for _, vfInfo := range ifInfo[0].VFList {
-		if vfInfo.VF == vfID {
-			vf = vfInfo // Found a match.
-			found = true
+	for _, vf := range link.Attrs().Vfs {
+		if vf.ID == vfID {
```
gwenya, line 360: I am not confident about this ID being the number we want.

In the iproute2 code the vf is taken from `IFLA_VF_MAC`, and that's where the netlink library puts it when setting it, but this `ID` seems to be just the index in the list of vfs returned from the kernel. 

I don't have any SR-IOV capable devices to try out if this is the same. If not then we might have to request a change on the netlink library to expose the correct value.

gwenya:
The vlan link type is blocked on https://github.com/vishvananda/netlink/pull/1078, since the netlink library currently does not support setting the `gvrp` flag on vlan links.

internal/server/device/device_utils_network.go
stgraber: it's fine to leave them with the change I think

internal/server/ip/link.go
stgraber: You may want to request access to https://discuss.linuxcontainers.org/t/opening-up-the-linux-containers-lab-environment/22588

The lab has quite a few SR-IOV capable systems:
 - lantea (dual-port ConnectX-4, single-port Intel 10Gbit)
 - asuras (dual-port ConnectX-6)
 - entak (dual-port ConnectX-6)
 - velona (dual-port ConnectX-6)

stgraber: Of those at least lantea is known to be correctly configured at the firmware level as it does run SR-IOV daily tests for us.

gwenya: Finally managed to get around to this, it appears that the IDs and indices are in fact the same so the code should be fine as is.

gwenya:
@stgraber can you check if there's something wrong with the cluster tests at the moment or if my changes broke something?

stgraber:
We've seen cluster failures on other PRs too, no idea what changed.

stgraber:
@gwenya since you cleared the draft status, I'm assuming this is now good to go?

internal/server/ip/addr.go
```diff
@@ -1,43 +1,95 @@
...
+		return fmt.Errorf("failed to get addresses for device %s: %w", a.DevName, err)
 	}
 
-	_, err := subprocess.RunCommand("ip", cmd...)
+	scope, err := a.scopeNum()
 	if err != nil {
 		return err
 	}
 
+	// TODO: iproute2 supposedly does some kind of batch delete in multiple rounds, which is probably more performant than deleting addresses one by one, but the iproute2 code is arcane and uncommented and I haven't yet figured out how it works
```
stgraber: Maybe update to something like:

```
// NOTE: If this becomes a bottleneck, there appears to be support for batching those kind of changes within netlink.
```

internal/server/ip/filter.go
```diff
@@ -50,33 +47,80 @@ type Filter struct {
...
 	Filter
-	Value   string
-	Mask    string
+	Value   uint32
+	Mask    uint32
 	Actions []Action
 }
 
+func parseProtocol(proto string) (uint16, error) {
+	// TODO: add other proto values
```
stgraber: I think it's fine to drop the TODO since it seems to cover what we need for now and a clear error will be returned in other cases, so not really something needed to do in this PR.

internal/server/ip/link_veth.go
```diff
@@ -1,29 +1,38 @@
...
 	Link
 	Peer   Link
-	Master string
+	Master string // TODO: Link already has a master, and the way this one is used does not suggest that it is different
 }
 
 // Add adds new virtual link.
 func (veth *Veth) Add() error {
-	err := veth.Link.add("veth", append([]string{"peer"}, veth.Peer.args()...))
+	veth.Link.Master = veth.Master // TODO: this is just to see if it works
```
stgraber: Did that work? If so, then we could remove the Master from Veth and clear both TODO?

internal/server/ip/link_vxlan.go
```diff
@@ -1,49 +1,84 @@
...
-		args = append(args, "dev", vxlan.DevName)
+		dev, err := linkByName(vxlan.DevName)
+		if err != nil {
+			return err
+		}
+
+		devIndex = dev.Attrs().Index
 	}
 
+	// TODO: all of these these can be passed net.IP
```
stgraber: What's left to do here?

internal/server/ip/tuntap.go
```diff
@@ -14,21 +16,35 @@ type Tuntap struct {
...
+	switch t.Mode {
+	case "tun":
+		mode = netlink.TUNTAP_MODE_TUN
+	case "tap":
+		mode = netlink.TUNTAP_MODE_TAP
+	default:
+		return fmt.Errorf("invalid tuntap mode %q", t.Mode)
+	}
+
+	// TODO: there is TUNTAP_DEFAULTS and TUNTAP_MULTI_QUEUE_DEFAULTS in the netlink package, I don't know if these should be used
```
stgraber: Yeah, looking at the definitions and kernel doc, I think it'd make sense to set `flags` to `TUNTAP_DEFAULTS` and then if MultiQueue is requested to set it to `TUNTAP_MULTI_QUEUE_DEFAULTS` instead.

stgraber:
@gwenya this work looks great, thanks!

On top of the comments above addressing the TODO left in this PR, the only other thing that caught my eye are the error messages. Incus' pattern for error messages is to start with a capital letter.

(Yes, this goes counter to Go recommendations but we've generally found this to be more readable and we're trying to keep things consistent. We've been testing Go's recommendation in some other projects like Incus OS and may roll that out to Incus eventually if found to work well enough)

stgraber:
@gwenya ping

gwenya:
Oh sorry I completely missed your review, thanks for pinging me!
I’ll take a closer look at it later today.

internal/server/ip/link_veth.go
gwenya: I'll look into that again, I don't actually remember.

internal/server/ip/link_vxlan.go
gwenya: I was unsure if we want to pass IPs as strings or as net.IP (or net.IPNet for subnets), and intended to bring that up but forgot about it. 
In favour of using the net types, is more type safety and better API design as well as avoiding round-trips via strings in cases where we already have a net.IP or net.IPNet; in favour of strings is not changing the existing API so much and not having to parse config strings to IP addresses in various places. Also when parsing to net.IPNet, care needs to be taken if the host portion is not zero.

gwenya:
I just noticed that we parse the routing table and various other network information from procfs in various places like here: https://github.com/lxc/incus/blob/28452f93d607f07bca0aa1c01de50bc4e2ec5b13/internal/server/network/network_utils.go#L650

Should we see if we can use the ip package there as well? Might be best in a separate PR though.

internal/server/ip/addr.go
gwenya: Another thing I just remembered regarding flushing, the `ip` tool does the flushing in multiple rounds, which I think is tied into the batching to some degree but also has the effect of deleting addresses that are created while the flush command runs, giving up after 10 rounds to avoid getting stuck forever if addresses keep getting added. I don't know if this is an issue we might have, but if we do then we can just wrap most of the function in a for loop that runs 10 times or so and breaks if no addresses exist anymore.

stgraber:
> I just noticed that we parse the routing table and various other network information from procfs in various places like here: https://github.com/lxc/incus/blob/28452f93d607f07bca0aa1c01de50bc4e2ec5b13/internal/server/network/network_utils.go#L650
> 
> Should we see if we can use the ip package there as well? Might be best in a separate PR though.

Yeah, that'd make sense to move that over to the ip package and see if there's a more optimized way to get that out of netlink.

internal/server/ip/link_vxlan.go
stgraber: I'd be in favor of using net.IP/net.IPNet for this.

stgraber:
@gwenya branch looks good to me.

Do you want to do the net.IP/net/IPNet switch over and remove that last TODO before we merge this one?

gwenya:
@stgraber will do!

gwenya:
@stgraber I did the change to using net.IP and net.IPNet, both in the file that had the TODO and in the other places as well. Let me know if it looks good and I'll rebase it into the existing commits.

stgraber:
Remaining issue to track down:
```
2025-06-22T23:25:22.0304357Z time="2025-06-22T23:25:22Z" level=error msg="Failed to restore route" driver=bridge err="Failed to replace route {Ifindex: 3 Dst: 0.0.0.0/0 Src: <nil> Gw: 100.64.1.1 Flags: [] Table: 254 Realm: 0}: network is unreachable: Nexthop has invalid gateway" network=inc27523x project=default
```

That hits in the cluster tests. It's non-fatal as it's a non-fatal backup/restore of routes, but the failure is unexpected.

gwenya:
I added the scope to our route struct so we can add it back with the same scope, I think that should fix it

stgraber:
Sounds good. I'm planning on doing another quick review/rebase pass on this one and then merging it shortly after 6.14 is out. That will give us a full dev cycle to catch any extra regressions.

stgraber (approved):

stgraber:
@gwenya thanks for the great work!

stgraber:
@gwenya FYI, I'm seeing a couple of potentially related CI failures in Jenkins today.
One is a network I/O limit validation which suggests that the limits are no longer being applied for some reason.

Another is a macvlan related network test which is now failing to give us an IPv6 address.


I'll look at these two a bit closer this week. It shouldn't be too difficult to figure out what's going on and fix it and that's exactly why I wanted this merged right after a stable release so we can catch those weird edge cases ;)

gwenya:
@stgraber At a quick glance, I think I'm not setting the all multicast flag properly, could that be the cause for the IPv6 issue?

stgraber:
Quite likely. IPv6 actively relies on multicast for router advertisements.

We also got this is failure that showed up a few hours ago: https://jenkins.linuxcontainers.org/job/incus-test-usb/651/console

gwenya:
Ah yeah, we should probably ignore that error in Flush.

Should I push fixes for these things in here or in a new PR?

stgraber:
In a new PR since we've merged this one. Can still come out of the same branch though.

PR #1992, opened 2025-04-25
doc: Fix default value of ipv6.routes network_bridge

hnakamur:
It was changed from "IPv6 address" to "IPv6"
in
https://github.com/lxc/incus/commit/70897e34e459f53e3ef1b189cc84706546e9aca1#diff-ea939356d9c610505874f52d8134c991919241ef130a0d667572db54ce8ded46L97 and
https://github.com/lxc/incus/commit/47599d89620f930d7423f0a2252651504ad7e969#diff-5f4225b439a90e9d63cf9e7463cb3c8ae433e26a64e277494cbc124c97dee6daR1893

I think it was a mistake.

### DIFF
--- a/doc/config_options.txt
+++ b/doc/config_options.txt
@@ -3104,7 +3104,7 @@ User keys can be used in search.
 ```
 
 ```{config:option} ipv6.routes network_bridge-common
-:condition: "IPv6"
+:condition: "IPv6 address"
 :default: "-"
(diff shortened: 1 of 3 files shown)


stgraber:
Thanks!

We technically need to go fix the code rather than the .txt file and then regenerate the .txt from that, but I'll take care of it.

In this case, it would have been in `internal/server/network/driver_bridge.go` and then running `make update-metadata`  would have refreshed the .txt and .yaml.

stgraber (approved):

PR #2000, opened 2025-04-27
rename reverters from revert to reverter to not conflict with the package

The127:
A lot of the reverters were named `revert` which conflicts with the package name, while others were called `reverter`. I decided to unify the naming to `reverter` so that they are all named the same and do not conflict with the package name anymore.

I dont know how to name the commit, since it covers a lot of code parts.


The127:
the failing linstor ubuntu test looks like a flake to me

stgraber:
Yeah, that's fine, don't worry about it.

stgraber (approved):

PR #2005, opened 2025-04-27
incusd: Refactor getting heartbeat mode name into a function

The127:
Refactors getting the heartbeat mode name into a function. The original code did not cover all cases.


The127:
That failing test seems like a flake (i only touched `heartbeatMode`).
However idk if it means that there is a bug somewhere in the relevant code of that failing test.

stgraber (approved):

PR #1964, opened 2025-04-22
incus/internal/server/device/nic_routed.go: Added host_tables

AbhinavTiruvee:
Working on specifying which tables routes can go, would appreciate any tips/pointers.

(https://github.com/lxc/incus/issues/1559)

Issue #1559, opened 2025-01-05
Add option to disable entry in main routing table if `host_table` is set for routed NIC


ibot3:
For routed NICs, the `host_table` parameter can be specified to create a route in the specified routing table.
However, the same route is always installed in the main routing table too:

https://github.com/lxc/incus/blob/fc4d0ca3527545991d4f83dd2682b0b81509c6f5/internal/server/device/nic_routed.go#L410-L416

It would be great if this behavior could be changed by another config option so that the route is only present in the specified routing table.

**Our scenario:**
We want to have VMs in different VRFs on the incus host.
But for routed NICs to work, ip_forwarding must be enabled on the host.
This causes that from the main VRF, all VMs with routed NICs can be reached, what we don't want, because they should only be reachable within their VRF.

stgraber:
Hmm, I don't recall how we settled on this behavior for `host_table`...

In hindsight, we'd have been better served by a `host_tables` config key taking a comma separated list of table names and not having that default behavior of always putting in the main table too.

I think the best path forward here is to implement `ipv4.host_tables` and `ipv6.host_tables` defined as a comma separated list of table IDs and whose default is `254`. Making them conflict with `ipv4.host_table` and `ipv6.host_table` respectively and then mark the old keys as deprecated in documentation.

AbhinavTiruvee:
@stgraber Hi, I'm part of a group at UT Austin taking a class on virtualization. Could my partner and I take a look at this issue and try to solve it?

stgraber:
Assigned it to you. I'll be able to assign your partner once they also comment in here.

stgraber:
So for this one, we'll want to implement the behavior I described in the comment above (https://github.com/lxc/incus/issues/1559#issuecomment-2571681699)

avipatel805:
Hello, I am the partner mentioned above. Could you please add me to the issue as well?

avipatel805:
Hello,

We wanted to check if our current approach for implementing host_tables support looks good before moving on to testing and documentation. Here's what we've done so far:

1. Field Addition and Validation

- Added ipv4.host_tables and ipv6.host_tables to optional_fields.

- Added a validator to ensure list items are within the 0–255 range.

2. Table Retrieval Logic

- Created a get_tables function:

- If host_tables is specified, it returns those values.

- If the (now deprecated) host_table is specified, it returns that value.
   → Should we also return the default value 254 in this case, or only what's provided?

- If neither is provided, it defaults to 254.

[...]

PR #2009, opened 2025-04-27
incusd/server/device/nic_routed: Added host_tables

AbhinavTiruvee:
We closed our old PR (https://github.com/lxc/incus/pull/1964) and are opening a new one due to it being pretty outdated compared to the codebase. We added our changes to nic_routed.go and wanted to know where else we need to edit. We assume the documentation we need to add would be through gendoc and we would mark the ipv4.host_table and ipv6.host_table as deprecated there.

Issue #1559

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2795,3 +2795,7 @@ This is limited to bridged networks as OVN doesn't support flexible enough SNAT 
 ## `memory_hotplug`
 
 This adds memory hotplugging for VMs, allowing them to add memory at runtime without rebooting.
+
+## `instance_nic_routed_host_tables`
+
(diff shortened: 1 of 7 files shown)


stgraber:
A few small changes seem needed:
 - I'd split the addition of the new logic with marking the old options as deprecated
 - The commit with the deprecation notice should also change the doc string (gendoc)
 - You need to add a commit (typically titled `doc: Update configs`) with the result of `make update-metadata`
 - As this adds new config keys, an API extension should be added, this is done by editing both `doc/api-extensions.md` and `internal/version/api.go`, in this case, the string may be something like `instance_nic_routed_host_tables`

stgraber (approved):

stgraber (approved):

PR #2011, opened 2025-04-28
Rewrite legacy QEMU config code

bensmrs:


bensmrs:
Now that I’m using maps, `memoryConfigSectionToMap` iterates in an arbitrary order the `host-nodes.X` keys, making the test fail sometimes. Is the order important here?

stgraber:
I wouldn't expect the order to be meaningful, no

PR #2014, opened 2025-04-28
Add support for running the Incus agent on Windows

stgraber:
This is the initial support for running an Incus agent on Windows.

Because there is currently no signed VirtIO vsock drivers for Windows, nor VirtIO vsock support for Go on Windows, the agent is currently being accessed over HTTPS.

This will only work in environments where Incus is aware of the IP address(es) of the Windows system and where the Incus host(s) can connect to that address on port tcp/8443.

We'll keep an eye on the ongoing work on the VirtIO vsock driver for Windows and once a signed build is readily available, will push to get support within Go so we can move away from using HTTPS for this and use the same transport regardless of guest operating systems.

As far as what's supported at this stage:
 - `incus exec`, though not in fully interactive mode (you must pass `-T`)
 - `incus file` operations
 - `incus info` to get guest OS details

At this point, system metrics are being reported by QEMU rather than the agent.

This pull request also adds extra Github build steps to produce the needed agent binary on both Linux and Windows.


stgraber:
With this change merged, any Incus package shipping with the static Windows agent (`incus-agent.windows.x86_64`) will be able to provide a suitable agent drive to Windows.

`incus config device add win11 agent disk source=agent:config` will generate and attach such a drive.

On the Windows side, simply run `incus-agent.exe` from that drive, either directly or by manually adding it in as a service.

PR #2016, opened 2025-04-28
incusd: remove unreachable code in api internal

The127:
This PR removes unrechable code in cmd/incud/api_internal.go and also updates an outdated comment .

### DIFF
--- a/cmd/incusd/api_internal.go
+++ b/cmd/incusd/api_internal.go
@@ -743,7 +743,7 @@ func internalImportFromBackup(ctx context.Context, s *state.State, projectName s
 		return err
 	}
 
-	if allowNameOverride && instName != "" {
+	if allowNameOverride {
 		backupConf.Container.Name = instName
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/incusd/api_internal.go
```diff
@@ -907,29 +907,11 @@ func internalImportFromBackup(ctx context.Context, s *state.State, projectName s
 			return err
 		}
 
-		// If a storage volume entry exists only proceed if force was specified.
+		// If the storage volume entry does already exist we error here
 		if dbVolume != nil {
 			return fmt.Errorf(`Storage volume for snapshot %q already exists in the database`, snapInstName)
 		}
 
-		if snapErr == nil {
```
The127, line 915: The code is currently unreachable, thats why i removed it. But i am unsure if this is a bug and instead of removing it it needs to be reworked.
If it is a bug then the last one or two commits can be dropped (depending on desired behaviour).

stgraber: I think it's fine to drop that logic. Looks like we used to just delete the snapshot silently, but that probably caused issues and we're now erroring out instead.

stgraber (approved):

PR #2021, opened 2025-04-28
incus/internal/server/nic_sriov: add hwspoofchk to disable hardware spoofcheck support

fstern:
I've added a `hwspoofchk` flag to the SRIOV nic device, because not all NICs support hardware spoof checking. 

In my case I'm running this on Bluefield 3 VFs which only return "Operation not supported" when using the `spoofchk` flag. 

This makes the option in `networkSRIOVSetupVF` configurable through a setting in the profile. 

The default is enabled, to keep with existing setups. 

Signed-off-by: Falk Stern <falk@fourecks.de>

### DIFF
--- a/internal/server/device/nic.go
+++ b/internal/server/device/nic.go
@@ -33,6 +33,7 @@ func nicValidationRules(requiredFields []string, optionalFields []string, instCo
 		"security.ipv4_filtering":              validate.IsAny,
 		"security.ipv6_filtering":              validate.IsAny,
 		"security.port_isolation":              validate.Optional(validate.IsBool),
+		"hwspoofchk":				validate.Optional(validate.IsBool),
 		"ipv4.address":                         validate.Optional(validate.IsNetworkAddressV4),
 		"ipv6.address":                         validate.Optional(validate.IsNetworkAddressV6),
 		"ipv4.routes":                          validate.Optional(validate.IsListOf(validate.IsNetworkV4)),
--- a/internal/server/device/nic_sriov.go
+++ b/internal/server/device/nic_sriov.go
@@ -97,6 +97,15 @@ func (d *nicSRIOV) validateConfig(instConf instance.ConfigReader) error {
 		//  shortdesc: Prevent the instance from spoofing another instance's MAC address
 		"security.mac_filtering",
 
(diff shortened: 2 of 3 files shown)


stgraber:
I think it'd be better to introduce `security.mac_filtering` to the `sriov` NIC type and it'd be fine for it to default to off, same way we do for every other NIC type.

I'm actually unsure why we decided to enforce it on sriov NICs by default when we don't on the SRIOV accelerated OVN NICs.

fstern:
The `sriov` nics already utilize `security.mac_filtering` in e.g. https://github.com/lxc/incus/blob/3540539ab5fa1ed3dc84a3f8112183d8d1ba28e4/internal/server/device/device_utils_network.go#L764

The problem here is, that even when it's off, the code wants to explicitly disable the feature with `ip link set dev enp5s0 vf 0 spoofchk off`.

Even disabling hardware spoofchecking is not supported with some cards, the process returns a non-zero exit code and the VM or container doesn't start. 

The current code path already uses a `useSpoofCheck` flag, which just defaults to `true` in the function call of `networkSRIOVSetupVF`. This patch just makes it configurable.

stgraber:
Oops, sorry, I was reading the wrong sriov documentation on our end...

Right, so since `security.mac_filtering` is already a config option, we should make it so if it's enabled, we require the spoofcheck to be turned on, which would then fail on cards like yours. If it's not enabled, then we should attempt to turn off the spoofcheck, but not fail in the event that the operation fails (as we're now aware that some cards do not have that knob).

stgraber:
That would provide the expected behavior without having to offer another low level knob that we'd then need to be careful about when to allow for restricted environments.

fstern:
Yes, https://github.com/lxc/incus/blob/3540539ab5fa1ed3dc84a3f8112183d8d1ba28e4/internal/server/device/device_utils_network.go#L794 could fail silently, but that already has that knob. 

I don't feel comfortable enough to change that behaviour there, so I just made the `useSpoofCheck` knob accessible through config. 

You've already run the tests - anything you want me to change in the commit?

stgraber:
I'll send my own PR which will relax the check when `security.mac_filtering` isn't specifically on.

fstern:
Great, thank you!

Issue #1877, opened 2025-04-01
Extend `incus-migrate` to also support uploading filesystems and disks as custom volumes


stgraber:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

Currently `incus-migrate` can be used to create a remote instance on Incus based on either a filesystem tree (container) or a disk or image file (VM).

We should also allow for streaming into a custom volume instead, with the same logic, so a filesystem tree would land into a filesystem storage volume and a disk or image file would end up in a block storage volume.

The APIs are near identical for instances and custom volumes, so this should be pretty easy to add to the tool.

DinglyCoder:
Hi, @parthgupta2005 and I are students at the University of Texas at Austin. As part of our virtualization class, we would like to contribute to LXC Incus and resolve this issue.

stgraber:
Assigned it to you now. I'll need @parthgupta2005 to comment in here so I can assign them too.

parthgupta2005:
Hello, apologies for the late response. We look forward to resolving the issue!

parthgupta2005:
Hi, @stgraber we mistakenly committed to this issue without discussing with our professor first. Could you please unassign us from this issue. Our sincerest apologies.

stgraber:
Done!

PR #2022, opened 2025-04-28
Extend incus-migrate to support uploading filesystems and disks as custom volumes

presztak:
Closes: #1877

### DIFF
--- a/cmd/incus-migrate/main_migrate.go
+++ b/cmd/incus-migrate/main_migrate.go
@@ -64,6 +64,7 @@ type cmdMigrateData struct {
 	SourceFormat     string
 	Mounts           []string
 	InstanceArgs     api.InstancesPost
+	CustomVolumeArgs api.StorageVolumesPost
 	Pool             string
 	Project          string
 }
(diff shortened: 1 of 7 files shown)


cmd/incus-migrate/main_migrate.go
```diff
@@ -659,8 +840,10 @@ func (c *cmdMigrate) run(_ *cobra.Command, _ []string) error {
 What would you like to create?
 1) Container
 2) Virtual Machine
+3) Custom Volume (from filesystem)
+4) Custom Volume (from disk)
```
stgraber, line 844: @presztak I'll merge the PR because the logic is fine, but in a follow-up we should be able to turn that just into a single `Custom volume` option and detect whether it's a filesystem or a disk based on what the user provides us.

If it's a directory, it's a filesystem, if not, it's a disk.

stgraber:
@presztak I've dropped the API extension as there is no actual API change in this.
We're adding new functions to the `client` package, but the API itself remains identical.

stgraber (approved):

PR #2023, opened 2025-04-28
incus: refactor admin_init.go config initialization

The127:
Refactors the initialization code for the config in admin_init.go. This switch case based approach represents the intention of the if chain, since the cases are exclusive.


The127:
im gonna add some tests for this

The127:
the failing test looks like a flake to me at least

stgraber (approved):

PR #2024, opened 2025-04-28
incusd/network/ovn: Wait up to 10s for OVN northd to allocate an IP

stgraber:
In some busy environments, the 2.5s timeout wasn't quite enough.

Sponsored-by: Luizalabs (https://luizalabs.com)


diegofernandes:
@stgraber , I was thinking may we should have a config for the sleep time between retries.

hallyn (approved):

PR #2025, opened 2025-04-28
feat: support NFS storage pools

### LINKED ISSUE #1311: Add a `nfs` storage driver
`nfs` would be a small shim storage driver on top of our `dir` driver that would handle mounting the NFS share as needed on all servers within the cluster.

This should be restricted to only custom volumes and only filesystem type volumes for those as storing block/loop files on top of NFS seems like a bad idea and most NFS servers aren't quite POSIX enough to properly handle full containers (locking is always a bit weird).

Foxboron:
WIP draft PR for the feature. So far it works, a bit slow on my machine due to the id remapping which should probably be investigated.

- [x] Investigate the ID remapping
- [x] I'm unsure about the different migration options
- [x] Only support `vers=4.2`, should be fine?
- [x] Error out on missing `source`
- [ ] Any specific options we should include?
- [x] `Info` array needs some QA. Not sure I understand all of it.
- [ ] `nfs` can't support `xattr`, do we need to handle this in other places than migrate?

Fixes: https://github.com/lxc/incus/issues/1311

### DIFF
--- a/internal/server/storage/drivers/driver_nfs.go
+++ b/internal/server/storage/drivers/driver_nfs.go
@@ -1,0 +1,208 @@
+package drivers
+
+import (
+	"errors"
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"github.com/lxc/incus/v6/internal/linux"
+	"github.com/lxc/incus/v6/internal/migration"
+	deviceConfig "github.com/lxc/incus/v6/internal/server/device/config"
+	localMigration "github.com/lxc/incus/v6/internal/server/migration"
+	"github.com/lxc/incus/v6/shared/util"
+	"github.com/lxc/incus/v6/shared/validate"
+)
+
+type nfs struct {
+	dir
+}
+
+// Info returns info about the driver and its environment.
+func (n *nfs) Info() Info {
+	return Info{
+		Name:                         "nfs",
+		Version:                      "1",
+		DefaultVMBlockFilesystemSize: deviceConfig.DefaultVMBlockFilesystemSize,
+		OptimizedImages:              false,
+		PreservesInodes:              false,
+		Remote:                       n.isRemote(),
+		VolumeTypes:                  []VolumeType{VolumeTypeBucket, VolumeTypeCustom, VolumeTypeImage, VolumeTypeContainer, VolumeTypeVM},
+		VolumeMultiNode:              n.isRemote(),
+		BlockBacking:                 false,
+		RunningCopyFreeze:            false,
+		DirectIO:                     true,
+		MountedRoot:                  true,
+		Buckets:                      false,
+		SameSource:                   true,
+		IgnoreCleanup:                true,
+	}
+}
+
+// isRemote returns true indicating this driver uses remote storage.
+func (n *nfs) isRemote() bool {
+	return true
+}
+
+func (n *nfs) getMountOptions() string {
+	// Allow overriding the default options.
+	if n.config["nfs.mount_options"] != "" {
+		return n.config["nfs.mount_options"]
+	}
+	// We only really support vers=4.2
+	return fmt.Sprintf("vers=4.2,addr=%s", n.config["nfs.host"])
+}
+
+// Create is called during pool creation and is effectively using an empty driver struct.
+// WARNING: The Create() function cannot rely on any of the struct attributes being set.
+func (n *nfs) Create() error {
+	if n.config["source"] == "" {
+		return fmt.Errorf(`The "source" property must be defined`)
+	}
+
+	sourceStr := n.config["source"]
+
+	// Taken from the truenas driver
+	var host, path string
+	if strings.HasPrefix(sourceStr, "[") {
+		// IPv6 with brackets
+		endBracket := strings.Index(sourceStr, "]")
+		if endBracket == -1 || endBracket+1 >= len(sourceStr) || sourceStr[endBracket+1] != ':' {
+			// Malformed, treat whole string as path
+			host = ""
+			path = sourceStr
+		} else {
+			host = sourceStr[:endBracket+1]
+			path = sourceStr[endBracket+2:] // skip over "]:"
+		}
+	} else {
+		// Try normal IPv4/hostname
+		h, p, ok := strings.Cut(sourceStr, ":")
+		if ok {
+			host = h
+			path = p
+		} else {
+			// No colon: whole thing is path
+			host = ""
+			path = sourceStr
+		}
+	}
+
+	if path == "" {
+		fmt.Println(filepath.IsAbs(path))
+		return errors.New(`NFS driver requires "source" to be specified using the format: [<remote host>:]<remote path>`)
+	}
+
+	if host == "" {
+		if n.config["nfs.host"] == "" {
+			return errors.New(`NFS driver requires "nfs.host" to be specified or included in "source": [<remote host>:]<remote path>`)
+		}
+
+		host = n.config["nfs.host"]
+	} else {
+		n.config["nfs.host"] = host
+	}
+
+	n.config["source"] = fmt.Sprintf("%s:%s", host, path)
+	n.config["nfs.path"] = path
+
+	// Mount the nfs driver.
+	mntFlags, mntOptions := linux.ResolveMountOptions(strings.Split(n.getMountOptions(), ","))
+	err := TryMount(n.config["source"], GetPoolMountPath(n.name), "nfs4", mntFlags, mntOptions)
+	if err != nil {
+		return err
+	}
+
+	defer func() { _, _ = forceUnmount(GetPoolMountPath(n.name)) }()
+
+	return nil
+}
+
+// Mount mounts the storage pool.
+func (n *nfs) Mount() (bool, error) {
+	path := GetPoolMountPath(n.name)
+
+	// Check if already mounted.
+	if linux.IsMountPoint(path) {
+		return false, nil
+	}
+
+	sourcePath := n.config["source"]
+
+	// Check if we're dealing with an external mount.
+	if sourcePath == path {
+		return false, nil
+	}
+
+	// Mount the nfs driver.
+	mntFlags, mntOptions := linux.ResolveMountOptions(strings.Split(n.getMountOptions(), ","))
+	err := TryMount(sourcePath, GetPoolMountPath(n.name), "nfs4", mntFlags, mntOptions)
+	if err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
+
+// MigrationTypes returns the type of transfer methods to be used when doing migrations between pools in preference order.
+func (n *nfs) MigrationTypes(contentType ContentType, refresh bool, copySnapshots bool, clusterMove bool, storageMove bool) []localMigration.Type {
+	// NFS does not support xattr
+	rsyncFeatures := []string{"delete", "bidirectional"}
+	if util.IsTrue(n.Config()["rsync.compression"]) {
+		rsyncFeatures = append(rsyncFeatures, "compress")
+	}
+
+	return []localMigration.Type{
+		{
+			FSType:   migration.MigrationFSType_BLOCK_AND_RSYNC,
+			Features: rsyncFeatures,
+		},
+		{
+			FSType:   migration.MigrationFSType_RSYNC,
+			Features: rsyncFeatures,
+		},
+	}
+}
+
+// Validate checks that all provide keys are supported and that no conflicting or missing configuration is present.
+func (n *nfs) Validate(config map[string]string) error {
+	rules := map[string]func(value string) error{
+		// gendoc:generate(entity=storage_nfs, group=common, key=source)
+		//
+		// ---
+		//  type: string
+		//  scope: local
+		//  default: -
+		//  shortdesc: NFS remote storage path. Format: `[<host>:]<remote path>`. If `host` is omitted here, it must be set via `nfs.host`.
+		"source": validate.IsAny, // can be used as a shortcut to specify dataset and optionally host.
+
+		// gendoc:generate(entity=storage_nfs, group=common, key=nfs.host)
+		//
+		// ---
+		//  type: string
+		//  scope: global
+		//  default: -
+		//  shortdesc: Hostname or IP address of the remote NFS server. Optional if included in `source`, or a configuration is used.
+		"nfs.host": validate.IsAny,
+
+		// gendoc:generate(entity=storage_nfs, group=common, key=nfs.path)
+		//
+		// ---
+		//  type: string
+		//  scope: local
+		//  default: -
+		//  shortdesc: Remote NFS path. Typically inferred from `source`, but can be overridden.
+		"nfs.path": validate.IsAny,
+
+		// gendoc:generate(entity=storage_nfs, group=common, key=nfs.mount_options)
+		//
+		// ---
+		//  type: string
+		//  scope: local
+		//  default: -
+		//  shortdesc: Additional mount options for the NFS mount.
+		"nfs.mount_options": validate.IsAny,
+	}
+
+	return n.validatePool(config, rules, map[string]func(value string) error{})
+}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -142,6 +142,7 @@ jobs:
           - ceph
           - linstor
           - random
+          - nfs
         os:
           - ubuntu-24.04
         include:
@@ -436,6 +437,16 @@ jobs:
           # Update the runner env.
           echo "INCUS_LINSTOR_CLUSTER=${runner_ip}" >> "$GITHUB_ENV"
 
+      - name: Setup NFS
+        if: ${{ matrix.backend == 'nfs' }}
+        run: |
+          set -x
+          sudo apt-get install --no-install-recommends -y nfs-kernel-server
+          echo "/media 10.0.0.0/8(rw,sync,no_subtree_check,no_root_squash,no_all_squash) 100.64.0.0/8(rw,sync,no_subtree_check,no_root_squash,no_all_squash)" | sudo tee /etc/exports
+          sudo exportfs -a
+          sudo systemctl restart nfs-server.service
+          echo "INCUS_NFS_SHARE=$(hostname -I | cut -d' ' -f1):/media" >> "$GITHUB_ENV"
+
       - name: "Ensure offline mode (block image server)"
         run: |
           sudo nft add table inet filter
@@ -459,7 +470,7 @@ jobs:
           chmod +x ~
           echo "root:1000000:1000000000" | sudo tee /etc/subuid /etc/subgid
           cd test
-          sudo --preserve-env=PATH,GOPATH,GITHUB_ACTIONS,INCUS_VERBOSE,INCUS_BACKEND,INCUS_CEPH_CLUSTER,INCUS_CEPH_CEPHFS,INCUS_CEPH_CEPHOBJECT_RADOSGW,INCUS_LINSTOR_LOCAL_SATELLITE,INCUS_LINSTOR_CLUSTER,INCUS_OFFLINE,INCUS_SKIP_TESTS,INCUS_REQUIRED_TESTS, INCUS_BACKEND=${{ matrix.backend }} ./main.sh ${{ matrix.suite }}
+          sudo --preserve-env=PATH,GOPATH,GITHUB_ACTIONS,INCUS_VERBOSE,INCUS_BACKEND,INCUS_CEPH_CLUSTER,INCUS_CEPH_CEPHFS,INCUS_CEPH_CEPHOBJECT_RADOSGW,INCUS_LINSTOR_LOCAL_SATELLITE,INCUS_LINSTOR_CLUSTER,INCUS_NFS_SHARE,INCUS_OFFLINE,INCUS_SKIP_TESTS,INCUS_REQUIRED_TESTS, INCUS_BACKEND=${{ matrix.backend }} ./main.sh ${{ matrix.suite }}
 
   client:
     name: Client
--- a/cmd/incusd/instance_post.go
+++ b/cmd/incusd/instance_post.go
@@ -1023,10 +1023,13 @@ func migrateInstance(ctx context.Context, s *state.State, inst instance.Instance
 
 		// Cleanup instance paths on source member if using remote shared storage
 		// and there was no storage pool change.
-		if sourcePool.Driver().Info().Remote && req.Pool == "" {
-			err = sourcePool.CleanupInstancePaths(inst, nil)
-			if err != nil {
-				return fmt.Errorf("Failed cleaning up instance paths on source member: %w", err)
+		driverInfo := sourcePool.Driver().Info()
+		if driverInfo.Remote && req.Pool == "" {
+			if !driverInfo.IgnoreCleanup {
+				err = sourcePool.CleanupInstancePaths(inst, nil)
+				if err != nil {
+					return fmt.Errorf("Failed cleaning up instance paths on source member: %w", err)
+				}
 			}
 		} else {
 			// Delete the instance on source member if pool isn't remote shared storage.
--- a/cmd/incusd/storage_pools.go
+++ b/cmd/incusd/storage_pools.go
@@ -230,7 +230,9 @@ func storagePoolsGet(d *Daemon, r *http.Request) response.Response {
 			// If no member is specified and the daemon is clustered, we omit the node-specific fields.
 			if s.ServerClustered {
 				for _, key := range db.NodeSpecificStorageConfig {
-					delete(poolAPI.Config, key)
+					if key != "source" || !pool.Driver().Info().SameSource {
+						delete(poolAPI.Config, key)
+					}
 				}
 			} else {
 				// Use local status if not clustered. To allow seeing unavailable pools.
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2920,3 +2920,7 @@ A `used_by` field was added to the `GET /1.0/cluster/groups/{name}` endpoint.
 ## `bpf_token_delegation`
 
 This adds support for [eBPF token delegation](https://docs.ebpf.io/linux/concepts/token/).
+
+## `storage_driver_nfs`
+
+This adds an NFS storage driver.
--- a/doc/config_options.txt
+++ b/doc/config_options.txt
@@ -5700,6 +5700,40 @@ This value is required by some providers.
 ```
 
 <!-- config group storage_lvm-common end -->
+<!-- config group storage_nfs-common start -->
+```{config:option} nfs.host storage_nfs-common
+:default: "-"
+:scope: "global"
+:shortdesc: "Hostname or IP address of the remote NFS server. Optional if included in `source`, or a configuration is used."
+:type: "string"
+
+```
+
+```{config:option} nfs.mount_options storage_nfs-common
+:default: "-"
+:scope: "local"
+:shortdesc: "Additional mount options for the NFS mount."
+:type: "string"
+
+```
+
+```{config:option} nfs.path storage_nfs-common
+:default: "-"
+:scope: "local"
+:shortdesc: "Remote NFS path. Typically inferred from `source`, but can be overridden."
+:type: "string"
+
+```
+
+```{config:option} source storage_nfs-common
+:default: "-"
+:scope: "local"
+:shortdesc: "NFS remote storage path. Format: `[<host>:]<remote path>`. If `host` is omitted here, it must be set via `nfs.host`."
+:type: "string"
+
+```
+
+<!-- config group storage_nfs-common end -->
 <!-- config group storage_truenas-common start -->
 ```{config:option} source storage_truenas-common
 :default: "-"
--- a/doc/reference/storage_drivers.md
+++ b/doc/reference/storage_drivers.md
@@ -15,6 +15,7 @@ storage_cephfs
 storage_cephobject
 storage_linstor
 storage_truenas
+storage_nfs
 ```
 
 See the corresponding pages for driver-specific information and configuration options.
@@ -24,21 +25,21 @@ See the corresponding pages for driver-specific information and configuration op
 
 Where possible, Incus uses the advanced features of each storage system to optimize operations.
 
-| Feature                                   | Directory | Btrfs | LVM   | ZFS     | Ceph RBD | CephFS | Ceph Object | LINSTOR | TRUENAS |
-| :---                                      | :---      | :---  | :---  | :---    | :---     | :---   | :---        | :---    | :---    |
-| {ref}`storage-optimized-image-storage`    | no        | yes   | yes   | yes     | yes      | n/a    | n/a         | yes     | yes     |
-| Optimized instance creation               | no        | yes   | yes   | yes     | yes      | n/a    | n/a         | yes     | yes     |
-| Optimized snapshot creation               | no        | yes   | yes   | yes     | yes      | yes    | n/a         | yes     | yes     |
-| Optimized image transfer                  | no        | yes   | no    | yes     | yes      | n/a    | n/a         | no      | no      |
-| {ref}`storage-optimized-volume-transfer`  | no        | yes   | no    | yes     | yes      | n/a    | n/a         | no      | no      |
-| Copy on write                             | no        | yes   | yes   | yes     | yes      | yes    | n/a         | yes     | yes     |
-| Block based                               | no        | no    | yes   | no      | yes      | no     | n/a         | yes     | yes     |
-| Instant cloning                           | no        | yes   | yes   | yes     | yes      | yes    | n/a         | yes     | yes     |
-| Storage driver usable inside a container  | yes       | yes   | no    | yes[^1] | no       | n/a    | n/a         | no      | no      |
-| Restore from older snapshots (not latest) | yes       | yes   | yes   | no      | yes      | yes    | n/a         | no      | no      |
-| Storage quotas                            | yes[^2]   | yes   | yes   | yes     | yes      | yes    | yes         | yes     | yes     |
-| Available on `incus admin init`           | yes       | yes   | yes   | yes     | yes      | no     | no          | no      | no      |
-| Object storage                            | yes       | yes   | yes   | yes     | no       | no     | yes         | no      | no      |
+| Feature                                   | Directory | Btrfs | LVM   | ZFS     | Ceph RBD | CephFS | Ceph Object | LINSTOR | TRUENAS | NFS  |
+| :---                                      | :---      | :---  | :---  | :---    | :---     | :---   | :---        | :---    | :---    | :--- |
+| {ref}`storage-optimized-image-storage`    | no        | yes   | yes   | yes     | yes      | n/a    | n/a         | yes     | yes     | no   |
+| Optimized instance creation               | no        | yes   | yes   | yes     | yes      | n/a    | n/a         | yes     | yes     | no   |
+| Optimized snapshot creation               | no        | yes   | yes   | yes     | yes      | yes    | n/a         | yes     | yes     | no   |
+| Optimized image transfer                  | no        | yes   | no    | yes     | yes      | n/a    | n/a         | no      | no      | no   |
+| {ref}`storage-optimized-volume-transfer`  | no        | yes   | no    | yes     | yes      | n/a    | n/a         | no      | no      | no   |
+| Copy on write                             | no        | yes   | yes   | yes     | yes      | yes    | n/a         | yes     | yes     | no   |
+| Block based                               | no        | no    | yes   | no      | yes      | no     | n/a         | yes     | yes     | n/a  |
+| Instant cloning                           | no        | yes   | yes   | yes     | yes      | yes    | n/a         | yes     | yes     | no   |
+| Storage driver usable inside a container  | yes       | yes   | no    | yes[^1] | no       | n/a    | n/a         | no      | no      | yes  |
+| Restore from older snapshots (not latest) | yes       | yes   | yes   | no      | yes      | yes    | n/a         | no      | no      | yes  |
+| Storage quotas                            | yes[^2]   | yes   | yes   | yes     | yes      | yes    | yes         | yes     | yes     | no   |
+| Available on `incus admin init`           | yes       | yes   | yes   | yes     | yes      | no     | no          | no      | no      | yes  |
+| Object storage                            | yes       | yes   | yes   | yes     | no       | no     | yes         | no      | no      | no   |
 
 [^1]: Requires [`zfs.delegate`](storage-zfs-vol-config) to be enabled.
 [^2]: % Include content from [storage_dir.md](storage_dir.md)
--- a/doc/reference/storage_nfs.md
+++ b/doc/reference/storage_nfs.md
@@ -1,0 +1,28 @@
+(storage-nfs)=
+# NFS - `nfs`
+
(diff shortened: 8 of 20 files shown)


Foxboron:
@stgraber if you have any opinions or pointers to the checkboxes feel free to look over and I can investigate a bit.

I suspect some research needs to be done to figure out how we should interact with the uid/gid and squashing behavior of NFS mounts.

bensmrs:
Hi! Is this PR stalled? Do you need help?

Foxboron:
@bensmrs I think I need a bit of help to make sure that the `Info` struct is correct and that I'm not missing any details from the migration steps. I was hoping @stgraber had time to point in the correct direction on this part.

internal/server/storage/drivers/driver_nfs.go
```diff
@@ -0,0 +1,125 @@
...
+// Info returns info about the driver and its environment.
+func (n *nfs) Info() Info {
+	return Info{
+		Name:                         "nfs",
+		Version:                      "1",
+		DefaultVMBlockFilesystemSize: deviceConfig.DefaultVMBlockFilesystemSize,
+		OptimizedImages:              false,
+		PreservesInodes:              false,
+		Remote:                       n.isRemote(),
+		VolumeTypes:                  []VolumeType{VolumeTypeBucket, VolumeTypeCustom, VolumeTypeImage, VolumeTypeContainer, VolumeTypeVM},
```
stgraber, line 30: So initially my thought was to only allow `VolumeTypeCustom` on there, so only allowing NFS to be used for additional storage to instances either stored on local storage or stored on a proper block shared storage solution (clustered LVM, Ceph or Linstor).

I think we should at least exclude `VolumeTypeBucket` and `VolumeTypeImage`.

Container and VM, we'd need to see how well they work. My gut feeling given NFS locking is "not well", the question is then whether we still allow it but document that it's really not recommended or we plain not allow it and only keep custom volumes on it.

internal/server/storage/drivers/driver_nfs.go
```diff
@@ -0,0 +1,125 @@
...
+		VolumeMultiNode:              n.isRemote(),
+		BlockBacking:                 false,
+		RunningCopyFreeze:            false,
+		DirectIO:                     true,
+		MountedRoot:                  true,
+		Buckets:                      true,
```
stgraber: I think that should be set to false given above.

internal/server/storage/drivers/driver_nfs.go
```diff
@@ -0,0 +1,125 @@
...
+		Version:                      "1",
+		DefaultVMBlockFilesystemSize: deviceConfig.DefaultVMBlockFilesystemSize,
+		OptimizedImages:              false,
+		PreservesInodes:              false,
+		Remote:                       n.isRemote(),
+		VolumeTypes:                  []VolumeType{VolumeTypeBucket, VolumeTypeCustom, VolumeTypeImage, VolumeTypeContainer, VolumeTypeVM},
+		VolumeMultiNode:              n.isRemote(),
+		BlockBacking:                 false,
+		RunningCopyFreeze:            false,
+		DirectIO:                     true,
```
stgraber, line 34: Not sure about that one, does NFS handle O_DIRECT?

stgraber:
Ah, I just left a few comments in the Info function now.

internal/server/storage/drivers/driver_nfs.go
bensmrs: Well it depends on the amount of work required. 4 hours of work that could potentially go to the trash is ok. 4 days, not so much.

bensmrs:
> I was hoping @stgraber had time to point in the correct direction on this part.

Well now you’re served :þ

I can help review, fix stuff, and even write tests, don’t hesitate to ping me. I’d actually be happy to see it working pretty soon.

Foxboron:
@bensmrs thanks!

I'll probably not touch this in July, and there is a hackercamp and work stuff happening in August on my end that might not allow me a lot of energy to pick this up after hours.

I'd appreciate some pointers on the id remapping incus does and how that should play with the ID squashing NFS does. I *think* there should be some guidance and testing there to make sure it behaves as expected. It also takes a bunch of time which might not be needed if nfs reassings UID/GID anyway.

If you have time to write tests and stuff I'd be happy to give you access to my fork.

stgraber:
I don't believe that VFS idmap works on top of NFS at this point, so we'd be dealing with traditional shifting where Incus rewrites all the uid/gid as needed. 

What you want to ensure is that NFS is mounted the same way on all machines and that no uid/gid squashing is performed, then that should work fine.

Foxboron:
Should we call the driver `nfs4` to make sure we don't end up in a weird situation down the line where we don't want to support a `v5` along with the `v4` code? Is that realistic?

stgraber:
I think we should stick to `nfs` as `nfs4` may give the impression that we don't support nfs3 whereas the featureset we really need is perfectly fine for NFS3, if anything some of the bits in NFS4 may be problematic (built-in uid/gid mapping and such).

Exactly what kind of server version and configuration we can handle is probably something that's best addressed through the documentation for the driver.

symgryph:
NFS would be SO nice. Just an interested party!

internal/server/storage/drivers/driver_nfs.go
Foxboron: https://lore.kernel.org/linux-nfs/173810390475.3944426.14717347728766778857.pr-tracker-bot@kernel.org/T/#t

This was actually included with 6.14. I'm a bit unsure how to test this though.

Foxboron: I've left and spawing qemu instances seems to work just fine.

Foxboron:
@bensmrs

I did some changes so we can pass ipv6 `source=` paths. It's not great but couldn't come up with a more reliable way to split up on `::1:/somepath` or similar paths.

What sort of testing do we want on this? Locally creating pools and creating contianers and VMs work. Also attaching additional volumes.

bensmrs:
> What sort of testing do we want on this? Locally creating pools and creating contianers and VMs work. Also attaching additional volumes.

We’d basically extend the available filesystem drivers in the test matrix, to test everything we’re already testing: https://github.com/lxc/incus/blob/f614b4c4a724845f901c84b482757b5959e0b665/.github/workflows/tests.yml#L137-L144

The tests initialization routines allow you to define how your driver should be initialized in `test/backends`, although it’s not enough. I can have a quick look at it as soon as I’m done with my other PR.

bensmrs:
I started looking at the tests and am a bit confused by the following error:
```
> incus storage create incustest-KAg nfs source=10.1.0.227:/media
Error: NFS driver requires "nfs.host" to be specified or included in "source": [<remote host>:]<remote path>
```

Did I do something wrong here?
IMO, if you have an `nfs.host` option, I don’t think you should look for a host in `source`. Just make `source` be the path on `nfs.host`, or on localhost if no `nfs.host` is provided.

Foxboron:
> Did I do something wrong here?

Uhh, so it turns out that the `Makefile` has been moving things from being built into `build/` to running `go install`. So there is a slight chance I have missed testing an iteration while working on this.

> IMO, if you have an nfs.host option, I don’t think you should look for a host in source. Just make source be the path on nfs.host, or on localhost if no nfs.host is provided.

I was thinking mimicking the behavior of `truenas.host` and `source` from the `truenas` driver. Where you can just specify `nfs.host` globally and pass a path to `source`. But I haven't nailed the entire behavior there I suspect. I'm a little bit unsure what the most ergonomic way for the options to behave, that also harmonizes with the existing drivers.

bensmrs:
> Uhh, so it turns out that the Makefile has been moving things from being built into build/ to running go install. So there is a slight chance I have missed testing an iteration while working on this.

No problem, I ended up using `nfs.host` for driver initialization.

> I was thinking mimicking the behavior of truenas.host and source from the truenas driver. Where you can just specify nfs.host globally and pass a path to source. But I haven't nailed the entire behavior there I suspect. I'm a little bit unsure what the most ergonomic way for the options to behave, that also harmonizes with the existing drivers.

Ok I can’t beat this argument :)

I’m getting some permission denied errors in my tests, I’m investigating.

bensmrs:
So, maybe I’m missing something when initializing the pool, but I can’t get past permission errors. Container creation and launch work well, but that’s not enough, see https://github.com/bensmrs/incus/actions/runs/18648669885/job/53161380407?pr=4

Am I missing something in my setup? See https://github.com/lxc/incus/blob/5733ce8ba08e118b68dd6900546bc0de8f4722d9/.github/workflows/tests.yml#L394-L395

Foxboron:
Hmmm, I've only tried `incus exec container bash` and manually created files on NFS volumes on both ends. And it worked. Are we sure that `/media` doesn't have any weird permissions in ubuntu? Does it exist etc?

bensmrs:
Container launch works, so regular operations don’t seem to cause any problem.
I’ll debug a bit more later, I was just wondering if your setup had anything specific other than the squash options.

bensmrs:
Well I’m currently out of ideas. `user_is_instance_user` fails with a permission error. We can’t list `/root` from within the container, it appears owned by `0:0`, `whoami` fails because user `0` is not found, and from the host, the mount appears owned by `1000000:1000000`. I don’t know where to go from there; maybe I’m missing something obvious happening in the OpenFGA tests.

Foxboron:
I'm honestly not sure.

My `export` mount has this
```nfs
/srv/nfs/incus          192.168.1.105/32(rw,sync,no_subtree_check,no_root_squash,no_all_squash)
```

The top directory is owned by a backup user with UID `1001`.

```
$ ls /srv/nfs | grep incus
drwxr-xr-x 1 backup backup 188 Oct 20 21:11 incus
```

Adding the storage
```
λ nfs morten/nfs Ɇ » sudo incus storage create test-pool nfs source='192.168.1.8:/srv/nfs/incus'
Storage pool test-pool created
λ nfs morten/nfs Ɇ » sudo incus launch images:archlinux/current/amd64 --storage test-pool
```


The container works as expected.
```
[root@moving-jennet ~]# touch lol
[root@moving-jennet ~]# ls -lah
total 0
drwxr-x--- 1 root root  26 Oct 20 19:25 .
drwxr-xr-x 1 root root 178 Oct 20 19:25 ..
drwx------ 1 root root  12 Oct  1 16:11 .gnupg
drwx------ 1 root root   0 Oct  1 16:11 .ssh
-rw-r--r-- 1 root root   0 Oct 20 19:28 lol
[root@moving-jennet ~]# whoami
root
```

With the actual `rootfs` on the NAS looks like.
```
[root@nassen root]# pwd
/srv/nfs/incus/containers/moving-jennet/rootfs/root
[root@nassen root]# ls -lah
total 0
drwxr-x--- 1 1000000 1000000  26 Oct 20 21:25 .
drwxr-xr-x 1 1000000 1000000 178 Oct 20 21:25 ..
drwx------ 1 1000000 1000000  12 Oct  1 18:11 .gnupg
-rw-r--r-- 1 1000000 1000000   0 Oct 20 21:28 lol
drwx------ 1 1000000 1000000   0 Oct  1 18:11 .ssh
```

bensmrs:
It works on my machine too, but the testing environment is a bit more involved unfortunately, and it tests plenty of non-trivial stuff… It gets very tricky very fast :)

Foxboron:
> It works on my machine too, but the testing environment is a bit more involved unfortunately, and it tests plenty of non-trivial stuff… It gets very tricky very fast :)

Ahh, right! Is there a good way to replicate the test setup?

bensmrs:
Several ways! Either on the CI, like I’ve done on my own branch (I’ve done plenty of changes just for debugging purposes, but most of what I did in `test/` should be kept), or on your local machine. From my branch, I just have to run `INCUS_CONCURRENT=1 INCUS_VERBOSE=1 INCUS_OFFLINE=1 INCUS_TMPFS=1 INCUS_INSPECT=1 INCUS_BACKEND=nfs INCUS_NFS_HOST=127.0.0.1 ./main.sh standalone` in the `test/` directory (note that it won’t work in your branch).

I have a bit of time to continue looking at the tests, so I think you can just focus on the other tests failing for now :)

bensmrs:
Ok I managed to reproduce the bug. The BusyBox image that the Incus test suite uses does indeed not work with the NFS driver…
Smells like an idmap problem, but a Trixie image works well.

bensmrs:
Minimal (non-)working example:

```
> root@incus-dev:~/incus# test/deps/import-busybox
Image imported as: bd06ce5e307a730411aad70ee4583a07cf62ad38c466a45a9d1e431a7eccb054
> root@incus-dev:~/incus# incus launch bd06ce5e307a730411aad70ee4583a07cf62ad38c466a45a9d1e431a7eccb054 busybox-test --storage nfs
Launching busybox-test
> root@incus-dev:~/incus# incus exec busybox-test -- ls -al
ls: .: Permission denied
```

If you can confirm the behavior on your side @Foxboron, then I guess you’ll have to investigate :)
Else, we’ll have to spot differences between Arch and Debian environments :)

Foxboron:
```
> root@incus-dev:~/incus# incus exec busybox-test -- ls -al
ls: .: Permission denied
```

*This* specific error is because the permissions on `/root` lacks executable bit. So setting `chmod +x /root` solves that. So I *think* that is an actual bug in the image being created.

I'll try and reproduce the error reported earlier.

Foxboron:
Okay, I think the error reported above I the same thing.

```
 timeout --foreground 120 /home/runner/go/bin/incus file push /home/runner/work/incus/incus/test/tmp.gTh/tmp oidc-openfga:user-foo/root/tmpfile.txt --verbose
```

You are pushing to `/root` without the executable bit.

bensmrs:
Except… it works on every other driver.

Foxboron:
I don't disagree it's weird, but the permissions are obviously wrong it seems.

```diff
diff --git a/test/deps/import-busybox b/test/deps/import-busybox
index 3c72e87cec67..7dab4e42e747 100755
--- a/test/deps/import-busybox
+++ b/test/deps/import-busybox
@@ -236,6 +236,7 @@ class BusyBox(object):
         for path in ("dev", "mnt", "proc", "root", "sys", "tmp"):
             directory_file = tarfile.TarInfo()
             directory_file.type = tarfile.DIRTYPE
+            directory_file.mode = 0o755
             if split:
                 directory_file.name = "%s" % path
                 target_tarball_rootfs.addfile(directory_file)
```

or similar is what we want for `/root` at least.

bensmrs:
On ZFS,
```
> root@incus-dev:~# incus launch bd06ce5e307a busybox-test-2 --storage zfs
Launching busybox-test-2
> root@incus-dev:~# incus exec busybox-test-2 -- ls -al
total 2
drw-r--r--    2 0        0                2 Jan  1  1970 .
drwxr-xr-x   12 0        0               13 Oct 20 21:27 ..
```
From the host,
```
> root@incus-dev:~# ls -ld /var/lib/incus/storage-pools/nfs/containers/busybox-test/rootfs/root
drw-r--r-- 2 1000000 1000000 4096 Jan  1  1970 /var/lib/incus/storage-pools/nfs/containers/busybox-test/rootfs/root
> root@incus-dev:~# ls -ld /var/lib/incus/storage-pools/zfs/containers/busybox-test-2/rootfs/root
drw-r--r-- 2 root root 2 Jan  1  1970 /var/lib/incus/storage-pools/zfs/containers/busybox-test-2/rootfs/root
```

bensmrs:
Maybe @stgraber could provide some insight on this one.

stgraber:
@bensmrs all the other filesystems support VFS idmap shifting so the kernel does the shifting for us. NFS doesn't, so instead we have to rely on the old-school manual shifting.

bensmrs:
Oh, right! But what about that missing `+x`? Idmap shifting doesn’t touch that, does it?

stgraber:
Hmm, yeah, not sure what's going on with that.

I agree that the busybox root folder doesn't seem to have the correct mode, just not sure why this was never a problem before ;)

bensmrs:
@Foxboron, does `chown` work for you?
```
root@trixie:~# ls -ld foo
-rw-r--r-- 1 root root 7 Oct 20 23:01 foo
root@trixie:~# chown 55:55 foo
chown: changing ownership of 'foo': Operation not permitted
```

This is required by the test suite.

bensmrs:
Also, pretty nasty effect of remote cluster storage: `Create` is **not** called on joining nodes: https://github.com/lxc/incus/blob/eb0ce6cac16e9bfd04935891c5bf98782108460a/internal/server/storage/backend.go#L191-L203

This means that https://github.com/Foxboron/incus/blob/62b497dc3ee324f5b56742c7f906e1f3610b7c0a/internal/server/storage/drivers/driver_nfs.go#L57-L59 is not checked (which is ok), but more importantly, that https://github.com/Foxboron/incus/blob/62b497dc3ee324f5b56742c7f906e1f3610b7c0a/internal/server/storage/drivers/driver_nfs.go#L104 never gets called. The consequence is that https://github.com/Foxboron/incus/blob/62b497dc3ee324f5b56742c7f906e1f3610b7c0a/internal/server/storage/drivers/driver_nfs.go#L128 leads to `sourcePath = ""`. I think it should be reconstructed from `nfs.host` and `nfs.path`, which are properly sent on cluster join.

The consequence is the following error: `Error: Failed to join cluster: Failed to initialize member: Failed to initialize storage pools and networks: Failed to create storage pool "data": Failed to mount "" on "/home/runner/work/incus/incus/test/tmp.HOe/H2X/storage-pools/data" using "nfs4": invalid argument` (notice the `""`).

The previous message and this one are the two current blockers for respectively the standalone and the clustered test suites.

stgraber:
@bensmrs did you try setting `SameSource: true` in the driver Info?
That handles this same situation but in the LVMcluster case, basically telling Incus that whatever the resulting `source` is from the initial server, it should be replicated to all others.

bensmrs:
Gah! if we start using obscure options now…
Just tested, and nope, same behavior.
I don’t think `storagePoolsPostCluster` (the only place using `SameSource`) even gets called, as I can’t find any of its debug messages in my logs. From a quick skim-read, I think that `SameSource` only really matters in case we create the pool on an already-existing cluster, not when joining new nodes (but I may be wrong).

stgraber:
Ah, that feels like a bug in the SameSource handling then, it should also enforce that any joining server gets the existing source config and doesn't get prompted for it.

Not sure if you want to tackle this one or if I should have @presztak look into it?

bensmrs:
Sure, I’ll have a look

Foxboron:
> @Foxboron, does chown work for you?

Yes, it does work from the busybox image.

```
/ # chown 55:55 lol
/ # ls -lah
total 4K
-rw-r--r--    1 55       55             0 Oct 21 18:59 lol
```

bensmrs:
> > @Foxboron, does chown work for you?
> 
> Yes, it does work from the busybox image.
> 
> ```
> / # chown 55:55 lol
> / # ls -lah
> total 4K
> -rw-r--r--    1 55       55             0 Oct 21 18:59 lol
> ```

Hmmm then we have a different behavior on our systems, and we definitely have to solve this one…

bensmrs:
Ok I think I found my way around for the clustering part.
Quick question: if we’re using a remote storage driver, don’t we want to allow migration?

stgraber:
> Quick question: if we’re using a remote storage driver, don’t we want to allow migration?

We should try it. No idea how well that works with the file on NFS, but worth a shot :)
Anyway, in theory, yes, we should allow it.

bensmrs:
Alright, I’m trying stuff in my branch. I think trivial no-ops should work for most operations, we’ll see.

Foxboron:
@bensmrs I feel a bit bad leaving you with a bunch of pieces you have to pick up :x thanks for the effort! 

> Hmmm then we have a different behavior on our systems, and we definitely have to solve this one…

I'll try and figure out the `chmod` issue. But this is a container on nfs root on Ubuntu?

Foxboron:
Fwiw, we can just retain the custom volume support if it turns out that hosting rootfs on NFS is way too problematic for us.

bensmrs:
> @bensmrs I feel a bit bad leaving you with a bunch of pieces you have to pick up :x thanks for the effort!

Please don’t! There’s unfortunately quite a big gap in Incus storage pools between “it works on my machine” and “it passes the tests”.

> I'll try and figure out the chmod issue. But this is a container on nfs root on Ubuntu?

Continue using BusyBox really, as I have the same problem with it.

> Fwiw, we can just retain the custom volume support if it turns out that hosting rootfs on NFS is way too problematic for us.

Nah that would be too sad :)

Foxboron:
> Continue using BusyBox really, as I have the same problem with it.

Yes, but your host is Ubuntu? It the difference is between Arch host and Ubuntu host there might be kernel things going on.

bensmrs:
Right, sorry. My host is running Debian and the runners are running Ubuntu. Both fail with chown.

bensmrs:
Hmmm my new blocker in the tests is there: https://github.com/lxc/incus/blob/caf5adadc60a47b3c76070c368dfdaa2963586fb/test/suites/clustering.sh#L867-L869

The check doesn’t fail. Could you tell me why it’s supposed to, @stgraber?
I’m disabling it for NFS for now until it’s clearer.

Edit: apart from that, clustered tests are now working

bensmrs:
I’m also having trouble with this one: https://github.com/lxc/incus/blob/caf5adadc60a47b3c76070c368dfdaa2963586fb/test/suites/clustering.sh#L2790-L2794

I can’t find the code paths leading to `CreateImage` not being repeatedly called by the two current remote drivers. And surely enough, in the tests, I’m getting 3 and not 1 for NFS.

bensmrs:
@Foxboron the amount of test-related commits is piling up on my side, could you give me write access to your branch (with force-push) to include them? If you’re not comfortable with that, you can totally cherry pick from https://github.com/bensmrs/incus/tree/nfs.

bensmrs:
Setting `security.privileged` to `true` semi-obviously solves the `chown` problem. Maybe some missing caps or idmap issue?

stgraber:
> The check doesn’t fail. Could you tell me why it’s supposed to, @stgraber?

That check is in a block that covers the storage drivers which use block as the backing for their custom filesystem volumes. On those, it's unsafe to attach the same custom volume to instances on two different servers as that would lead to filesystem corruption.

The `nfs` driver should generally follow the lead of `cephfs` in that regard where this is allowed.

bensmrs:
Ok thanks. I’ll keep this single excerpt conditioned to `!= "nfs"` then.

@Foxboron in order to better debug our `chown` problem, could you tell me the real on-host `uid:gid` you have when you set an apparent `55:55` from within the container?

Foxboron:
> @Foxboron in order to better debug our chown problem, could you tell me the real on-host uid:gid you have when you set an apparent 55:55 from within the container?

A little bit unsure how to get that information. I'm personally just running everything as root locally :x So I suspect everything on my end is priviledged?

bensmrs:
Just `ls -al` the actual path from the host (after performing a `chown` from within your container), I’m interested in seeing if you have idmap working.

Foxboron:
> Just ls -al the actual path from the host (after performing a chown from within your container), I’m interested in seeing if you have idmap working.

Argh. I realized now that I have been swapping around too much. The case where I said `chown` works was me spinning up a `busybox` container without the nfs root. With the NFS root I'm getting the same issue as you are seeing!

bensmrs:
Well, have fun debugging this one :þ

(I’m rebasing the commits right now)

jack9603301:
This draft seems very cool, but why not choose to manually perform NFS automatic mounting from the bottom of the operating system and then run incus in dir mode?

stgraber:
> This draft seems very cool, but why not choose to manually perform NFS automatic mounting from the bottom of the operating system and then run incus in dir mode?

If you do that, then Incus doesn't know that the same data can be accessed from all systems in a cluster.

Foxboron:
> This draft seems very cool, but why not choose to manually perform NFS automatic mounting from the bottom of the operating system and then run incus in dir mode?

I would really want to avoid having to manage the cluster node `fstabs` for every new NFS mount I want.

jack9603301:
> I would really want to avoid having to manage the cluster node `fstabs` for every new NFS mount I want.

Seems like a good reason

jack9603301:
> If you do that, then Incus doesn't know that the same data can be accessed from all systems in a cluster.

Sorry, I don't understand what you mean

bensmrs:
> > If you do that, then Incus doesn't know that the same data can be accessed from all systems in a cluster.
> 
> Sorry, I don't understand what you mean

The `dir` backend is not suited for cluster operation because it just considers the local filesystems available to the host. This makes moving data from node to node basically copy it from one node to the other. There are two problems in using NFS with this: i) copying the data is useless, as it is available to every node, and, more critical even, ii) moving data, e.g. instances, would essentially make the receiving node write such data at the same place as it is currently being stored, without any kind of FS locking mechanism.

jack9603301:
> The `dir` backend is not suited for cluster operation because it just considers the local filesystems available to the host. This makes moving data from node to node basically copy it from one node to the other. There are two problems in using NFS with this: i) copying the data is useless, as it is available to every node, and, more critical even, ii) moving data, e.g. instances, would essentially make the receiving node write such data at the same place as it is currently being stored, without any kind of FS locking mechanism.

I don't think it's a problem in regular operation and maintenance applications, because the dir backend ultimately relies on the Linux file system tree structure, which will bring more options for deployment. For example, operation and maintenance personnel can choose NFS server remote control mounting (with a single point of failure), Ceph cluster file system or glusterfs cluster file system, DRBD disk synchronization system and deploy incus container data on top. The data will always be automatically synchronized by the underlying system. This is Linux - each part only does what it does best.

Foxboron:
> I don't think it's a problem in regular operation and maintenance applications, because the dir backend ultimately relies on the Linux file system tree structure, which will bring more options for deployment. For example, operation and maintenance personnel can choose NFS server remote control mounting (with a single point of failure), Ceph cluster file system or glusterfs cluster file system, DRBD disk synchronization system and deploy incus container data on top. The data will always be automatically synchronized by the underlying system. This is Linux - each part only does what it does best.

I don't think arguing against the merits of an NFS storage driver in the PR adding said storage driver is helpful nor welcome.

jack9603301:
> I don't think arguing against the merits of an NFS storage driver in the PR adding said storage driver is helpful nor welcome.

Sorry, I don't understand what you mean

bensmrs:
Please keep debate away from this PR. The forum is here for this kind of things, and, more moderately, the issue section, to discuss early design-stage decisions. Your use case, as legitimate as it may seem to you, unfortunately doesn’t cover a tenth of all the use cases of Incus users which we have to account for.

jack9603301:
> Please keep debate away from this PR. The forum is here for this kind of things, and, more moderately, the issue section, to discuss early design-stage decisions. Your use case, as legitimate as it may seem to you, unfortunately doesn’t cover a tenth of all the use cases of Incus users which we have to account for.

OK, but in fact this is the most typical problem of production deployment. They may not use nfs backend. Due to the complexity of the production system, they directly use dir backend.

bensmrs:
The more I think about it, the more I feel we need to intercept syscalls. It shouldn’t be incredibly hard, and I think it would look pretty similar to the pre-existing `setxattr` interception. However, special care would need to be taken to filter out non-NFS-mounted paths. WDYT @stgraber? Should we go that way?

(The other idea could be to mount the FS at another location and perform bind-mounting to the actual rootfs location.)

stgraber:
> The more I think about it, the more I feel we need to intercept syscalls. It shouldn’t be incredibly hard, and I think it would look pretty similar to the pre-existing `setxattr` interception. However, special care would need to be taken to filter out non-NFS-mounted paths. WDYT @stgraber? Should we go that way?
> 
> (The other idea could be to mount the FS at another location and perform bind-mounting to the actual rootfs location.)

Can you provide a summary of the issue as it currently stands? There's been a lot of back and forth over the past few days :)

In general, if we're talking about intercepting chown, that'd be a terrible idea because it's called extremely frequently (so could cause a DoS) and has various variants that may be very hard to handle safely (including calls from within namespaces or chroots).

Foxboron:
> Can you provide a summary of the issue as it currently stands? There's been a lot of back and forth over the past few days :)

On an NFS root `chown` does not work.

Minimal reproducer from what I can tell.
`lxc-usernsexec -m u:0:1000000:1000000000 -m g:0:1000000:1000000000 -- /bin/chown 55:55 testfile`

It's unclear to me *why* this is. I have ruled out `busybox` bugs as this happens with an Arch userland as well as the busybox test images.

stgraber:
@brauner can you look into this with your VFS maintainer hat on?

To reproduce this on pengar, you can do:
```
root@pengar:~# mkdir -p /mnt/nfs
root@pengar:~# mount -t nfs truenas01.shf.lab.linuxcontainers.org:/mnt/test/incus-nfs /mnt/nfs
Created symlink /run/systemd/system/remote-fs.target.wants/rpc-statd.service → /lib/systemd/system/rpc-statd.service.
root@pengar:~# ls -lh /mnt/nfs
total 512
drwxr-xr-x 2 100000 100000 3 Oct 23 21:01 rootfs
root@pengar:~# ls -lh /mnt/nfs/rootfs/
total 512
-rw-r--r-- 1 100000 100000 0 Oct 23 21:01 foo
root@pengar:~# lxc-usernsexec -m b:0:100000:65536 -- chown 1234:1234 /mnt/nfs/rootfs/foo
chown: changing ownership of '/mnt/nfs/rootfs/foo': Operation not permitted
root@pengar:~# lxc-usernsexec -m b:0:100000:65536 -- ls -l /mnt/nfs/rootfs/foo
-rw-r--r-- 1 root root 0 Oct 23 21:01 /mnt/nfs/rootfs/foo
root@pengar:~# chown 101234:101234 /mnt/nfs/rootfs/foo 
root@pengar:~# lxc-usernsexec -m b:0:100000:65536 -- ls -l /mnt/nfs/rootfs/foo
-rw-r--r-- 1 1234 1234 0 Oct 23 21:01 /mnt/nfs/rootfs/foo
```

So basically, doing the chown from within a userns going through the uid/gid map fails but doing the exact same chown as real root on the host and then accessing the result from within the namespace is fine.

Pengar is on 6.16.4, I've also reproduced it on 6.16.11. Not sure what the other two have been running, but probably similarly recent kernels. This is NFSv3 so there shouldn't be too much NFS black magic in theory.

stgraber:
Okay, so @brauner and I got to the bottom of this one and it's unfortunately not good.

There is no fundamental kernel issue here, it's more of a fundamental design issue with how NFS works...

Since kernel idmap shift isn't supported on top of NFS, we instead only rely on the user namespace. This means that when we have root in the container (uid=0) outside, chown /foo from its current ownership (uid/gid 0/0) to a new ownership (uid/gid 1234/4567), it actually gets translated to uid 100000 wanting to chown a file owned by 100000/100000 to 101234/104567.

That's the request (SETATTR) which gets sent to the server and the server rejects that as the credential (uid=100000/gid=100000) isn't privileged over the target uid/gid (101234/104567).

There is no provision in the NFS protocol to pass through the uidmap/gidmap to effectively inform the server of what uid/gid range the requestor is privileged over and so no way to actually properly handle such a request.

stgraber:
So we have a few things to consider here:

1) At the kernel level, if we were to implement VFS idmap on top of NFSv3 (v4 would be a massive headache), it would solve this situation, but ONLY for containers where the container's uid=0 is mapped to the host's uid=0, any other situation would still fail the server-side check
2) At the userspace level, there is an NFS FUSE client (https://github.com/sahlberg/fuse-nfs) but it's FUSE2 only as far as I can tell and as a result, doens't have @mihalicyn's support for VFS idmap. Getting it to support VFS idmap would provide a (slower) way forward.
3) We limit ourselves to what works today. That is 1) shared custom volumes (with a clear note about chown not working from within an unpriv container) 2) VM storage (possibly)

3) seems safe to do now, it's basically just putting some extra checks on this branch to provide the limited subset. 2) would be nice to have as that NFS client codebase isn't too hard to build, so if we can apply something like (https://github.com/mihalicyn/fuse-overlayfs/commit/89a1af3af9b989438962803524a749b522f3fa5c) to it, it would provide a slow but otherwise functional option for unpriv containers. 1) would need a motivated kernel dev to work on, as @brauner was pointing out, the fact that nobody really talked about this issue before doesn't bode too well as far as general interest for something like this.

bensmrs:
And couldn’t an idmapped bind mount be a solution? The host mounts the NFS shares somewhere, then bind-mounts it to `/var/lib/incus/containers/<instance>/` with the proper shift.

Or maybe it’s bindfs-specific, in which case we’re back to using FUSE…

stgraber:
> And couldn’t an idmapped bind mount be a solution? The host mounts the NFS shares somewhere, then bind-mounts it to `/var/lib/incus/containers/<instance>/` with the proper shift.
> 
> Or maybe it’s bindfs-specific, in which case we’re back to using FUSE…

That's 1), idmap bind-mount requires per-filesystem kernel code to handle it. NFS does not have support for it. FUSE in general does but requires each filesystem to add some logic, which is why I also brought up 2) as a quicker option to get something done.

bensmrs:
Ok, I had in mind bindfs’ `uid-offset` and `gid-offset` options and just thought that maybe this translation mechanism existed for bind mounts (the `bind` of `bindfs` made my brain shortcut to my previous question).

I’ll have a quick look at fuse-nfs.

bensmrs:
https://github.com/bensmrs/fuse-nfs should be ok-ish for FUSE3; I’ll try to implement VFS idmap tomorrow.

bensmrs:
My patches should now work (I also integrated a long-waiting PR that I feel is desirable for us), but testing it will probably be painful. Are there quick ways for me to test the patched libfuse, or do I have to compile @mihalicyn’s tree? (in which case, big meh on my side as I have little time for it).
Or you could test on your side @stgraber if you already have everything ready…

stgraber:
Cool!

If we get this working, we'd probably do a scheme like `/var/lib/storage-pools/my-nfs` is the NFS kernel mount, then `/var/lib/storge-pools/my-nfs/.fuse` is the same share but mounted over FUSE. Then whenever dealing with an unprivileged container, we'd use the `.fuse` path and everything else goes through the kernel client.

Foxboron:
@stgraber thanks for investigating the issue!

Whats the proper way forward for this? Try the `fuse` strategy? I guess it means we also need more dependencies packages in downstreams as well?

stgraber:
I think the next steps should be:

 - Try to get the FUSE filesystem to run on FUSE3 and support VFS idmap (I think @bensmrs was looking into that)
 - Get the driver to detect the presence of the FUSE client and if present, mount both the kernel version and user space version of the share
 - Tweak the storage driver to adjust its featureset based on what's present on the system

Given the new FUSE client is going to take a little while to be packed in all distros, I don't want us to be hard depending on it.

On my end, I'll be happy to just build and ship the client as part of the Zabbly package, so we can at least get it out to a whole bunch of users (including folks using IncusOS) pretty rapidly while the distro packagers get it through to the other distros.

bensmrs:
> Try to get the FUSE filesystem to run on FUSE3 and support VFS idmap (I think @bensmrs was looking into that)

I’d half-confidently say that my fork does that already. The blocker for me is to have a kernel with the necessary patches to make it run. I don’t really have time planned to build such a kernel before, say, december. I’d be more than happy to steal a disk image to deploy on my staging environment though ^^

stgraber:
> I’d half-confidently say that my fork does that already. The blocker for me is to have a kernel with the necessary patches to make it run. I don’t really have time planned to build such a kernel before, say, december. I’d be more than happy to steal a disk image to deploy on my staging environment though ^^

Ah, what do you need on the kernel side? I thought we'd had FUSE VFS idmap for a bit now (6.12 apparently).

bensmrs:
I adapted the code from the patch you linked, which uses `#if FUSE_VERSION >= FUSE_MAKE_VERSION(3, 18)` conditions. FUSE 3.18 doesn’t seem to be a thing yet (and my testing environment running Trixie indeed only has 3.17).
So I guessed that the VFS patches didn’t land in mainline kernels.

stgraber:
> I adapted the code from the patch you linked, which uses `#if FUSE_VERSION >= FUSE_MAKE_VERSION(3, 18)` conditions. FUSE 3.18 doesn’t seem to be a thing yet (and my testing environment running Trixie indeed only has 3.17). So I guessed that the VFS patches didn’t land in mainline kernels.

That's odd because I'm definitely seeing the commits in my clone of Linus' tree.

stgraber:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=16e1503eaf329129170e4e7a078aee17686967a5

bensmrs:
I remember the build breaking when removing the version condition. Checking libfuse source code, d393ffa85b0926374c8df543a9ffc81b1d0ce232 suggests some kind of idmap support, but `struct fuse_ctx` has not been patched (so we may need to partially apply https://github.com/mihalicyn/libfuse/tree/idmap_support).

I’m not comfortable with Incus/Zabbly providing patched libfuse, though, so maybe a nudge to upstream would be good.

stgraber:
Yeah, I don't mind doing the patched libfuse dance in /opt/incus/lib like we do some other libraries but that would be something I'd only really want to do after upstream has merged the needed support.

I pinged @mihalicyn about it.

bensmrs:
And to think that it just started with “it would be nice to have NFS support” :D

stgraber:
Yeah, that one became complicated pretty quick :)

Foxboron:
I took your word for it when you said it would "probably" be easy :melting_face:

bensmrs:
It’s always nice when the issue on which you’ve been working for 6 month is marked “Easy” (which, may I remind the description, means “Good for new contributors”) :)

stgraber:
> I took your word for it when you said it would "probably" be easy 🫠

Haha, yeah, the storage logic really should have been pretty easy, take the `dir` backend and make it act as clustered similar to `lvmcluster`.

Now the underlying NFS protocol being annoying with idmap wasn't something I had seen coming ;)
It kinda makes me want to look at CIFS, see if we have the same problem over there...

bensmrs:
> It kinda makes me want to look at CIFS

Said… no one? ever?

stgraber:
> > It kinda makes me want to look at CIFS
> 
> Said… no one? ever?

Well, in this case it coming from more of a Windows world where UIDs/GIDs are not really a thing makes me hopeful that the protocol doesn't just carry the user cred and file details but instead has the kernel module handle the checks and then tells the remote storage what to write.

If that's the case, then it should work fine here. NFS is only broken because it's too UNIX :)

bensmrs:
Ok, interesting. Now I’m curious too!

stgraber:
@mihalicyn provided this one as an example of conversion

https://gitlab.com/virtio-fs/virtiofsd/-/merge_requests/245/diffs

bensmrs:
I’m confused. What question are you answering with this message?

stgraber:
> I’m confused. What question are you answering with this message?

How to get a FUSE filesystem to use the VFS idmap stuff without needing libfuse changes.

bensmrs:
Ok, I’ll see how this can be translated to my code then.

bensmrs:
So, after a frustrating attempt at making it work, I realized that even though the needed patches are in the main branch of libfuse, it’s not in any released version, so I need to compile libfuse myself. It also looks like Trixie’s stock kernel is not fit for these patches, so I also had to install the backported kernel. I should be nearing something that might work pretty soon (yeah, that doesn’t exude confidence).

bensmrs:
Well, I give up, I can’t get any of my attempts to work, even though the proper capabilities are advertised to and by FUSE. Either I’m missing a pretty obvious wiring, or something’s just not ready yet. I’ve left two branches on my fork, in case someone wants to have a look.

PR #2045, opened 2025-05-01
internal/iprange: add tests for the iprange.Range struct

The127:
Adds tests for the ContainsIp and String methods on the iprange.Range struct type.

### DIFF
--- a/internal/iprange/range_test.go
+++ b/internal/iprange/range_test.go
@@ -1,0 +1,136 @@
+package iprange_test
+
+import (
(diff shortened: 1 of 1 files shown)


The127:
I'm gonna go out on a limb here and assume the tests failures are flakes, since i did not touch any code, I just added tests ^^https://github.com/lxc/incus/actions/runs/14771884305/job/41473150090?pr=2045

stgraber (approved):

PR #2051, opened 2025-05-01
lint: Codespell exclude generated docs

breml:


test/lint/codespell.sh
```diff
@@ -1,2 +1,2 @@
 #!/bin/sh -eu
-codespell -S 'internal/server/apparmor/*.profile.go' -S 'doc/html/*' -I .codespell-ignore client cmd doc grafana internal scripts shared test
+codespell -S 'internal/server/apparmor/*.profile.go' -S 'doc/html/*' -I .codespell-ignore client cmd doc grafana internal scripts shared test --ignore-regex '"nd"'
```
stgraber, line 2: Instead we should ignore all of `doc/reference/manpages/*` as that's also generated content.

stgraber: I've made the change

stgraber (approved):

Issue #1869, opened 2025-03-31
stateful IPv6 for OCI containers


defect-track:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

This feature request is based on the following forum thread https://discuss.linuxcontainers.org/t/no-ipv6-for-incus-oci-containers/21086

In general you can work with dynamic IP addresses but for some services like DNS it is recommended to use a static IP. With the latest release 6.11 we can now configure custom DNS for OVN, bridge, etc. networks which is great. Trying to setup an OCI DNS on a static IPv6 only network. Current workaround is to have IPv4 enabled or use an container.

### What do you think would need to be added

From my understanding of the forum topic stateful IPv6 support for OCI images.

davidbockelman:
Hello, I'm a student at UT in Professor Chidambaram's Virtualization course. Could me and my partner take on this issue?

stgraber:
For sure!

davidbockelman:
Hello, our professor has advised us that this issue is harder than we initially thought, so we have chosen different issues to work on. I hope this doesn't cause you any inconvenience.

stgraber:
Okay. I unassigned you.

For anyone else looking at this one, it should actually be pretty simple.
We have a very minimal DHCP client in `cmd/incusd/main_forknet.go` (skip through all the C stuff, it's not relevant to the DHCP client, it's the `RunDHCP` struct and functions you want).

The current DHCPv4 client is implemented by using github.com/insomniacslk/dhcp which has support for both IPv4 and IPv6 DHCP client, so wiring up a basic stateful DHCPv6 client should be pretty straightforward unless there's something particularly different from their equivalent DHCPv4 client.

rahafjrw:
hi! I'm a student at UT Austin's virtualization course and my partner and I were hoping to take this issue on. Can this be assigned to us? thank you!

orcadennis:
I'm in the group with rahafjrw, could I also be assigned this issue?

stgraber:
Done!

rahafjrw:
hello! I wanted to clarify--how do we know whether to set up a DHCPv4 client or a DHCPv6 clients? Specifically, in RunDHCP line 308, "client, err := nclient4.New(iface)" is executed. How do we determine whether to use nclient4 vs nclient6?
Thank you in advance!

stgraber:
We don't really know, so we'll just run both.

stgraber:
[...]

PR #2060, opened 2025-05-02
Implement stateful DHCPv6

rahafjrw:
This PR addresses #1869

The main differences between nclient4 and nclient6 is that:
- nclient6 does not have a Lease struct, but rather works with dhcpv6.Message (the final Reply in the Solicit -> Advertisement -> Request -> Reply conversation)
- It doesn't seem like router information is provided in the Reply, so that's not added in the network configuration
- There is no Renew() function in nclient6 like there is in nclient4, so a Renew message is manually created and sent

### DIFF
--- a/cmd/incusd/main_forknet.go
+++ b/cmd/incusd/main_forknet.go
@@ -479,6 +479,9 @@ func (c *cmdForknet) dhcpRunV4(errorChannel chan error, iface string, hostname s
 }
 
 func (c *cmdForknet) dhcpRunV6(errorChannel chan error, iface string, hostname string, logger *logrus.Logger) {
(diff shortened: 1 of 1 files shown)


stgraber:
Looking into this one now.

stgraber:
I fixed a couple of minor issues with error handling, error messages and the formatting of the IPv6 address when passed to iproute2. I also spent quite a bit of time implementing stateless DHCPv6 too as that's what our setup provides by default so I figured it was worth the effort :)

This was a bit tricky to do as the dhcpv6 client package doesn't really support it, forcing us to mangle/hand-craft the request, but in the end it all worked well and I've validate both stateful and stateless.

stgraber (approved):

Issue #1527, opened 2024-12-18
File operations on filesystem custom storage volumes


stgraber:
It would be good to have a similar file transfer API on the custom volumes as we have on instances, allowing for easy file push/pull to those volumes.

We should be able to extract and share most of the logic we have for the instances but will obviously need to be careful around symlinks, id mapping and the usual issues they bring in.

Aryan470:
Hi, I am in a virtualization course at UT Austin, and we would like to tackle this issue. Could we be assigned to it?

stgraber:
Done!

lahariguduru:
Hi, I am also a part of this group -- can I be assigned to this issue as well?

EeshaKandukuri:
Hi! I am in this group too, can I also be assigned?

stgraber:
Are you still working on this one?

Aryan470:
Unfortunately we haven’t had the time to work on this. Please feel free to reassign it.

PR #2061, opened 2025-05-02
Added API Endpoint and Logic for File Storage Volumes 

lahariguduru:
This PR is in contribution of issue #1527. Most of the work done here closely follows the instance file logic; however, we are not sure if we are going in the right direction completely with this work and still need to test the logic. Any guidance/suggestions would be greatly appreciated!


stgraber:
Looking into this one now.

stgraber:
Replaced by #2081

PR #2068, opened 2025-05-02
Port network_load_balancers to database generator

### LINKED ISSUE #1800: Port `network_load_balancers` to the database generator

davidbockelman:
Created `internal/server/db/cluster/networks_load_balancers.go` which allows for the network_load_balancers SQL code to be auto-generated with `make update-schema`. Changed the call sites to functions in `internal/server/db/network_load_balancers.go` to use the new auto-generated functions. Functions in `internal/server/db/network_network_load_balancers.go` are no longer used.
Fixes #1800


tonyn10:
Hi @davidbockelman . 

Your pull request has been super helpful to me, guiding me on how to tackle the database generator issue I was assigned. I'm working on `network_forwards` and it seems pretty analogous to yours.

One thing I think you missed is calling `CreateNetworkLoadBalancerConfig` after each time you call `CreateNetworkLoadBalancer`. With the existing framework, I don't think creating a network load balancer automatically creates a config. You would need to do something like how `network_integrations` handles it.

As I'm working through mine and comparing against yours, I'll let you know if I find anything else.

cmd/incusd/network_allocations.go
```diff
@@ -219,9 +219,30 @@ func networkAllocationsGet(d *Daemon, r *http.Request) response.Response {
...
+				dbLoadBalancers, err := dbCluster.GetNetworkLoadBalancers(ctx, tx.Tx())
+				if err != nil {
+					return err
+				}
 
-				return err
+				loadBalancers = make(map[int64]*api.NetworkLoadBalancer)
+
+				for _, lb := range dbLoadBalancers {
+					// memberSpecific filtering
```
tonyn10, line 232: Shouldn't it be `if memberSpecific ...`? Also, isn't `memberSpecific` false in this case so no additional filtering needs to be done?

internal/server/network/driver_common.go
```diff
@@ -1142,10 +1142,56 @@ func (n *common) getExternalSubnetInUse(ctx context.Context, tx *db.ClusterTx, u
...
+			loadBalancers, err := dbCluster.GetNetworkLoadBalancers(ctx, tx.Tx(), dbCluster.NetworkLoadBalancerFilter{
+				NetworkID: &networkID,
+			})
+			if err != nil {
+				return nil, fmt.Errorf("Failed loading network load balancer listen addresses: %w", err)
+			}
+			for _, lb := range loadBalancers {
+				// memberSpecific requirements
+				if !memberSpecific || !lb.NodeID.Valid || (lb.NodeID.Valid && lb.NodeID.Int64 == tx.GetNodeID()) {
+					if projectNetworksLoadBalancersOnUplink[projectName] == nil {
```
tonyn10, line 1182: Initialize `projectNetworksLoadBalancersOnUplink = make(map[string]map[string][]string)`

internal/server/network/driver_common.go
```diff
@@ -1142,10 +1142,56 @@ func (n *common) getExternalSubnetInUse(ctx context.Context, tx *db.ClusterTx, u
 	}
 
 	// Get all network load balancer listen addresses for all networks (of any type) connected to our uplink.
-	projectNetworksLoadBalancersOnUplink, err = tx.GetProjectNetworkLoadBalancerListenAddressesByUplink(ctx, uplinkNetworkName, memberSpecific)
+	// all network names for all project names
+	networksByProjects, err := tx.GetNetworksAllProjects(ctx)
 	if err != nil {
-		return nil, fmt.Errorf("Failed loading network forward listen addresses: %w", err)
+		return nil, fmt.Errorf("Failed loading network load balancer listen addresses: %w", err)
```
tonyn10, line 1148: I'll be using this logic for retrieving the network forward listen addresses too, so maybe we can have some of these `return nil, err`?

internal/server/network/driver_ovn.go
```diff
@@ -5757,7 +5797,20 @@ func (n *ovn) LoadBalancerCreate(loadBalancer api.NetworkLoadBalancersPost, clie
...
+				Int64: tx.GetNodeID(),
+			}
+			network_id := n.ID()
+			lb := dbCluster.NetworkLoadBalancer{
+				NetworkID: network_id,
+				NodeID: node_id,
+				ListenAddress: loadBalancer.ListenAddress,
+				Description: loadBalancer.Description,
+				Backends: loadBalancer.Backends,
+				Ports: loadBalancer.Ports,
```
tonyn10, line 5811: I think you should check these two fields if they're null or not, and assign Backends/Ports to []. Otherwise, I believe this messes up the JSON marshalling and unmarshalling.

Issue #1804, opened 2025-03-18
Allow providing certificates as part of preseed data


stgraber:
It would be quite convenient to be able to directly load trusted certificates through the preseed file. This should be pretty straightforward to implement by adding a `Certificates` member to `InitLocalPreseed` (shared/api/init.go) and making the configuration logic load any entry found in there.

DinglyCoder:
Hi, @parthgupta2005 and I are students at the University of Texas at Austin. As part of our virtualization class, we would like to contribute to LXC Incus and resolve this issue.

stgraber:
Assigned it to you now. I'll need @parthgupta2005 to comment in here so I can assign them too.

parthgupta2005:
Hello, apologies for the late response. We look forward to resolving the issue!

parthgupta2005:
Hello, we have a preliminary approach for handling the issue, but have a few confusions. The way we understand it, we would change the InitLocalPreseed struct to include a certificate. Since 'admin_init_preseed:41' automatically reads in data to the respective fields, we wouldn't have to change the logic that parses the pre-seed file. After the struct has been created with the certificate, how should we change the configuration logic to handle the certificate? Sorry for the many questions, thank you so much.

stgraber:
Yep, that's right:
 - Add `Certificates []CertificatePost `json:"certificates" yaml:"certificates"` to `shared/api/init.go`
 - Update `ApplyServerPreseed` in `client/incus_server.go` to go through those entries and call `CreateCertificate` for each of those.

PR #2078, opened 2025-05-05
Allow providing certificates as part of preseed data #1804

DinglyCoder:
Allows directly loading trusted certificates through the preseed file. Added Certificates field to InitLocalPreseed and updated config logic to populate field with entries.


DinglyCoder:
@stgraber, Our implementation seems to be failing the specific test: test_clustering_image_refresh in "Tests / System (tip, cluster, dir, ubuntu-22.04-arm)". We're unsure what could be causing this issue since our updated code shouldn't modify the test image. Is this a concern that we should look into further?

stgraber (approved):

PR #2083, opened 2025-05-06
Adding support for "physical" instances with bridge parent

### LINKED ISSUE #1735: Support attaching instances to a "physical" network that's using a bridge as parent
The host system provides a bridge interface that is connected to a physical interface.

When integrating that bridge via the `physical` nic type and passing the bridge interface as the `parent`I get the following error when starting the VM.

```
Failed to start device "eth-1": Failed to get PCI device info for "br-lan": open /sys/class/net/br-lan/device/uevent: no such file or directory
```

This is because the bridge does not have a `device` directory, but the code requires it for looking up PCIe information.

https://github.com/lxc/incus/blob/36701b95b6c1d83ce6d71798670988d6cf580cdf/internal/server/device/nic_physical.go#L211

The documentation claims I can pass a bridge:

> The physical network type connects to an existing physical network, which can be a network interface or a bridge, and serves as an uplink network for OVN.

https://linuxcontainers.org/incus/docs/main/reference/network_physical/

This issue is reproducible on incus 6.10.1.

NathanChase22:
resolves #1735 

`nic_bridged` logic has been ported over to `nic_physical` with bridge attachment behavior guarded by a check for a parent that's a managed bridge

**Before:**
Screenshot 2025-05-06 at 02 49 04

**After:**
Screenshot 2025-05-06 at 12 05 54


NathanChase22:
Have found a better solution that's more streamlined and runs on up-to-date code

PR #2084, opened 2025-05-07
ported network_peers to database generator

### LINKED ISSUE #1801: Port `network_peers` to the database generator

oronila:
Closes #1801


stgraber:
Okay, I had to do a bunch more changes to the generator functions to get this to behave.
I still need to run this through our OVN testsuite to make sure that the peer logic actually functions.

stgraber:
Fixed a bunch more regressions, now clearing our OVN testsuite so hopefully we're good to go with this one now.

PR #2085, opened 2025-05-07
Port network_forwards to the database generator

### LINKED ISSUE #1799: Port `network_forwards` to the database generator

tonyn10:
Closes #1799


NathanChase22:
I can handle the code style errors today

stgraber:
Did a first pass of cleanups though this won't quite build yet as I need to write a bit more logic and resolve the server-specific logic.

Then it will be time to send this through both system tests and our separate OVN testsuite.

stgraber:
Did a whole bunch of cleanups and aligned the logic with load balancers.
CI tests are passing, OVN tests are passing and manual test for some of the edge cases also look good now.

Issue #2054, opened 2025-05-01
Option to configure the default console for instances


ecapson:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

Currently the console is not configurable by instance basis and not possible to add arguments without a workaround on the OS. for ex: incus console instance --type=vga will launch spicy or remote-viewer. 

### What do you think would need to be added

It would be nice to be able to define default console behaviour in the instance config.
ex yaml config with incus config edit instance: 
console.type = console or vga
console.executable = /usr/bin/somespice-bin
console.args = [somespice-bin args]

stgraber:
This wouldn't be an instance config as those are server-side and it makes no sense for the server side config to keep track of client side preferences.

But we're adding support for more "default" options in the CLI so that could fit in there.

efficacy38:
hello @stgraber, I think I can take this issue.

My current thought is to add a config under `XDG_CONFIG_HOME/incus/console.yml`, which has the following structure. 

```
  virtual-machine: vga
  container: console
  override:
      - project: deafult
        instance: c1
        type: vga
      - project: deafult
        instance: v1
        type: console
```

Does it make sense?

stgraber:
Containers can only do text console anyway, so the option is only really relevant for virtual machines. I also would probably stay away from doing per-VM settings at this stage, I think just a couple of global knobs will be enough here.

We have the new `defaults` section in `config.yml` that we should use for that and can then introduce:
 - `console_prefer_vga bool`
 - `console_vga_command string` (takes a string where the SOCKET keyword is replaced with the path to the SPICE socket)

ecapson:
Yeah, it's only relevant to the VM's and global settings would be enough flexibility for users. 
That'd be awesome. Thanks!

MaheshPunjabi:
@stgraber Can I try?

stgraber:
sure!

stgraber:
@MaheshPunjabi are you still working on this?

MaheshPunjabi:
Sorry got sidetracked last time. I have a couple of follow up questions, will post them here soon.

MaheshPunjabi:
@stgraber Follow up question- when you get a chance, can you suggest where I can find some example strings for the spice socket spec.

Draft plan:
1. Add the 2 vars in shared/cliconfig/default.go under this struct:
// DefaultSettings are the configurations for the Config Struct.
type DefaultSettings struct {
	// Default flag format for list commands.
[...]

PR #2086, opened 2025-05-07
console start, solving issue #2054

kevans1124:


stgraber:
This PR includes a whole bunch of random files that definitely shouldn't be part of it.

A quick scan through also shows that unlike what I mentioned in the issue, this is using server-side configuration keys rather than using the CLi config.

PR #2087, opened 2025-05-07
Restore, solving issue #1928

kevans1124:


stgraber:
This looks like a copy of the other PR which I just closed.

Same issue with including a whole bunch of other files and also using the wrong approach for the console command handling.

Also worth mentioning that the commits aren't titled in a way that describes what they're doing, nor are they signed off. Please read CONTRIBUTING.md

PR #2105, opened 2025-05-10
Use our own QMP monitor

nanjj:
`digitalocean/go-qemu/qmp` depends on `digitalocean/go-qemu` and `digitalocean/go-libvirt` as well, incus does not use anything related with go-libvrit, and the qmp part less maintaining, and do not associate response id with command id. So remove it.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,6 @@ require (
 	github.com/cenkalti/backoff/v4 v4.3.0
 	github.com/checkpoint-restore/go-criu/v6 v6.3.0
 	github.com/cowsql/go-cowsql v1.22.0
-	github.com/digitalocean/go-qemu v0.0.0-20250212194115-ee9b0668d242
 	github.com/digitalocean/go-smbios v0.0.0-20180907143718-390a4f403a8e
 	github.com/dustinkirkland/golang-petname v0.0.0-20240428194347-eebcea082ee0
 	github.com/flosch/pongo2/v6 v6.0.0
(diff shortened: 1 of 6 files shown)


nanjj:
@stgraber I have another two PRs are blocking on this one, please help review.

stgraber:
I've been trying to review this PR a few times over the past few days and even spent a couple of hours trying to chop it into pieces but not very successfully.

This PR is basically a mix of various un-needed changes that break internal API combined with moving away from go-qmp and adding tests for the new implementation.

This really needs to be cut down into multiple commits so we have a chance to review this.

I would basically expect:
 - One commit that adds the in-house QMP implementation (preferably as new files in the existing package)
 - One commit which adds tests on top of that
 - One commit which cuts over from the old go-qmp to the in-house implementation
 - One commit that updates go.mod and go.sum (removing go-qmp)

Changes that really shouldn't be part of this:
 - Renaming of variables/structs/arguments (this just makes reviewing more painful, e.g. logPath -> logFile, requestArgs -> req, ...)
 - Changing of existing function signatures (e.g. RunJSON)
 - Removal of existing feature, moving them under an opt-in mechanism (e.g. logging now requiring INCUS_QMP_EVENTS_DEBUG)

nanjj:
@stgraber only the bytes copy issue being fixed in this refactor. The other fixes like ID association and error handling will raise new PR.

stgraber (approved):

PR #2109, opened 2025-05-12
Add support for bpf token delegation

### LINKED ISSUE #2099: BPF Tokens support

Hello,

as of now, interacting with BPF inside containers is impossible. 

Some versions ago (later than 5.0, but earlier than the current), LXD added support for BPF Tokens. BPF Tokens are a relatively new feature that allow access to privileged containers to some granularly-defined BPF functionality. 

https://docs.ebpf.io/linux/concepts/token/
https://documentation.ubuntu.com/lxd/latest/explanation/bpf/

LXD implements this feature, allowing each container to specify which BPF program types, attach types, etc... can be handled.

Will this feature be implemented in Incus at a certain point?

gwenya:
This adds support for bpf token delegation (closes #2099).

TODOs:
 - [ ] better documentation, right now it's a bit barebones, but I'm not really sure what to add either
 - [ ] tests if the CI can support them
 - [ ] validation if possible: the delegation options are kernel strings and which ones are available depend on the kernel version, hardcoding the current allowed set seems like a bad idea; if validation is impossible then document this fact
 - [ ] improve failure mode, right now if the delegation options are wrong or the feature is not supported by the kernel the forkbpf hook fails and prevents instance startup; this is recorded in lxc.log but the specific error message only shows up when running incus with the `--debug` flag
 - [ ] kernel support check? token delegation is a pretty new feature, it's not supported yet on standard ubuntu installations, so maybe we should check if it's available and issue according error messages, and document the minimum kernel version and config
 - [ ] clean up the forkbpf code
 - [ ] restrict to projects with `low_level` access


cmd/incusd/main_forkbpf.go
```diff
@@ -0,0 +1,296 @@
...
+	int fds[2];
+
+	pidstr = getenv("LXC_PID");
+	if (!pidstr) {
+		fprintf(stderr, "No LXC_PID in environment\n");
+		_exit(EXIT_FAILURE);
+	}
+
+	if (socketpair(AF_UNIX, SOCK_STREAM, 0, fds) < 0) {
+		fprintf(stderr, "Failed to create socke pair: %s", strerror(errno));
```
irhndt, line 54: Typo socke -> socket

irhndt (commented):
Just took a look to learn and found a typo.

cmd/incusd/main_forkbpf.go
gwenya: thanks!

gwenya:
@stgraber would it be possible / make sense for lxc to capture stderr from hooks and log it at a higher level than debug?

cmd/incusd/main_forkbpf.go
```diff
@@ -0,0 +1,285 @@
...
+				_, _ = fmt.Fprint(os.Stdout, "[parent]: Child process failed")
+				os.Exit(1)
+			}
+
+		}
+	}
+}
+
+func (c *cmdForkbpf) runChild(socket int, mountPath string) error {
+	fsfd, err := unix.Fsopen("bpf", 0)
```
gwenya, line 167: I think if this call fails, and the parent has already reached the `Recvmsg()` call, it will hang forever. 
I'm not quite sure how to best work around this.

Maybe the parent should call both waitpid and recvmsg in goroutines and send their results through channels so it can select on them?

cmd/incusd/main_forkbpf.go
```diff
@@ -0,0 +1,284 @@
...
+		snprintf(path, sizeof(path), "/proc/%s/ns/user", pidstr);
+		if (dosetns_file(path) < 0) {
+			_exit(EXIT_FAILURE);
+		}
+
+		// Attach to the user namespace.
+		snprintf(path, sizeof(path), "/proc/%s/ns/mnt", pidstr);
+		if (dosetns_file(path) < 0) {
+			_exit(EXIT_FAILURE);
+		}
```
cyphar, line 79: I guess you probably want to use `pidfd_nsfd`+`change_namespaces`, as well as pidfds if possible (just like `forkproxy` does)?

gwenya:
@stgraber do you know if we can test this in the github runners? I have a feeling that they might not get enough privileges. Apart from that, I think this could be ready, the TODO points I listed would be nice to have of course but not mandatory I think, and I currently don't have the time to figure them out.

stgraber:
> @stgraber do you know if we can test this in the github runners? I have a feeling that they might not get enough privileges. Apart from that, I think this could be ready, the TODO points I listed would be nice to have of course but not mandatory I think, and I currently don't have the time to figure them out.

The tests run as root inside of the Github runner VMs, so that should be fine as far as permissions.

gwenya:
Alright then I'll try to come up with some tests for this.
Do they have unrestricted internet connectivity? I will probably need to use libbpf-bootstrap and download it from git within the runner or instance.

stgraber:
Yep, no internet restrictions from the runners, so you can download and build a test then push it into a test container easily enough.

stgraber:
@gwenya ping

gwenya:
Sadly didn’t get around to this yet, was more busy with work recently. I hope I’ll get around to it soon, but it feels like a somewhat tricky thing to figure out so it might actually take a while.
I think I should first figure out how to run the tests locally, they don’t work on my system but I should be able to do it in a VM.

stgraber:
@gwenya did you have some time to look at this one again?

gwenya:
Sorry for not responding in so long, I will give it a try now

stgraber:
@gwenya great to have you looking into this again!

gwenya:
@stgraber do you have an idea what might cause this test failure? I have never seen that error message ("ERROR: Output buffer space exceeded") before and google only finds me an unconfirmed bug report in the `file` utility...

I'm not sure if it even makes sense to build that tool in the tests, we could instead build it once and host it somewhere to download, or just wait until bpftool tags a new release that includes the `bpftool token` subcommands.

I also appreciate input on the other open points in the task list, I'm not coming up with any good solutions for them and am tempted to just document those shortcomings and leave them be for now.

stgraber:
@gwenya the main issue was that the function was printing quite a bit of stuff to stdout when the caller was expecting the function to only print out a path.

I've re-shuffle things a bit now so the function will instead write the binary to a fixed spot so we can use stdout for normal output.

stgraber:
@gwenya is this ready for review now that the tests are behaving?

gwenya:
@stgraber yes it is ready for review.

PR #2127, opened 2025-05-20
Add +invtsc to cpu extensions ensure tsc as current clocksource

nanjj:


internal/server/instance/drivers/driver_qemu.go
```diff
@@ -1627,6 +1627,9 @@ func (d *qemu) start(stateful bool, op *operationlock.InstanceOperation) error {
 			cpuType = "kvm64"
 			cpuExtensions = append(cpuExtensions, cpuFlags...)
 		}
+	} else {
```
stgraber, line 1630: This still needs to check that `migration.stateful` isn't true as the check above is for both clustered + migration.stateful

nanjj: fixed

stgraber (approved):

PR #2130, opened 2025-05-20
Feature: Added support for "physical" interface with bridge as parent

### LINKED ISSUE #1735: Support attaching instances to a "physical" network that's using a bridge as parent

NathanChase22:
fixes #1735 

After checking if parent property points to a bridge network, we create and instantiate a bridged device which we use to call `Start()` and `Stop()`. 

Meaning we are using 'bridged' logic already implemented, avoiding attempts to attach the parent interface into the container namespace which was the cause of previous errors.

This is the log showing success in launching a container following [these steps](https://github.com/lxc/incus/issues/1735#issuecomment-2799170436) :

```
Name: c1
Description: 
Status: RUNNING
Type: container
Architecture: x86_64
PID: 270002
Created: 2025/05/20 22:48 UTC
Last Used: 2025/05/20 23:26 UTC
Started: 2025/05/20 23:26 UTC

Resources:
  Processes: 10
  CPU usage:
    CPU usage (in seconds): 0
  Memory usage:
    Memory (current): 20.75MiB
  Network usage:
    eth0:
      Type: broadcast
      State: UP
      Host interface: veth57af7d0c
      MAC address: 32:da:df:6b:fc:d6
      MTU: 1500
      Bytes received: 300B
      Bytes sent: 904B
      Packets received: 2
      Packets sent: 5
      IP addresses:
        inet6: fe80::a046:6bff:fe0a:156d/64 (link)
    lo:
      Type: loopback
      State: UP
      MTU: 65536
      Bytes received: 0B
      Bytes sent: 0B
      Packets received: 0
      Packets sent: 0
      IP addresses:
        inet:  127.0.0.1/8 (local)
        inet6: ::1/128 (local)
```

### DIFF
--- a/internal/server/device/nic_physical.go
+++ b/internal/server/device/nic_physical.go
@@ -180,6 +180,25 @@ func (d *nicPhysical) Start() (*deviceConfig.RunConfig, error) {
 		return nil, err
 	}
 
+	// Handle the case where the parent is a bridge.
+	isParentBridge := util.PathExists(fmt.Sprintf("/sys/class/net/%s/bridge", d.config["parent"]))
+	if isParentBridge {
+		// Convert the device to a nictype=bridged internally.
+		bridgedConfig := d.config.Clone()
+		bridgedConfig["type"] = "nic"
+		bridgedConfig["nictype"] = "bridged"
+		bridgedConfig["network"] = ""
+
+		// Instantiate the new device.
+		bridged, err := load(d.inst, d.state, d.inst.Project().Name, d.inst.Name(), bridgedConfig, d.volatileGet, d.volatileSet)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to initialize bridged device: %w", err)
+		}
+
+		// Forward the start call.
+		return bridged.Start()
+	}
+
 	// Lock to avoid issues with containers starting in parallel.
 	networkCreateSharedDeviceLock.Lock()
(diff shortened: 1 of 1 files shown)


NathanChase22:
Both tests failed at the setup stage, is it possible to rerun?

NathanChase22:
> Not worth re-running since I'll most likely be making a few changes to this PR before merging it. I just ran out of time before the Incus 6.13 release, but I'll be looking at the remaining open PRs over the next few days.

What sort of changes do you have in mind?

stgraber (approved):

PR #2138, opened 2025-05-23
Make sure limits.memory <= root,size.state

nanjj:


internal/server/instance/drivers/driver_qemu.go
```diff
@@ -1176,7 +1176,7 @@ func (d *qemu) checkStateStorage() error {
 	}
 
 	if stateDiskSize < memoryLimit {
-		return errors.New("Stateful stop and snapshots require that the instance limits.memory is less than size.state on the root disk device")
+		return fmt.Errorf("Stateful stop and snapshots require that the instance limits.memory is less than or equal to size.state on the root disk device")
```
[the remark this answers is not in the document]
nanjj: done

internal/server/instance/drivers/driver_qemu.go
```diff
@@ -1176,7 +1176,7 @@ func (d *qemu) checkStateStorage() error {
 	}
 
 	if stateDiskSize < memoryLimit {
-		return errors.New("Stateful stop and snapshots require that the instance limits.memory is less than size.state on the root disk device")
+		return errors.New("Stateful stop and snapshots require the instance limits.memory should less than or equal to the root disk size.state")
```
stgraber, line 1179: Hmm, that doesn't read right.

It should probably be:
```
Stateful stop and snapshots require that the instance limits.memory be less than or equal to the root disk size.state property
```

nanjj: done

stgraber (approved):

PR #2148, opened 2025-05-27
incusd/storage: fix squashfs unpacking to NFS destinations

mrstux:
NFS4v2 supports user.xattrs, and can be used with Incus via the dir driver, but image unpacking will fail when the container image uses non-user xattrs such as security.xattr

This fix will skip extraction of security.xattrs when extracting to an NFS hosted directory

The fix is not specific to the dir driver, and could benefit a future NFS driver (#2025)

In my testing, privileged containers can successfully be started from NFS hosted dir storage pools.

### DIFF
--- a/internal/linux/filesystem.go
+++ b/internal/linux/filesystem.go
@@ -42,6 +42,16 @@ func DetectFilesystem(path string) (string, error) {
 	return FSTypeToName(int32(fs.Type))
 }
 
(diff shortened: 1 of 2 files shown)


stgraber:
Superseded by #2149. I'd normally push minor fixes directly to the branch but that doesn't typically work when the origin repository belongs to an organization (as is the case here with truenas).

PR #2157, opened 2025-05-29
Add TrueNAS Storage Driver

mrstux:
The TrueNAS Storage driver allows an Incus node to utilise a remote TrueNAS storage server to host one or more incus storage pools. 

When the node is in a cluster, then all members of the cluster can use the storage pool simultaneously. This works well , for example, when live migrating a VM from one Incus node to another.

The driver is block-based, ie all Incus volumes are created as zvols on the remote server, and the zvol block devices are accessed on the local incus node via iscsi

The driver is modelled after the existing ZFS driver, and supports most existing ZFS functionality, but on a remote TrueNAS server. 

For example, a local VM can be snapshotted and then cloned. The snapshot and cloning will be performed on the remote server after synchronizing the local filesystem, and the clone will then be activated via iSCSI as necessary.

The driver uses the `truenas_incus_ctl` tool, which provides a simple CLI interface to the TrueNAS API to effect changes on the remote server. Additionally, the tool can also activate/deactivate remote zvols using open-iscsi.

_If the tool is not found, then the driver is disabled._

The latest version of the truenas tool can be obtained from : https://github.com/truenas/truenas_incus_ctl, and must be installed in the PATH supplied to incusd (v0.7.2+ is required)

`open-iscsi` can be installed with `apt install open-iscsi`

To login to the remote TrueNAS host first run the command:

`sudo truenas_incus_ctl config login`

Example:

```
$ sudo truenas_incus_ctl config login
[sudo] password for user: 
Enter a name for this connection: example
Enter the TrueNAS hostname or IP address: server.example.com
Setting up connection to TrueNAS host: server.example.com
Choose authentication method (1 for API Key, 2 for Username/Password): 2
Testing connection to wss://server.example.com/api/current...
Enter your TrueNAS username: truenas_admin
Enter your TrueNAS password: 
Generating API key...
API key successfully generated
Successfully connected to wss://server.example.com/api/current
Configuration for 'example' (connecting to server.example.com) saved to /root/.truenas_incus_ctl/config.json
```

After logging in, ensure iSCSI is configured and working with the following command:

`sudo truenas_incus_ctl share iscsi setup --test`

As an example of the functionality of the TrueNAS tool, this command can be used to list all datasets on the remote server

```
sudo truenas_incus_ctl dataset ls
```
 
 Once logged in and iSCSI has been configured, the following command can be used to create a remote storage pool

`incus storage create <poolname> truenas source=[host:]<pool>/<dataset>/[remote-poolname]`

`source` is used to describe the location where the storage pool will be created on the remote TrueNAS host
`host` is optional, and `remote-poolname` will be set to `poolname` if not supplied.

`truenas.config` can be used to specify a specific login config
`truenas.host` can be used to specify a specific host, which the tool will lookup in the config.json

**Running Tests**

There are a number of `INCUS_TRUENAS_` environment variables that can  be used to configure the remote TrueNAS that will be used by the driver during the test. The only environment variable that is required is `INCUS_TRUENAS_DATASET`

I use the following commands in a script called `start-tests.sh` to run the suite using the default host configured in `truenas_incus_ctl` config file

`sudo -E env PATH=$PATH INCUS_BACKEND=truenas INCUS_TRUENAS_DATASET=tank/se/incus-tests ./main-tn.sh $1`

Alternatively, I specify a `config` for a different host. Hosts and API keys can also be specified manually to avoid the use of a config file. Alternate config files can be specified.

`sudo -E env PATH=$PATH INCUS_VERBOSE=0 INCUS_TMPFS=0 INCUS_BACKEND=truenas INCUS_TRUENAS_DATASET=puddle/se/incus-tests INCUS_TRUENAS_CONFIG=goldeye2 ./main-tn.sh $1`

I run the standalone tests using `./start-tests standalone`, as the `cluster` tests require net ns support from `truenas_incus_ctl` which is not implemented.

`main-tn.sh` is the same as `main.sh`, except some tests which are problematic on my testing hosts are disabled.

`main-tn.sh` should probably be deleted once the `./main.sh standalone` tests are verified

When running the test `truenas_incus_ctl` will be launched in daemon mode. It can be useful to prelaunch the daemon to view its console as the test suite runs.

for example:
`sudo /home/<user>/go/bin/truenas_incus_ctl daemon /home/<user>/tncdaemon.sock`

**Incus CI Integration**

I would suggest that a VM with a "golden" image of GoldEye with the middleware defer patch is used as a target. The `truenas_incus_ctl` tool should be built from master and the standalone tests used. A preconfigured config.json for the tool could be used to provide the host/api-key to use for the tests. A container dataset should be created and removed after the test suite is run.

**Known Issues**

- *Creating and Deleting remote Volumes is relatively slow* due to current TrueNAS bottlenecks. Work is ongoing on resolving and hopefully will be incorporated in a future builds of Goldeye. https://github.com/truenas/middleware/pull/16614
- Requires TrueNAS Goldeye for `snapshot rename` capability (currently in Nightlies), although there is a patch that can be easily installed in Fangtooth (25.04) https://github.com/truenas/truenas_incus_ctl#zfssnapshotrename-backport
- Cluster tests are not currently passing (requires net ns work-arounds), but cluster usage is supported. Use `standalone` tests
- TrueNAS hosts don't currently support zvol-shrinking via API. The driver attempts to zvol-shrink via the tool, and then ignores the inevitable "shrink failure", but warns with the correct `zfs set` command required. If the API is ever changed, the tool will be updated to add support.
- Optimized migrations between storage pools are not currently implemented.

### DIFF
--- a/internal/server/storage/drivers/driver_truenas_volumes.go
+++ b/internal/server/storage/drivers/driver_truenas_volumes.go
@@ -36,7 +36,7 @@ func (d *truenas) CreateVolume(vol Volume, filler *VolumeFiller, op *operations.
 
 	if vol.contentType == ContentTypeFS {
 		// Create mountpoint.
-		err := vol.EnsureMountPath()
+		err := vol.EnsureMountPath(true)
 		if err != nil {
 			return err
 		}
@@ -268,7 +268,7 @@ func (d *truenas) CreateVolume(vol Volume, filler *VolumeFiller, op *operations.
 		if vol.contentType == ContentTypeFS {
 			// Run EnsureMountPath again after mounting and filling to ensure the mount directory has
 			// the correct permissions set.
-			err := vol.EnsureMountPath()
+			err := vol.EnsureMountPath(true)
 			if err != nil {
 				return err
 			}
@@ -330,7 +330,7 @@ func (d *truenas) createOrRefeshVolumeFromCopy(vol Volume, srcVol Volume, refres
 
 	if vol.contentType == ContentTypeFS {
 		// Create mountpoint.
-		err = vol.EnsureMountPath()
+		err = vol.EnsureMountPath(false)
 		if err != nil {
 			return err
 		}
@@ -506,7 +506,7 @@ func (d *truenas) createOrRefeshVolumeFromCopy(vol Volume, srcVol Volume, refres
 
 		// Mount the volume and ensure the permissions are set correctly inside the mounted volume.
 		err := vol.MountTask(func(_ string, _ *operations.Operation) error {
-			return vol.EnsureMountPath()
+			return vol.EnsureMountPath(false)
 		}, op)
 		if err != nil {
 			return err
@@ -543,7 +543,7 @@ func (d *truenas) CreateVolumeFromCopy(vol Volume, srcVol Volume, copySnapshots 
 func (d *truenas) CreateVolumeFromMigration(vol Volume, conn io.ReadWriteCloser, volTargetArgs localMigration.VolumeTargetArgs, preFiller *VolumeFiller, op *operations.Operation) error {
 	if volTargetArgs.ClusterMoveSourceName != "" && volTargetArgs.StoragePool == "" {
 		d.logger.Debug("Detected migration between cluster members on the same storage pool")
-		err := vol.EnsureMountPath()
+		err := vol.EnsureMountPath(false)
 		if err != nil {
 			return err
 		}
@@ -1284,7 +1284,7 @@ func (d *truenas) MountVolume(vol Volume, op *operations.Operation) error {
 	case ContentTypeFS:
 		mountPath := vol.MountPath()
 		if !linux.IsMountPoint(mountPath) {
-			err := vol.EnsureMountPath()
+			err := vol.EnsureMountPath(false)
 			if err != nil {
 				return err
 			}
@@ -1476,7 +1476,7 @@ func (d *truenas) CreateVolumeSnapshot(vol Volume, op *operations.Operation) err
 	}
 
 	// Create snapshot directory.
-	err = vol.EnsureMountPath()
+	err = vol.EnsureMountPath(false)
 	if err != nil {
 		return err
 	}
@@ -1649,7 +1649,7 @@ func (d *truenas) MountVolumeSnapshot(snapVol Volume, op *operations.Operation) 
 		mountPath := snapVol.MountPath()
 		l.Debug("Content type FS", logger.Ctx{"mountPath": mountPath})
 		if !linux.IsMountPoint(mountPath) {
-			err := snapVol.EnsureMountPath()
+			err := snapVol.EnsureMountPath(false)
 			if err != nil {
 				return err
 			}


william-gr:
Launching a new instance with a TrueNAS without iSCSI configure says:

```
Launching u1
Error: Failed instance creation: Failed creating instance from image: Failed to run: truenas_incus_ctl --host 192.168.1.10 share iscsi create tank/incusstorage/images/eb7ec38152a187619b1004f15f922a44cbbc11e46c6443bd119a6bd74854877f_ext4 --target-prefix=incus: exit status 1 (Error: 
No iSCSI portal was found for this host. Use:
<truenas_incus_ctl> share iscsi portal create --ip <IP address> --port <port number>
To create one.
```

<truenas_incus_ctl> share iscsi portal create --ip <IP address> --port <port number>
suggestions seems wrong as that command is not valid.

william-gr:
Logging in using the truenas incus ctl does not seem to work without a SSL certificate:

```
~/go/bin/truenas_incus_ctl config login --allow-insecure
Enter a name for this connection: test
Enter the TrueNAS hostname or IP address: 192.168.1.10
Setting up connection to TrueNAS host: 192.168.1.10
Choose authentication method (1 for API Key, 2 for Username/Password): 2
Testing connection to wss://192.168.1.10/api/current...
Error: Failed to create connection to wss://192.168.1.10/api/current: failed to connect: tls: failed to verify certificate: x509: cannot validate certificate for 192.168.1.10 because it doesn't contain any IP SANs
Usage:
  truenas_incus_ctl config login [flags]

```

mrstux:
> Logging in using the truenas incus ctl does not seem to work without a SSL certificate:
> 
> ```
> ~/go/bin/truenas_incus_ctl config login --allow-insecure
> Enter a name for this connection: test
> Enter the TrueNAS hostname or IP address: 192.168.1.10
> Setting up connection to TrueNAS host: 192.168.1.10
> Choose authentication method (1 for API Key, 2 for Username/Password): 2
> Testing connection to wss://192.168.1.10/api/current...
> Error: Failed to create connection to wss://192.168.1.10/api/current: failed to connect: tls: failed to verify certificate: x509: cannot validate certificate for 192.168.1.10 because it doesn't contain any IP SANs
> Usage:
>   truenas_incus_ctl config login [flags]
> ```

This bug has been fixed in truenas_incus_ctl now

https://github.com/truenas/truenas_incus_ctl/pull/40/commits

mrstux:
> After manually creating the portal:
> 
> ```
> william@thinkpad[~]: ~/go/bin/incus launch -s tnpool images:debian/12 u1
> Launching u1
> Error: Failed instance creation: Failed creating instance from image: No path for activated TrueNAS volume: tank/incusstorage/images/eb7ec38152a187619b1004f15f922a44cbbc11e46c6443bd119a6bd74854877f_ext4
> ```
> 
> I had to manually start iSCSI service, a hint would have been great.

I've added a hint to the instructions at the top of this post, and there is an [issue](https://github.com/truenas/truenas_incus_ctl/issues/42) to add iscsi service control support to the tool. Once that's added I'll include the exact incantation to ensure the iscsi service is created.


> 
> After that:
> 
> ```
> Error: Failed instance creation: Failed creating instance from image: Failed to run: truenas_incus_ctl --host 192.168.1.10 share iscsi create tank/incusstorage/images/eb7ec38152a187619b1004f15f922a44cbbc11e46c6443bd119a6bd74854877f_ext4 --target-prefix=incus: exit status 1 (Error: 
> 
> Error -32602
> Invalid params
> [EINVAL] iscsi_target_create.groups.0.initiator: 1 Initiator not found in database)
> ```
> 
> I tried creating initiator group but got same error.
> 
> It seems truenas_incus_ctl is hard-coding initiator id 1.

Enhanced [Portal](https://github.com/truenas/truenas_incus_ctl/issues/44) + [Initiator](https://github.com/truenas/truenas_incus_ctl/issues/43) support is being implemented in the Tool. 

Once this is implemented, portal+initiator creation/usage should be automatic, but the driver has properties that will allow overriding the automatic selections (tool support is required)

jbendtsen:
The issues described in this thread should all be fixed, but more testing will be conducted to confirm.

mrstux:
```
./start-test standalone

...

==> Test result: success
```

mrstux:
@stgraber I don’t believe the Linstor failure is related to this PR

stgraber:
Yeah, it's not, we're seeing it everywhere now, something must have changed in Linstor.

stgraber:
Looking into this one now.

Because it originated from an organization rather than an individual account, I won't be able to work on the commits in-place. For now my goal is to go through all the changes and extract anything that's not directly part of the TrueNAS driver and then send those separately for immediate inclusion.

Once that's done, then the next step is a rebase of the branch on top of that and individual review of all the remaining changes (unless I start by squashing a bunch of commits together).

Because we like reviewing changes commit by commit, branches that have commits significantly altering the content of a previous commit in the same branch tend to waste a lot of review time, so I'll try to have a quick look see if that's the case here (given the very high commit count) and if so, instead start squashing things together and re-committing in more reviewable units.

stgraber:
@mrstux What's the reason for requiring `truenas_incus_ctl`?

Given the tool is itself also written in Go, wouldn't it be possible to consume the logic as a Go package?

It's not really a blocker to merging this, but as we're spending quite a bit of time lately working on Incus OS which is an extremely locked down platform where users cannot get a local shell nor install extra 3rd party tools, it may make it harder to properly support the TrueNAS driver in those kind of environments.

doc/howto/storage_pools.md
```diff
@@ -150,6 +150,50 @@ Use the existing Ceph Object Gateway `https://www.example.com/radosgw` to create
...
+truenas.api_key | Allows specifying the `api-key` flag when performing TrueNAS operations
+truenas.blocksize | blocksize to use when creating remote block devices
+truenas.host | Allows specifying the `host` flag when performing TrueNAS operations
+truenas.config | Allows specifying the `config` flag when performing TrueNAS operations
+truenas.config-file | Allows specifying the `config-file` flag when performing TrueNAS operations
+truenas.dataset | The remote dataset on the TrueNAS host
+truenas.initiator | Allows specifying the "initiator" flag when performing TrueNAS iSCSI operations
+truenas.portal  | Allows specifying the "initiator" flag when performing TrueNAS iSCSI operations
+truenas.remove_snapshots | Remove snapshots as needed
+truenas.use_refquota |  Use `refquota` instead of `quota` for space
```
stgraber: Some of those key descriptions aren't exactly useful :)

Basically any of the `Allows specifying the XYZ flag` type descriptions really don't tell anything useful to the user.

doc/howto/storage_pools.md
```diff
@@ -150,6 +150,50 @@ Use the existing Ceph Object Gateway `https://www.example.com/radosgw` to create
...
+-|-
+source | can be used to specify `truenas.dataset` and optionally `truenas.host`
+block.filesystem | Filesystem to use when formatting a filesystem block device
+block.mount_options | Mount options to use when mounting a filesystem block device
+truenas.allow-insecure | Allows specifying the `allow-insecure` flag when performing TrueNAS operations
+truenas.api_key | Allows specifying the `api-key` flag when performing TrueNAS operations
+truenas.blocksize | blocksize to use when creating remote block devices
+truenas.host | Allows specifying the `host` flag when performing TrueNAS operations
+truenas.config | Allows specifying the `config` flag when performing TrueNAS operations
+truenas.config-file | Allows specifying the `config-file` flag when performing TrueNAS operations
```
stgraber: As we work with clusters and now with immutable hosts, we should really stay away from anything that expects a config file to be present on the host as that's very difficult to ensure is identical across all servers in a cluster and is downright impossible to put in place when running on Incus OS or similar locked down environments.

doc/.wordlist.txt
```diff
@@ -323,10 +331,10 @@ UFW
...
 UID
 UIDs
+uncomment
 unconfigured
 unevictable
 unixgram
-uncomment
 unmanaged
 unmount
 unmounting
```
stgraber: This PR adds the following entries to the word list:
- api
- blocksize
- config
- filesystem
- iSCSI
- refquota
- truenas
- TrueNAS

Of those, the only ones that I think are reasonable to add are:
- iSCSI
- TrueNAS

The others should either use their long form equivalents ("block size", "file system", "configuration") or if they refer to a specific key name or internal construct, they should be marked as such like `refquota`.

stgraber:
I'd rather not have `main-tn.sh` in here, this looks like it's mostly a copy of `main.sh` but with a bunch of tests commented out. Instead if some tests are unsuitable for TrueNAS, they should have skipping logic directly within them.

test/main.sh
```diff
@@ -220,7 +223,7 @@ if [ "${1:-"all"}" != "cluster" ]; then
     run_test test_tls_jwt "JWT authentication"
 fi
 
-if [ "${1:-"all"}" != "standalone" ]; then
+if [ "${1:-"all"}" != "standalone" ] && [ "${INCUS_BACKEND}" != "truenas" ]; then
```
stgraber: Our test selection logic is quite messy right now, but instead of this, I think I'd prefer a check be added earlier in the "Available storage backends" logic to instantly fail if the testsuite is called with INCUS_BACKEND=truenas and `${1} != "standalone`, effectively explicitly requiring that the testsuite be run in standalone mode when testing truenas.

test/suites/basic.sh
```diff
@@ -693,6 +694,10 @@ test_basic_usage() {
     PID=$(incus info c1 | awk '/^PID/ {print $2}')
     kill -9 "${PID}"
     sleep 3
+      # FIXME: truenas volume create/delete is slow (known issue)
+    if [ "$incus_backend" = "truenas" ]; then
+      sleep 1
+    fi
```
stgraber: Is that still needed? If so, should this just be bumped to 5s to avoid the issue?

test/suites/basic.sh
```diff
@@ -242,6 +242,7 @@ test_basic_usage() {
   ! incus init testimage aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa || false
 
   # Test snapshot publish
+  echo "==> Test snapshot publish"
```
stgraber: What's that for?

test/suites/container_move.sh
```diff
@@ -13,9 +13,12 @@ test_container_move() {
   incus project create "${project}"
   if [ "${incus_backend}" = "linstor" ]; then
       incus storage create "${pool2}" "${incus_backend}" linstor.resource_group.place_count=1
+  elif [ "$incus_backend" = "truenas" ]; then
+    incus storage create "${pool2}" "${incus_backend}" "$(truenas_source)/$(uuidgen)" "$(truenas_config)" "$(truenas_config_file)" "$(truenas_api_key)"
   else
-      incus storage create "${pool2}" "${incus_backend}"
+    incus storage create "${pool2}" "${incus_backend}"
```
stgraber: Why the change of indent?

test/suites/snapshots.sh
```diff
@@ -364,7 +364,14 @@ test_snap_schedule() {
...
   incus info c1 | grep -q snap1
 
-  incus rm -f c1 c2 c3 c4 c5
+  if [ "$incus_backend" != "truenas" ]; then
+    incus rm -f c1 c2 c3 c4 c5
+  else
+    # FIXME: truenas create/delete is very slow, this causes issues with the default 120s timeout when doing mass instance deletes.
+    incus rm -f c1 c2
+    incus rm -f c3 c4
+    incus rm -f c5
```
stgraber: At that point, I'd just do a for loop and run them sequentially. `incus rm` internally runs the deletes sequentially anyways.

But I also have to ask, what exactly is going on with performance here?

If deleting 5 instances is taking over 2 minutes, that's really horrible performance and we definitely need a warning in the storage documentation telling folks to stay away from TrueNAS for anything that needs semi-frequent instance/volume creation.

Looking at my production clusters, I have a bunch of CI type jobs which commonly create a 50+ instances, runs for a couple minutes and deletes them. And production users running Terraform will pretty commonly create dozens if not hundreds of instances+volumes in parallel.

test/suites/storage.sh
```diff
@@ -15,9 +15,11 @@ test_storage() {
   storage_pool="incustest-$(basename "${INCUS_DIR}")-pool"
   storage_volume="${storage_pool}-vol"
   if [ "${incus_backend}" = "linstor" ]; then
-      incus storage create "$storage_pool" "$incus_backend" linstor.resource_group.place_count=1 --description foo
+    incus storage create "$storage_pool" "$incus_backend" linstor.resource_group.place_count=1 --description foo
```
stgraber: Another indentation change

test/suites/storage.sh
```diff
@@ -887,9 +889,11 @@ test_storage() {
   INCUS_DIR="${INCUS_DIR}"
   storage_pool="incustest-$(basename "${INCUS_DIR}")-pool26"
   if [ "${incus_backend}" = "linstor" ]; then
-      incus storage create "$storage_pool" "$incus_backend" linstor.resource_group.place_count=1
+    incus storage create "$storage_pool" "$incus_backend" linstor.resource_group.place_count=1
```
stgraber: Identation

test/suites/storage.sh
```diff
@@ -887,9 +889,11 @@ test_storage() {
...
+  elif [ "${incus_backend}" = "truenas" ]; then
+    incus storage create "$storage_pool" "$incus_backend" "$(truenas_source)/$(basename "${INCUS_DIR}")-pool26" "$(truenas_config)" "$(truenas_config_file)" "$(truenas_api_key)"
   else
-      incus storage create "$storage_pool" "$incus_backend"
+    incus storage create "$storage_pool" "$incus_backend"
```
stgraber: Indentation

test/suites/storage_snapshots.sh
```diff
@@ -19,10 +19,13 @@ test_storage_volume_snapshots() {
   storage_volume="${storage_pool}-vol"
 
   if [ "${incus_backend}" = "linstor" ]; then
-      incus storage create "$storage_pool" "$incus_backend" linstor.resource_group.place_count=1
+    incus storage create "$storage_pool" "$incus_backend" linstor.resource_group.place_count=1
+  elif [ "${incus_backend}" = "truenas" ]; then
+    incus storage create "$storage_pool" "$incus_backend" "$(truenas_source)/" "$(truenas_config)" "$(truenas_config_file)" "$(truenas_api_key)"
   else
-      incus storage create "$storage_pool" "$incus_backend"
+    incus storage create "$storage_pool" "$incus_backend"
```
stgraber: Indentation

test/suites/storage_snapshots.sh
```diff
@@ -19,10 +19,13 @@ test_storage_volume_snapshots() {
   storage_volume="${storage_pool}-vol"
 
   if [ "${incus_backend}" = "linstor" ]; then
-      incus storage create "$storage_pool" "$incus_backend" linstor.resource_group.place_count=1
+    incus storage create "$storage_pool" "$incus_backend" linstor.resource_group.place_count=1
```
stgraber: Indentation

stgraber:
Okay, so this branch is thankfully already nicely isolated as far as what it changes in our codebase. I found a couple of small things I could send up already which are now here: https://github.com/lxc/incus/pull/2329

For the rest, it mostly boils down to:
 - Configuration tweaks, documentation is pretty unclear on those and we should remove anything that expects the user to put configuration files in place outside of what can be done through the Incus API
 - A few small documentation tweaks to use long form terminology and/or quote internal terms
 - Dropping the duplicate testsuite
 - Cleanup of logic to skip the cluster tests
 - Eliminate unneeded indentation changes in the testsuite (yeah, I know our indentation is inconsistent, we should probably fix it, but that should be done as a single standalone commit re-indenting the whole thing automatically, for a PR like this, we want to just go with how things currently are and limit changes)

On the questions front, I have:
 - What's going on with create/delete performance. I see mentions in the testsuite of some operations taking several minutes. That seems like a rather major issue for anyone wanting to seriously use this. Can you elaborate on what's going on here and whether it's a temporary situation?
 - What's the reason for using the standalone command line tool rather than consume it as a Go package? One thing I just noticed now would be the license being an issue as it's GPLv3 which wouldn't be allowed as an import in Incus, but that's something you presumably have control over on your side and so could go with a more permissive license. The dependency on a 3rd party tool is going to severely reduce adoption of this driver as every distribution will now need to make a package for that tool.


I'm yet to actually test the driver myself but things look quite good overall, particularly in keeping this change nicely isolated from any wider Incus change (making the risk of inclusion pretty low).

mrstux:
> @mrstux What's the reason for requiring `truenas_incus_ctl`?
> 
> Given the tool is itself also written in Go, wouldn't it be possible to consume the logic as a Go package?

There were a few reasons.

1) minimise the blast radius of implementing the driver into the Incus architecture. Essentially 3 minor modifications + the driver files.
2) not require additional Incus PRs as the TrueNAS API/OS evolves (any changes/optimizations can be implemented in the tool)
3) minimize the logic required in the Incus driver. Most of the complexity is dealt with by the tool.
4) model the driver on the zfs driver
5) re-use the implementation with other hypervisor plugins / CSs
6) the tool is genuinely useful as is

> nor install extra 3rd party tools

Its worth pointing out that open-iscsi is also required to be installed.

The driver self-disables if the tool is not found in the path.

doc/howto/storage_pools.md
mrstux: > Some of those key descriptions aren't exactly useful :)
> 
> Basically any of the `Allows specifying the XYZ flag` type descriptions really don't tell anything useful to the user.

I take your point. They do actually refer to the flags in the tool's documentation. I can flesh out the documentation further.

doc/howto/storage_pools.md
mrstux: Specifying a config-file path is optional.

The tool can support specifying a host + api-key, or you can specify a host or config, and the api-key will be looked up in the local config-file (which may not exist), and you can additionally specify an alternate location for the config-file.

The config/config file mechanism works well for interactive use, but the main reason to utilize it when using Incus is to prevent the api-key being visible in the properties.

Perhaps an api-key secret path would be a more appropriate solution?

test/suites/basic.sh
mrstux: ok

doc/.wordlist.txt
mrstux: will fix

test/suites/basic.sh
mrstux: Its debug detritus. Will remove.

test/suites/container_move.sh
mrstux: Actually... I didn't ;)

Most of the test files use 2 space indents, and "linstor" support was added *after* this code was written and used 4 spaces.

Would you prefer 2 or 4 spaces here?

stgraber:
Thanks for the explanation. We already have the iSCSI tools in Incus OS and adding the extra tool for TrueNAS storage will be quite easy for us.

I suspect where things will be more painful is for other distros that have their own native Incus packages as now each of those is going to have to first package the extra tool before their packaged version of Incus can make use of the driver.

doc/howto/storage_pools.md
stgraber: We already have this logic for the storage pools:
```

if !hasEditPermission(auth.ObjectStoragePool(poolName)) {
    // Don't allow non-admins to see pool config as sensitive info can be stored there.
    poolAPI.Config = nil
}
```

So in restricted environments, less privileged users can't read the pool config.
Anyone who can read the pool config has enough privilege on Incus that they can just escalate to full root access anyway.

doc/howto/storage_pools.md
stgraber: Thanks!

test/suites/snapshots.sh
mrstux: Yes. What is happening is it can take up to 30 seconds to remove a zvol/iscsi export on the server via the API.

There is an open PR on the middleware to improve.

https://github.com/truenas/middleware/pull/16614

More diagnosis is ongoing. It appears there is a 4 second timeout/refresh in the iscsid on the server.

And because of the way the middleware is currently processing deletes multiples of this timeout are triggered on deletion. (this is what the PR addresses)

In theory, it should be close to instant for a delete, and the fix should be transparent to Incus.

I'll convert to a loop.

test/suites/storage.sh
mrstux: again, it was actually the linstor PR which changed the indention. The rest of the file mostly, but not entirely, uses 2 space indention.

I can undo.

test/suites/container_move.sh
stgraber: Going to standardize things a bit with https://github.com/lxc/incus/pull/2331
So it'll be 4 spaces moving forward.

test/suites/snapshots.sh
stgraber: Okay, good to hear that things should improve soon then :)

doc/howto/storage_pools.md
mrstux: I can remove the config-file key, which will simply things. Its unused by the tests iirc.

test/main.sh
mrstux: but `${1}` can be the name of any suite, not just "all", "cluster" or "standalone"

mrstux:
> I suspect where things will be more painful is for other distros that have their own native Incus packages as now each of those is going to have to first package the extra tool before their packaged version of Incus can make use of the driver.

Yes, but once the tool is installed, the driver will light up. Apologies, I forget if the `incusd` needs to be restarted in this case. I do know that updating the tool does not require the daemon to be restarted, which is neat.

It would be nice to have packages for the tool, as mentioned, it is genuinely useful outside of use with Incus.

mrstux:
> One thing I just noticed now would be the license being an issue as it's GPLv3 which wouldn't be allowed as an import in Incus

If that were a problem, at this stage I think a relicense could be arranged, as there are no 3rd party contributions. It would be up to @kmoore134

stgraber:
I believe we try to initialize all the pool drivers on daemon startup and record their availability and version at that point, so if the TrueNAS tool is installed after the fact, an Incus daemon restart will be required.

stgraber:
For the test indentation stuff, I marked them all as resolved as the tests/ change will need to be rebased on https://github.com/lxc/incus/pull/2331 once that lands in main in a few hours, at which point we'll be on a consistent 4 spaces indent.

mrstux:
For future reference, I discuss the current create/delete performance issue in this comment, which is now outdated:

https://github.com/lxc/incus/pull/2157#discussion_r2244241349

test/main.sh
mrstux: I see what you meant.

mrstux:
@stgraber 

How do I fix the `requota` / `truenas` spelling issues in storage_pools.md?

It looks like its correctly escaped to me.

```
Misspelled words:
<htmlcontent> doc/html/howto/storage_pools/index.html: html>body>div>div>div>div>article>section>section>section>div>div>div>table>tbody>tr>td>p
--------------------------------------------------------------------------------
refquota
truenas
```

kmoore134:
> > One thing I just noticed now would be the license being an issue as it's GPLv3 which wouldn't be allowed as an import in Incus
> 
> If that were a problem, at this stage I think a relicense could be arranged, as there are no 3rd party contributions. It would be up to @kmoore134

Yes, this does not need to be GPLv3, we can re-license to match the Incus license as needed. Stuart, would you be willing to submit the PR for changing that?

mrstux:
@stgraber any other changes required?

stgraber:
I'm getting ready to push it. I've done the commit squashing, just left with the doc being in the wrong place (should have its own storage_truenas matching the rest) and just noticed an inconsistency in config key naming, so will rename `allow-insecure` to `allow_insecure` in the doc. Looks like the code is already correct.

stgraber:
@mrstux
```
Error: Failed to create storage pool on TrueNAS host: test/d13, err: Failed to run: truenas_incus_ctl --allow-insecure --api-key 1-uKc41yiT4T97RDa6YsfaxLiH2f2hZ2KEgKfaNljgTGot5nvF6OJlYyg0MNUhz1rh --host truenas01.default.lxc.shf.stgraber.net dataset create --managedby incus.truenas --comments Managed by Incus.TrueNAS test/d13: exit status 1 (2025/08/02 03:28:40 $HOME is not defined)
```

Why does it need $HOME to be set?
Most users run the Incus daemon through systemd so you're not going to have all the usual environment variables you'd have in an interactive session.

stgraber:
We also primarily test on Debian these days. Debian 13 has no such thing as "iscsitarget" as a system service, so the tool doesn't work particularly well in our environment.

stgraber:
The error you're getting when iscsi isn't setup on the TrueNAS side is also pretty unclear and will likely lead to a bunch of users trying to start iscsid on their system when it's in fact referring to the service not running on the TrueNAS side.

stgraber:
Ah and our lab infra is accessible only over IPv6, which seems to have some issues?

```
root@d13:~# incus storage create foo truenas source=test/d13 truenas.allow_insecure=true truenas.host=truenas01.default.lxc.shf.stgraber.net truenas.api_key=1-uKc41yiT4T97RDa6YsfaxLiH2f2hZ2KEgKfaNljgTGot5nvF6OJlYyg0MNUhz1rh
Error: Failed to run: truenas_incus_ctl --allow-insecure --api-key 1-uKc41yiT4T97RDa6YsfaxLiH2f2hZ2KEgKfaNljgTGot5nvF6OJlYyg0MNUhz1rh --host truenas01.default.lxc.shf.stgraber.net share iscsi test --target-prefix=incus --parsable --setup: exit status 1 (Error: 

Error -32602
Invalid params
[EINVAL] iscsiportal_create.listen: IP 2602:fc62:b:8004:1266:6aff:fed0:1cf4 not configured on this system.)
```

```
root@truenas:~# ip -6 a show dev enp5s0
2: enp5s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    inet6 2602:fc62:b:8004:1266:6aff:fed0:1cf4/64 scope global dynamic mngtmpaddr proto kernel_ra 
       valid_lft forever preferred_lft forever
    inet6 fe80::1266:6aff:fed0:1cf4/64 scope link proto kernel_ll 
       valid_lft forever preferred_lft forever
```

So the IPv6 address is definitely correct but we can't get the iscsi connection to establish.

stgraber:
Okay, that's as far as I can go with testing this.

I could workaround the $HOME thing easily enough, but I can't really workaround the thing about iSCSI not working over IPv6 on TrueNAS. We don't have an IPv4 ingress into the lab environment where the TrueNAS instance is running, so we need IPv6 to work correctly.

mrstux:
@stgraber will look into resolving

mrstux:
> @stgraber will look into resolving

I fixed a couple of IPv6 related bugs in the tool, and IPv6 now works with the unmodified driver (using truenas.host)

An additional fix was needed for the “source” parsing, and I also needed to add INCUS_TRUENAS_ALLOW_INSECURE to tests, to allow using direct IPv6 with self signed certificates when testing. 

This means some tweaks to the tests are needed. 

Expect to have this finished tomorrow, but how do you want the commits? Just appended to this branch and you’ll cherry pick?

@stgraber

stgraber:
Yeah, you can append to this branch and I'll cherry-pick over into mine.

mrstux:
> Yeah, you can append to this branch and I'll cherry-pick over into mine.

I've pushed the commits. Also, requires the main branch of the truenas_incus_ctl tool for IPv6 compatibility (v0.7.3 not released yet)

I had to make an improvement to the auto-restart testing in order to pass reliably, and also removed some of the exclusions which are no longer needed.

Still running tests on my side.

stgraber:
Okay, I'll try to do another test tonight by building the truenas_incus_cli from main and run with the changes above.

mrstux:
> > Yeah, you can append to this branch and I'll cherry-pick over into mine.
> 
> I've pushed the commits. Also, requires the main branch of the truenas_incus_ctl tool for IPv6 compatibility (v0.7.3 not released yet)
> 
> I had to make an improvement to the auto-restart testing in order to pass reliably, and also removed some of the exclusions which are no longer needed.
> 
> Still running tests on my side.

Found an issue.

stgraber:
@mrstux I built `truenas_incus_ctl` from the main branch, but I'm still getting the exact same failures here:

```
root@d13:~# incus storage create test truenas source=test truenas.allow_insecure=true truenas.host=truenas01.default.lxc.shf.stgraber.net truenas.api_key=2-QfvqynvZbjrNuMWPpyYudDyixt9DBX7iPbAwyzSP9cbpVwBKJdsJq1gCjD5kAGAY

Error: Failed to create storage pool on TrueNAS host: test/test, err: Failed to run: truenas_incus_ctl --allow-insecure --api-key 2-QfvqynvZbjrNuMWPpyYudDyixt9DBX7iPbAwyzSP9cbpVwBKJdsJq1gCjD5kAGAY --host truenas01.default.lxc.shf.stgraber.net dataset create --managedby incus.truenas --comments Managed by Incus.TrueNAS test/test: exit status 1 (2025/08/08 03:48:37 $HOME is not defined)
```

And then after putting a temporary workaround for that HOME env variable, I'm getting:
```
root@d13:~# incus storage create test truenas source=test truenas.allow_insecure=true truenas.host=truenas01.default.lxc.shf.stgraber.net truenas.api_key=2-QfvqynvZbjrNuMWPpyYudDyixt9DBX7iPbAwyzSP9cbpVwBKJdsJq1gCjD5kAGAY
Error: Failed to run: truenas_incus_ctl --allow-insecure --api-key 2-QfvqynvZbjrNuMWPpyYudDyixt9DBX7iPbAwyzSP9cbpVwBKJdsJq1gCjD5kAGAY --host truenas01.default.lxc.shf.stgraber.net share iscsi test --target-prefix=incus --parsable --setup: exit status 1 (Error: 

Error -32602
Invalid params
[EINVAL] iscsiportal_create.listen: IP 2602:fc62:b:8004:1266:6aff:fed0:1cf4 not configured on this system.)
```

mrstux:
> @mrstux I built `truenas_incus_ctl` from the main branch, but I'm still getting the exact same failures here:
> 
> ```
> root@d13:~# incus storage create test truenas source=test truenas.allow_insecure=true truenas.host=truenas01.default.lxc.shf.stgraber.net truenas.api_key=2-QfvqynvZbjrNuMWPpyYudDyixt9DBX7iPbAwyzSP9cbpVwBKJdsJq1gCjD5kAGAY
> 
> Error: Failed to create storage pool on TrueNAS host: test/test, err: Failed to run: truenas_incus_ctl --allow-insecure --api-key 2-QfvqynvZbjrNuMWPpyYudDyixt9DBX7iPbAwyzSP9cbpVwBKJdsJq1gCjD5kAGAY --host truenas01.default.lxc.shf.stgraber.net dataset create --managedby incus.truenas --comments Managed by Incus.TrueNAS test/test: exit status 1 (2025/08/08 03:48:37 $HOME is not defined)
> ```
> 
> And then after putting a temporary workaround for that HOME env variable, I'm getting:
> 
> ```
> root@d13:~# incus storage create test truenas source=test truenas.allow_insecure=true truenas.host=truenas01.default.lxc.shf.stgraber.net truenas.api_key=2-QfvqynvZbjrNuMWPpyYudDyixt9DBX7iPbAwyzSP9cbpVwBKJdsJq1gCjD5kAGAY
> Error: Failed to run: truenas_incus_ctl --allow-insecure --api-key 2-QfvqynvZbjrNuMWPpyYudDyixt9DBX7iPbAwyzSP9cbpVwBKJdsJq1gCjD5kAGAY --host truenas01.default.lxc.shf.stgraber.net share iscsi test --target-prefix=incus --parsable --setup: exit status 1 (Error: 
> 
> Error -32602
> Invalid params
> [EINVAL] iscsiportal_create.listen: IP 2602:fc62:b:8004:1266:6aff:fed0:1cf4 not configured on this system.)
> ```

Can you share the results of running:

`
truenas_incus_ctl -allow-insecure --api-key 2-QfvqynvZbjrNuMWPpyYudDyixt9DBX7iPbAwyzSP9cbpVwBKJdsJq1gCjD5kAGAY --host truenas01.default.lxc.shf.stgraber.net share iscsi portal ls
`

Also the results of `ip a` on the TrueNAS host please

stgraber:
```
root@d13:~# truenas_incus_ctl --allow-insecure --api-key 2-QfvqynvZbjrNuMWPpyYudDyixt9DBX7iPbAwyzSP9cbpVwBKJdsJq1gCjD5kAGAY --host truenas01.default.lxc.shf.stgraber.net share iscsi portal ls
root@d13:~# 
```

```
root@truenas:~# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host noprefixroute 
       valid_lft forever preferred_lft forever
2: enp5s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 10:66:6a:d0:1c:f4 brd ff:ff:ff:ff:ff:ff
    inet 10.224.199.13/24 brd 10.224.199.255 scope global dynamic enp5s0
       valid_lft 2656sec preferred_lft 2656sec
    inet6 2602:fc62:b:8004:1266:6aff:fed0:1cf4/64 scope global dynamic mngtmpaddr proto kernel_ra 
       valid_lft forever preferred_lft forever
    inet6 fe80::1266:6aff:fed0:1cf4/64 scope link proto kernel_ll 
       valid_lft forever preferred_lft forever
root@truenas:~# 
```

mrstux:
I think I see the problem.

The tool is assuming a static IP (via IP Alias) has been set on the host. TrueNAS API will only allow an IP Alias to be used for explicit Portal creation.

I suspect if you added the "2602:fc62:b:8004:1266:6aff:fed0:1cf4/64" address as an IP alias on the interface in the network setting UI it may work.

There seems to be further missing support for IPv6 and portal creation in the tool. 

I'm not sure if the tool supports the "::" Portal. 

There is support for specifying a portal to use in the tool's command line, and that is exposed via the `truenas.portal` property. But is not exposed to the test suite currently (ie `INCUS_TRUENAS_PORTAL`).

Ie, it may work if you add a portal manually for "::" and specify "[::]" as the portal, but there may be further IPv6 parsing issues. (EDIT, specifying the portal by id would bypass the parsing issue, `-p, --portal string          iSCSI portal [ip]:[port] or id (default ":")`)

I'll look further into this.

stgraber:
I'm not seeing any IP alias setting in https://truenas01.default.lxc.shf.stgraber.net/ui/system/network

image

mrstux:
> I'm not seeing any IP alias setting in https://truenas01.default.lxc.shf.stgraber.net/ui/system/network
> 
> image

Need to start with this, then you should be able to add IP Aliases

IMG_3671

stgraber:
That seems like a very poor experience for IPv6 users.

With IPv6, your routers emit router advertisements with the IPv6 address then being derived from your MAC address. So that's effectively as good as a static IP address. Having to duplicate that into a static IP config feels pretty wrong.

There also doesn't appear to be a way to leave IPv4 on DHCP but then add a static alias for IPv6 (if you really need that).

stgraber:
Anyway, moving to a static IPv6 address got me a bit further.

I was able to create a storage pool:
```
root@d13:~# incus storage create test truenas source=test truenas.allow_insecure=true truenas.host=truenas01.default.lxc.shf.stgraber.net truenas.api_key=2-QfvqynvZbjrNuMWPpyYudDyixt9DBX7iPbAwyzSP9cbpVwBKJdsJq1gCjD5kAGAY
Storage pool test created
```

However creating a VM on it isn't working:
```
root@d13:~# incus launch images:debian/13 d13 --vm 
Launching d13
Error: Failed instance creation: Failed creating instance from image: No path for locate-activated TrueNAS volume: test/test/images/73b6d277b346f2361f8c2b98db61ae9c45c4bdc052607bf61480846fe58784ea_ext4
```

mrstux:
> root@d13:~# incus storage create test truenas source=test truenas.allow_insecure=true truenas.host=truenas01.default.lxc.shf.stgraber.net truenas.api_key=2-QfvqynvZbjrNuMWPpyYudDyixt9DBX7iPbAwyzSP9cbpVwBKJdsJq1gCjD5kAGAY
Storage pool test created

Is the name of your pool “test”?

can you try `source=test\`

> That seems like a very poor experience for IPv6 users.
> 

Yes. 

> With IPv6, your routers emit router advertisements with the IPv6 address then being derived from your MAC address. So that's effectively as good as a static IP address. Having to duplicate that into a static IP config feels pretty wrong.

Yes. I’m working on an improvement. 

The current behavior is focused on SAN vs LAN networking, which is why the portal is opening a specific IP. 

TrueNAS’s current behavior of only support IP Aliases for discrete portals is outside of my control

EDIT: I actually filed a [bug](https://ixsystems.atlassian.net/browse/NAS-134525) on this back in March 

> 
> There also doesn't appear to be a way to leave IPv4 on DHCP but then add a static alias for IPv6 (if you really need that).

Yes. 

None of these issues actually affect the code of the driver though.

stgraber:
> Is the name of your pool “test”?

Yeah, the pool on the TrueNAS side is named `test` and I already have `source=test` passed in that command.

stgraber:
That's what I've got on the TrueNAS side:
```
test                                                               1.70G  45.8G   104K  /mnt/test
test/.system                                                       1.70G  45.8G  1.66G  legacy
test/.system/configs-ae32c386e13840b2bf9c0083275e7941               684K  45.8G   684K  legacy
test/.system/cores                                                   96K  1024M    96K  legacy
test/.system/netdata-ae32c386e13840b2bf9c0083275e7941              36.8M  45.8G  36.8M  legacy
test/.system/nfs                                                    112K  45.8G   112K  legacy
test/.system/samba4                                                 180K  45.8G   180K  legacy
test/.system/vm                                                      96K  45.8G    96K  legacy
test/test                                                          1008K  45.8G   120K  /mnt/test/test
test/test/containers                                                 96K  45.8G    96K  /mnt/test/test/containers
test/test/custom                                                     96K  45.8G    96K  /mnt/test/test/custom
test/test/deleted                                                   504K  45.8G   120K  /mnt/test/test/deleted
test/test/deleted/containers                                         96K  45.8G    96K  /mnt/test/test/deleted/containers
test/test/deleted/custom                                             96K  45.8G    96K  /mnt/test/test/deleted/custom
test/test/deleted/images                                             96K  45.8G    96K  /mnt/test/test/deleted/images
test/test/deleted/virtual-machines                                   96K  45.8G    96K  /mnt/test/test/deleted/virtual-machines
test/test/images                                                     96K  45.8G    96K  /mnt/test/test/images
test/test/virtual-machines                                           96K  45.8G    96K  /mnt/test/test/virtual-machines
```

mrstux:
> source=test

Ahh, I see, I already fixed that ;)

```
	// a pool... means we create a dataset in the root
	if !strings.Contains(path, "/") {
		path += "/"
	}
```

Please bare with me,

The following truenas_incus_ctl should be run with the following parameters

`truenas_incus_ctl -allow-insecure --api-key 2-QfvqynvZbjrNuMWPpyYudDyixt9DBX7iPbAwyzSP9cbpVwBKJdsJq1gCjD5kAGAY --host truenas01.default.lxc.shf.stgraber.net`

Please run these commands as superuser, and provide the output

```
truenas_incus_ctl dataset create -V 1G test/test/test.block
truenas_incus_ctl share iscsi locate --create --parsable test/test/test.block

ls -l /dev/disk/by-path
truenas_incus_ctl share iscsi target ls
truenas_incus_ctl share iscsi extent ls
truenas_incus_ctl share iscsi delete test/test/test.block
truenas_incus_ctl dataset rm test/test/test.block
iscsiadm -m session
iscsiadm -m node
```

stgraber:
```
root@d13:~# truenas_incus_ctl --allow-insecure --api-key 2-QfvqynvZbjrNuMWPpyYudDyixt9DBX7iPbAwyzSP9cbpVwBKJdsJq1gCjD5kAGAY --host truenas01.default.lxc.shf.stgraber.net dataset create -V 1G test/test/test.block
root@d13:~# truenas_incus_ctl --allow-insecure --api-key 2-QfvqynvZbjrNuMWPpyYudDyixt9DBX7iPbAwyzSP9cbpVwBKJdsJq1gCjD5kAGAY --host truenas01.default.lxc.shf.stgraber.net share iscsi locate --create --parsable test/test/test.block
created	test/test/test.block
root@d13:~# ls -l /dev/disk/by-path
total 0
lrwxrwxrwx 1 root root   9 Aug  8 15:11 pci-0000:02:00.0-scsi-0:0:0:1 -> ../../sda
drwxr-xr-x 6 root root 120 Aug  8 15:11 pci-0000:02:00.0-scsi-0:0:0:1-part
lrwxrwxrwx 1 root root  10 Aug  8 15:11 pci-0000:02:00.0-scsi-0:0:0:1-part1 -> ../../sda1
lrwxrwxrwx 1 root root  10 Aug  8 15:11 pci-0000:02:00.0-scsi-0:0:0:1-part2 -> ../../sda2
root@d13:~# truenas_incus_ctl --allow-insecure --api-key 2-QfvqynvZbjrNuMWPpyYudDyixt9DBX7iPbAwyzSP9cbpVwBKJdsJq1gCjD5kAGAY --host truenas01.default.lxc.shf.stgraber.net share iscsi target ls
 id |         name         |        alias         
----+----------------------+----------------------
 5  | test:test:test-block | test/test/test.block 
root@d13:~# truenas_incus_ctl --allow-insecure --api-key 2-QfvqynvZbjrNuMWPpyYudDyixt9DBX7iPbAwyzSP9cbpVwBKJdsJq1gCjD5kAGAY --host truenas01.default.lxc.shf.stgraber.net share iscsi extent ls
 id |         name         |           disk            
----+----------------------+---------------------------
 5  | test:test:test-block | zvol/test/test/test.block 
root@d13:~# truenas_incus_ctl --allow-insecure --api-key 2-QfvqynvZbjrNuMWPpyYudDyixt9DBX7iPbAwyzSP9cbpVwBKJdsJq1gCjD5kAGAY --host truenas01.default.lxc.shf.stgraber.net share iscsi delete test/test/test.block
deleted	test/test/test.block
root@d13:~# truenas_incus_ctl --allow-insecure --api-key 2-QfvqynvZbjrNuMWPpyYudDyixt9DBX7iPbAwyzSP9cbpVwBKJdsJq1gCjD5kAGAY --host truenas01.default.lxc.shf.stgraber.net dataset rm test/test/test.block
root@d13:~# iscsiadm -m session
iscsiadm: No active sessions.
root@d13:~# iscsiadm -m node
iscsiadm: No records found
root@d13:~#
```

mrstux:
Yes. Its failing to activate, but more so, it appears that the discovery hasn't been performed.

Please try:

```
truenas_incus_ctl share iscsi setup --test
```

stgraber:
```
root@d13:~# truenas_incus_ctl --allow-insecure --api-key 2-QfvqynvZbjrNuMWPpyYudDyixt9DBX7iPbAwyzSP9cbpVwBKJdsJq1gCjD5kAGAY --host truenas01.default.lxc.shf.stgraber.net share iscsi setup --test
Portal ID: 1
Initiator ID: 1
# BEGIN RECORD 2.1.11
discovery.startup = manual
discovery.type = sendtargets
discovery.sendtargets.address = 2602:fc62:b:8004:1266:6aff:fed0:1cf4
discovery.sendtargets.port = 3260
discovery.sendtargets.auth.authmethod = None
discovery.sendtargets.auth.username = <empty>
discovery.sendtargets.auth.password = <empty>
discovery.sendtargets.auth.username_in = <empty>
discovery.sendtargets.auth.password_in = <empty>
node.session.auth.chap_algs = MD5
discovery.sendtargets.timeo.login_timeout = 15
discovery.sendtargets.use_discoveryd = No
discovery.sendtargets.discoveryd_poll_inval = 30
discovery.sendtargets.reopen_max = 5
discovery.sendtargets.timeo.auth_timeout = 45
discovery.sendtargets.timeo.active_timeout = 30
discovery.sendtargets.iscsi.MaxRecvDataSegmentLength = 32768
# END RECORD

root@d13:~# 
```

mrstux:
Try

```
truenas_incus_ctl dataset create -V 1G test/test/test.block
truenas_incus_ctl share iscsi create test/test/test.block
truenas_incus_ctl share iscsi activate test/test/test.block
ls -l /dev/disk/by-path
iscsiadm -m session
iscsiadm -m node
```

stgraber:
```
root@d13:~# truenas_incus_ctl --allow-insecure --api-key 2-QfvqynvZbjrNuMWPpyYudDyixt9DBX7iPbAwyzSP9cbpVwBKJdsJq1gCjD5kAGAY --host truenas01.default.lxc.shf.stgraber.net share iscsi activate test/test/test.block
Error: 
iscsiadm: cannot make connection to 2602:fc62:b:8004:1266:6aff:fed0:1cf4: Connection refused
iscsiadm: cannot make connection to 2602:fc62:b:8004:1266:6aff:fed0:1cf4: Connection refused
iscsiadm: cannot make connection to 2602:fc62:b:8004:1266:6aff:fed0:1cf4: Connection refused
iscsiadm: cannot make connection to 2602:fc62:b:8004:1266:6aff:fed0:1cf4: Connection refused
iscsiadm: cannot make connection to 2602:fc62:b:8004:1266:6aff:fed0:1cf4: Connection refused
iscsiadm: cannot make connection to 2602:fc62:b:8004:1266:6aff:fed0:1cf4: Connection refused
iscsiadm: connection login retries (reopen_max) 5 exceeded
iscsiadm: Could not perform SendTargets discovery: iSCSI PDU timed out
exit status 11
The iscsitarget service is running. It may need to be restarted with:
service restart iscsitarget
```

stgraber:
Fixed it, it was a firewall problem. Would be nice to get a clearer error though ;)

stgraber:
@mrstux so ideally we'd get those two fixed in truenas_incus_ctl then we should be good to go:
 - Make the tool work when $HOME isn't set in the environment (that's how it will typically get called by Incus)
 - Make the tool or Incus return a clearer error when iSCSI can't be established (as was the case above due to firewalling blocking the iSCSI port)

mrstux:
> @mrstux so ideally we'd get those two fixed in truenas_incus_ctl then we should be good to go:
> 
> * Make the tool work when $HOME isn't set in the environment (that's how it will typically get called by Incus)
> * Make the tool or Incus return a clearer error when iSCSI can't be established (as was the case above due to firewalling blocking the iSCSI port)

I believe the $HOME issue has now been fixed in the tool. Additionally, I found that the tool was swallowing the error from discovery, when creating the target.

This would mean in the case where target creation succeeded (because its via WebSocket), yet discovery failed, there would be no error, and activation would be skipped (because there were no discovered targets), thus there was no activation error either. My concern is this may have been because the discovery is prone to failure when its already been performed, but the incus basic_usage tests passed...

Additionally, I added explicit handling in the driver, so that if the tool created the target, but did not activate it, and did not actually error (which it should be now, and wasn't previously) then the driver would suggest that perhaps there is an iscsi communication issue.

Also, I improved the "locate-activate" error.

And finally, the "iscsitarget service" is now described as the "remote iscsitarget service", to hopefully make it more obvious that the driver/tool is referring to the remote TrueNAS host's iscsitarget. Its still confusing, but trying to parse out "helpful" text from the tool in the driver, is probably not a good idea.

The changes have been pushed to tool's master branch.

@stgraber so, please try again :)

mrstux:
And, I applied the EnsureMountPath changes. The neat thing is that the driver and the zfs driver are very similar... so easy to apply the change based on the zfs driver's changes.

PR #2168, opened 2025-06-02
incusd/instance/qemu: Skip invtsc on non-x86 and when running nested

stgraber:


hallyn (approved):

hallyn:
Bit of an unusual way to pass along a boolean :)

Issue #2178, opened 2025-06-05
Copying a storage volume in a cluster fails on local storage if destination is a different cluster member


victoitor:
### Is there an existing issue for this?

- [x] There is no existing issue for this bug

### Is this happening on an up to date version of Incus?

- [x] This is happening on a supported version of Incus

### Incus system details

```yaml
config:
  cluster.https_address: 10.11.16.11:8443
  core.https_address: 10.11.16.11:8443
  instances.placement.scriptlet: "# Error return strings\nerror_cpu_not_string        =
    \"Configuração de limits de CPU não é string.\"\nerror_cpu_syntax            =
    \"Sintaxe inválida de limite de CPU.\"\nerror_cpu_not_pinned        = \"Núcleos
    de CPU não foram fixados.\"\nerror_cpu_key               = \"Chave limits.cpu
    com erro: \"\nerror_pinned_cpu_not_allowed = \"CPUs fixados pela instância não
    são permitidos pelo projeto.\"\n\nerror_mem_not_string        = \"Configuração
    de limite de memória não é string.\"\nerror_mem_syntax            = \"Sintaxe
    inválida de limite de memória. \"\nerror_mem_key               = \"Chave limits.memory
    com erro: \"\nerror_mem_not_fixed         = \"Memória não fixada.\"\nerror_mem_limit_exceeded
    \   = \"Memória acima do permitido pelo projeto.\"\n\nerror_responsible_undefined
    = \"Responsável não definido para instância na chave user.responsavel.\"\nerror_project_config
    \       = \"Erro na configuração do projeto. Por favor fale com um administrador
    do cluster para resolver o problema. Erro: \"\nerror_no_available_node     = \"Não
    há máquina válida disponível.\"\n\n# Parses limits.cpu type strings which should
    have pinned cpus. Returns either an error string or a tuple containing the cpu
    bit vector and the most significant set bit\ndef parse_cpu_pin(cpu_string):\n
    \   cpu_bit_vector  = 0\n    mssb            = -1\n\n    if type(cpu_string) !=
    \"string\":\n        return error_cpu_not_string\n    if cpu_string.isdigit():\n
    \       return error_cpu_not_pinned\n    parts = cpu_string.split(\",\")\n    for
    p in parts:\n        r = p.split(\"-\")\n        if len(r) not in (1, 2):\n            return
    error_cpu_syntax\n        if len(r) == 1:\n            if not r[0].isdigit():\n
[...]

PR #2195, opened 2025-06-13
Remove target check when server clustered

presztak:
When the server is part of a cluster and the source volume is on different target, we don't need to check the `target` parameter. In such cases, `clusterCopyCustomVolumeInternal` should always be used.

That said, when executing the command:
`incus storage volume copy local/bug-test local/bug-test-copy`
if the source volume is located on a different cluster node, the system should always call `clusterCopyCustomVolumeInternal`.

Closes: #2178

### DIFF
--- a/cmd/incusd/storage_volumes.go
+++ b/cmd/incusd/storage_volumes.go
@@ -727,13 +727,11 @@ func storagePoolVolumesPost(d *Daemon, r *http.Request) response.Response {
 		return response.Conflict(errors.New("Volume by that name already exists"))
 	}
 
-	target := request.QueryParam(r, "target")
-
 	// Check if we need to switch to migration
 	serverName := s.ServerName
 	var nodeAddress string
 
-	if s.ServerClustered && target != "" && (req.Source.Location != "" && serverName != req.Source.Location) {
+	if s.ServerClustered && (req.Source.Location != "" && serverName != req.Source.Location) {
 		err := s.DB.Cluster.Transaction(r.Context(), func(ctx context.Context, tx *db.ClusterTx) error {
 			nodeInfo, err := tx.GetNodeByName(ctx, req.Source.Location)
 			if err != nil {


stgraber:
Merging despite the Linstor failure as it's already affecting main.

@bensmrs any idea what's up with Linstor?

bensmrs:
I’m afraid I can’t reproduce locally… and I’m having trouble understanding how this PR could in any way influence the standalone check.

stgraber:
@bensmrs yeah, this PR doesn't that's why I merged it anyway. We're seeing the same failure for main builds prior to this being merged so there's something weird going on with Linstor, at least in our test environment.

bensmrs:
@luissimas, could you see if the standalone tests run on Ubuntu? Works on my Debian™
I’m quite confused by the error though, as I don’t see how snapshot names could possibly get so long (we’re normally only using our prefix + the UUID without dashes). The actual error in all the messy error messages is `Name "linstor_linstor--incus-incus--volume--178c8981584e4a52ad9d8923cfe821c2_00000_safety--snap--394e5244--04f9--480b--a2f8--7fb3791b692c" too long.`

luissimas:
Hey @bensmrs! Sorry for the late reply on this one. Is this still an issue?

This `safety--snap` component of the snapshot name seems like something Linstor specific.

https://github.com/LINBIT/linstor-server/blob/2680d60d0059b44842d60e805b771081c0c656b8/controller/src/main/java/com/linbit/linstor/core/apicallhandler/controller/CtrlSnapshotRollbackApiCallHandler.java#L238-L243

Since this is an internal snapshot that Linstor itself creates, it seems weird that it would hit the 48 characters limit. But IIRC the constraint for the name length is at the database level, so it is possible that this is what's causing the issue.

bensmrs:
Hi!
The issue appeared, then disappeared. It’s one of these transient things with LINSTOR that I like to keep hidden under the rug. I’ll re-ping you when it comes back :D

PR #2227, opened 2025-06-26
generate-database: Add create_timestamp and update_timestamp

breml:
Add create_timestamp and update_timestamp as new statement tags.


masnax (approved):
LGTM, one extension for the future might be to have this work with the `Update` method as well, as right now it seems it only affects `Rename`. That might be a bit trickier to do since the caller is currently expected to provide the date.

breml:
> LGTM, one extension for the future might be to have this work with the `Update` method as well, as right now it seems it only affects `Rename`. That might be a bit trickier to do since the caller is currently expected to provide the date.

@masnax I am not sure, I follow. I covered [create](https://github.com/lxc/incus/pull/2227/files#diff-4e270c7d3896120e1e448908948b9a1a529575fdceabc469460da4de9ad14e1fR896-R900), [update](https://github.com/lxc/incus/pull/2227/files#diff-4e270c7d3896120e1e448908948b9a1a529575fdceabc469460da4de9ad14e1fR1223-R1227) and [rename](https://github.com/lxc/incus/pull/2227/files#diff-4e270c7d3896120e1e448908948b9a1a529575fdceabc469460da4de9ad14e1fR1085-R1098). The difference between update and create on one side and rename on the other is, that only for rename it was necessary to actually alter the query, since for update and create the queries already contain all the columns and I only need to make sure, that the respective timestamps get the correct value, see https://github.com/FuturFusion/operations-center/blob/main/internal/provisioning/repo/sqlite/entities/server.mapper.go#L74-L83 for an example.

So for create and update, the value provided by the caller is basically ignored, since it is always overwritten.

PR #2244, opened 2025-07-03
shared/idmap: Skip ACLs that are out of range

stgraber:
This is typically ACLs that were set from outside of the containers or that have already been shifted. That -1 value would cause the ACL calls to fail, leading to a partly shifted container.


hallyn (approved):

xlmnxp:
@stgraber is this MR will fix https://discuss.linuxcontainers.org/t/mysql-oci-container-use-100-cpu-without-even-working/24052?

PR #2249, opened 2025-07-03
client: request profile scope in oidc device code flow

xanderio:
Without this scope the IdP may not include claims, such as `preferred_username`, which incus expects to be present if configured in `oidc.claim`. This results in an error `Error: not authorized` which is confusing to users.

The requested scopes should be made configurable, maybe based on the `oidc.scopes` server configuration option, but my go skills are non-existent :)

### DIFF
--- a/client/incus_oidc.go
+++ b/client/incus_oidc.go
@@ -81,7 +81,7 @@ func (o *oidcTransport) RoundTrip(r *http.Request) (*http.Response, error) {
 
 var (
 	errRefreshAccessToken = errors.New("Failed refreshing access token")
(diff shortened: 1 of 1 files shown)


stgraber:
Your commit is missing the needed Signed-off-by line.

I also think that this is the wrong approach overall. We have an `oidc.scopes` config option on the server side, so I think what we want is push the list of expected scopes from the server to the client through an HTTP header so we can then have the client request the correct set of scopes.

That way we have a consistent behavior between when using the CLI in device mode and using something like the Incus UI directly with the server.

stgraber:
@xanderio ping
We unfortunately can't review this until the Signed-off-by issue clears.

PR #2253, opened 2025-07-04
Fix TC regressions

stgraber:


stgraber:
@gwenya this gets us limits.egress and limits.ingress working again, at least on my system.

tych0 (approved):

PR #2259, opened 2025-07-06
cmd/incus_agent: Replace gorilla/mux with http.ServeMux

0xk1f0:
Since Go 1.22, we have the ability to create a request multiplexer using `http.ServeMux` with native libraries. This makes the [`gorilla/mux`](https://github.com/gorilla/mux) library obsolete (which also hasn't received any new commits for over a year now).

This PR removes usage of `gorilla/mux` for the `incus_agent` as a start. I do think it would be a wise decision to adopt this for all usages of it in this repository eventually.

### DIFF
--- a/cmd/incus-agent/dev_incus.go
+++ b/cmd/incus-agent/dev_incus.go
@@ -4,13 +4,10 @@ import (
 	"fmt"
 	"net"
 	"net/http"
-	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
 	"time"
-
-	"github.com/gorilla/mux"
 
 	incus "github.com/lxc/incus/v6/client"
 	"github.com/lxc/incus/v6/internal/server/daemon"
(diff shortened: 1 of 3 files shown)


stgraber:
Yeah, we've moved away from gorilla/mux in some of the new projects (incus-os, migration-manager, operations-center) and it's been working well. Makes sense to start doing it here too.

stgraber (approved):

stgraber:
```
 Error: cmd/incus-agent/operations.go:178:10: indent-error-flow: if block ends with a return statement, so drop this else and outdent its block (revive)
		} else {
			timeoutSecs = formTimeout
		}
1 issues:
```

(Note that we have static-analysis configured to be stricter on modified code than on existing code, so it may trigger on code that you didn't introduce, that's normal, it just forces us to progressively fix those things)

0xk1f0:
Fixed that, should be cleaner now.

PR #2260, opened 2025-07-06
Fixed non-constant format string in call to fmt.Errorf

c4t3l:
This PR fixes formatting for string passed to `fmt.Errorf`.  This is inline with the recent golang 1.24 changes to its printf analyzer.  

I am one of the package maintainers for Fedora, incus fails our go vet checks due to this.

### DIFF
--- a/client/incus.go
+++ b/client/incus.go
@@ -474,7 +474,7 @@ func (r *ProtocolIncus) rawWebsocket(url string) (*websocket.Conn, error) {
 			}
 
 			if apiResp != nil && apiResp.Error != "" {
-				err = errors.Join(err, fmt.Errorf(apiResp.Error))
+				err = errors.Join(err, errors.New(apiResp.Error))
 			}
 		}
 


client/incus.go
```diff
@@ -474,7 +474,7 @@ func (r *ProtocolIncus) rawWebsocket(url string) (*websocket.Conn, error) {
 			}
 
 			if apiResp != nil && apiResp.Error != "" {
-				err = errors.Join(err, fmt.Errorf(apiResp.Error))
+				err = errors.Join(err, fmt.Errorf("%v", apiResp.Error))
```
stgraber: That's the wrong way to fix this warning. The error string is constant here, so it should be using `errors.New(apiResp.Error())` instead.

stgraber: The reason for the lint warning is because using `fmt.Errorf` is computationally expensive and should only be used when it's fed a format string. For cases where the error doesn't contain a format string (like here), then `errors.New` will very cheaply return an `error` struct from a fixed string.

stgraber (approved):

PR #2263, opened 2025-07-09
Make `tunnel.NAME.(interface|local)` a node specific network config

breml:


stgraber:
@breml I think we should make all the tunnel.XYZ keys be node specific as most tunnels use local IPs and the like anyway.

breml:
@stgraber I double checked the `tunnel.XYZ` keys and I don't think, it makes sense to make all of them node specific.

My proposal would be:

|key|description|node specific|
|-|-|-|
|tunnel.NAME.group|Multicast address for vxlan (used if local and remote aren’t set)|no|
|tunnel.NAME.id|Specific tunnel ID to use for the vxlan tunnel|no|
|tunnel.NAME.interface|Specific host interface to use for the tunnel|yes|
|tunnel.NAME.local|Local address for the tunnel (not necessary for multicast vxlan)|yes|
|tunnel.NAME.port|Specific port to use for the vxlan tunnel|no|
|tunnel.NAME.protocol|Tunneling protocol: vxlan or gre|no|
|tunnel.NAME.remote|Remote address for the tunnel (not necessary for multicast vxlan)|no|
|tunnel.NAME.ttl|Specific TTL to use for multicast routing topologies|no|

(taken from https://linuxcontainers.org/incus/docs/main/reference/network_bridge/#configuration-options)

stgraber:
You're right for the VXLAN mesh case that we care about now, but for the other types of tunnel, we'd definitely want to have some control as to what machine has a given tunnel set up. 

Otherwise if you combine a cluster-wide VXLAN tunnel with a regular GRE or VXLAN tunnel, you'll get all the servers suddenly sending the same traffic over that extra tunnel which may then send it back once of the other tunnels, creating a loop. 


Anyway, we may be able to make your approach work by being pretty careful about when we bring up the tunnel. 

If dealing with GRE, that's easy, we need a local IP, so if it's not set in the node config, we skip the tunnel. 

VXLAN is a bit trickier because of the multicast support, for that one, we basically need to check if we have a remote IP defined, then if we do, a local IP also becomes required for the tunnel to be brought online.

breml:
OK, I will implement the verification of the node specific config settings based on my table above, fix the linter issues and I will have a look at the conditions for when to bring up which tunnel and update the PR.

breml:
@stgraber I just looked at the code. The two checks you have mentioned in https://github.com/lxc/incus/pull/2263#issuecomment-3052750935 are already present. For [gre](https://github.com/lxc/incus/blob/main/internal/server/network/driver_bridge.go#L1694-L1696) and for [vxlan](https://github.com/lxc/incus/blob/main/internal/server/network/driver_bridge.go#L1717-L1721). So think we are good on that side.

stgraber (approved):

PR #2268, opened 2025-07-11
incusd/device/nic_macvlan: Fix IPv6 for VMs

stgraber:


stgraber:
@gwenya I'm pretty confused as to why I have to do it like this...

The existing logic for AllMulticast doesn't actually result in the link having the flag set.

gwenya:
That is weird, I'm looking into it.

If this is the only way to do it then I would move that logic directly into the ip package.

gwenya:
It seems the netlink code just doesn't set the allmulti flag when creating a link at all. I don't know if this is a limitation of the library or of netlink itself, but it occurs to me that `ip` doesn't support that either, you have to first add the link and then run `ip link set dev <name> allmulticast on`. 

I'll look into whether we have any other settings that aren't set on link creation.

gwenya:
#2269

PR #2277, opened 2025-07-15
incusd/auth/oidc/oidc: Add oidc.redirect_uri setting to allow custom redirect URIs

tibeer:
Helpful, if you try to run incus behind NAT but your IDP is online hosted.

### DIFF
--- a/cmd/incusd/api_1.0.go
+++ b/cmd/incusd/api_1.0.go
@@ -231,7 +231,7 @@ func api10Get(d *Daemon, r *http.Request) response.Response {
 	// Get the authentication methods.
 	authMethods := []string{api.AuthenticationMethodTLS}
 
-	oidcIssuer, oidcClientID, _, _, _ := s.GlobalConfig.OIDCServer()
+	oidcIssuer, oidcClientID, _, _, _, _ := s.GlobalConfig.OIDCServer()
 	if oidcIssuer != "" && oidcClientID != "" {
 		authMethods = append(authMethods, api.AuthenticationMethodOIDC)
 	}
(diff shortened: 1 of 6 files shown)


tibeer:
Hi, first time contributor here. I tried my best to catch all possible places where the OIDC settings are used, but I have no possibility to test it, as I am not a go developer. My intention was to allow me to use an upstream OIDC provider (like Zitadel) while Incus is running at my home network  with NAT from the router.

stgraber:
Logic seems fine though I'm unsure about the exact use case (which we'd probably want to put in the doc string).

If you have Incus hosted locally at say https://10.0.0.100:8443 and you're using an IDP at https://my-idp.example.net, we should see:
 - Client talks to https://10.0.0.100:8443
 - Get Unauthorized with redirect to IDP with redirect URI set to https://10.0.0.100:8443/oidc/callback
 - IDP goes through login flow and redirects client to https://10.0.0.100:8443/oidc/callback with the tokens URL encoded
 - Incus server extracts the tokens and sets them as cookies
 - Client connects again this time with the cookies set
 - Incus validates the token and allows the request

The URIs are generated based on what the client provided as the request URI so as long as the IDP allows the string as a valid redirect URI, you'll be good.


You're mentioning NAT, so I guess the issue is that you're only NATing some client and you're dealing with an IDP that doesn't allow for multiple allowed redirect URIs?

So you'd have Incus available as both https://10.0.0.100:8443 and https://192.168.0.100:8443 and would have the IDP configured for https://192.168.0.100:8443 but are connecting from the client through https://10.0.0.100:8443 causing an invalid redirect URI.

If so, your PR would fix the validation issue, but you'd need to be connecting from a client that can actually hit https://192.168.0.100:8443 for the redirect to work at all, this would then lead to cookies being set for https://192.168.0.100:8443 rather than https://10.0.0.100:8443 so subsequent requests would keep bouncing you to the IDP as the cookies would never be presented to the server.

What am I missing?

tibeer:
Nope, I guess you are totally right. In the meantime, I solved my home lab issues with [dex](https://dexidp.io). It sits here in a docker container and takes care of authenticating me with GitHub. And you summoned it up perfectly btw. Closing this as no longer needed.

PR #2296, opened 2025-07-21
incusd/storage: Fix ISO renaming

bensmrs:
Closes: #2290

Not a huge fan of the solution, as it leads to code duplication with `MountPath`. Some refactoring may be worth it, LMK if it’s something you’d like me to look at.


bensmrs:
Might also be worth investigating why this code path fails silently.

Issue #2238, opened 2025-06-30
Make the amount of hotpluggable memory configurable


stgraber:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

Currently the `maxmem` property for VMs is determined by Incus but that's been known to cause issues on some hosts as well as with some guest operating systems.

To give more user control, it'd be good to have a `limits.memory.hotplug` option which can be set to `false` to not allow any memory to be hotplugged, or an upper limit to how much memory the VM can get to through hotplug (larger or equal to `limits.memory`).

PR #2301, opened 2025-07-23
Add 'limits.memory.hotplug' configuration option

presztak:
The `limits.memory.hotplug` option controls how memory hotplug is handled for the virtual machine. It can be set to `false` to completely disable memory hotplugging. Alternatively, it can be set to a value that defines the maximum amount of memory the VM can reach through hotplug. This value must be greater than or equal to `limits.memory`.

Closes: #2238


internal/server/instance/drivers/driver_qemu.go
```diff
@@ -6470,6 +6487,12 @@ func (d *qemu) updateMemoryLimit(newLimit string) error {
 	if curSizeMB == newSizeMB {
 		return nil
 	} else if baseSizeMB < newSizeMB {
+		limitsMemoryHotplug := d.expandedConfig["limits.memory.hotplug"]
+		memoryHotplugEnabled := !util.IsFalse(limitsMemoryHotplug)
+		if !memoryHotplugEnabled {
```
stgraber: I'd shorten that to just `if util.IsFalse(d.expandedConfig["limits.memory.hotplug"])` so we don't have the double negation and temporary variable.

stgraber:
There's a typo in the commit title, `inctance` instead of `instance`.

stgraber:
Rest of logic looks good, so just those two small fixes.

internal/server/instance/drivers/driver_qemu.go
presztak: Fixed.

presztak:
All comments have been addressed.

PR #2325, opened 2025-07-30
incusd/device/disk: prevent file mounts on VMs

### LINKED ISSUE #2322: Individual file mounting not working with VMs

### Incus system details

```yaml
Same as before :)
```

### Instance details

_No response_

### Instance log

_No response_

### Current behavior

While having a look at #2230, I noticed that individual file mounting doesn’t work with VMs, but in a non-obvious way. It creates an empty directory instead of either refusing to create the device or creating (i.e.) a bind-mounted file.

### Expected behavior

_No response_

### Steps to reproduce

```bash
# incus config device add trixie foo disk source=/etc/fstab path=/srv/fstab
Device foo added to trixie
# incus exec trixie file /srv/fstab
/srv/fstab: directory
# incus exec trixie ls /srv/fstab
<nothing>
```

bensmrs:
Closes #2322

### DIFF
--- a/internal/server/device/disk.go
+++ b/internal/server/device/disk.go
@@ -1310,12 +1310,17 @@ func (d *disk) startVM() (*deviceConfig.RunConfig, error) {
 				// takes effect event if using virtio-fs (which doesn't support read only mode) by
 				// having the underlying mount setup as readonly.
 				var revertFunc func()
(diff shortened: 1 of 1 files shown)


bensmrs:
Hmm it’s actually a bit trickier than it looks, I’m closing this and continuing the discussion in the issue.

Issue #2257, opened 2025-07-05
Image to profile association issue


victoitor:
### Is there an existing issue for this?

- [x] There is no existing issue for this bug

### Is this happening on an up to date version of Incus?

- [x] This is happening on a supported version of Incus

### Incus system details

```yaml
config:
  cluster.https_address: 10.11.16.11:8443
  core.https_address: '[::]:8443'
  instances.placement.scriptlet: "# Error return strings\nerror_cpu_not_string        =
    \"Configuração de limits de CPU não é string.\"\nerror_cpu_syntax            =
    \"Sintaxe inválida de limite de CPU.\"\nerror_cpu_not_pinned        = \"Núcleos
    de CPU não foram fixados.\"\nerror_cpu_key               = \"Chave limits.cpu
    com erro: \"\nerror_pinned_cpu_not_allowed = \"CPUs fixados pela instância não
    são permitidos pelo projeto.\"\n\nerror_mem_not_string        = \"Configuração
    de limite de memória não é string.\"\nerror_mem_syntax            = \"Sintaxe
    inválida de limite de memória. \"\nerror_mem_key               = \"Chave limits.memory
    com erro: \"\nerror_mem_not_fixed         = \"Memória não fixada.\"\nerror_mem_limit_exceeded
    \   = \"Memória acima do permitido pelo projeto.\"\n\nerror_responsible_undefined
    = \"Responsável não definido para instância na chave user.responsavel.\"\nerror_project_config
    \       = \"Erro na configuração do projeto. Por favor fale com um administrador
    do cluster para resolver o problema. Erro: \"\nerror_no_available_node     = \"Não
    há máquina válida disponível.\"\n\n# Parses limits.cpu type strings which should
    have pinned cpus. Returns either an error string or a tuple containing the cpu
    bit vector and the most significant set bit\ndef parse_cpu_pin(cpu_string):\n
    \   cpu_bit_vector  = 0\n    mssb            = -1\n\n    if type(cpu_string) !=
    \"string\":\n        return error_cpu_not_string\n    if cpu_string.isdigit():\n
    \       return error_cpu_not_pinned\n    parts = cpu_string.split(\",\")\n    for
    p in parts:\n        r = p.split(\"-\")\n        if len(r) not in (1, 2):\n            return
    error_cpu_syntax\n        if len(r) == 1:\n            if not r[0].isdigit():\n
    \               return error_cpu_syntax\n            cpu = int(r[0])\n            cpu_bit_vector
[...]

PR #2327, opened 2025-07-30
Fix association between images and profiles

presztak:
1. Populate `images_profiles` with an entry linking the image to the default profile from the default project, if the image belongs to a project where `features.images` is disabled.
2. Fix issue with setting the cached option for images belonging to projects where `features.images` is disabled.

Fixes: #2257


internal/server/db/images.go
```diff
@@ -767,9 +767,18 @@ func (c *ClusterTx) UpdateImageLastUseDate(ctx context.Context, projectName stri
 
 // SetImageCachedAndLastUseDate sets the cached and last_use_date field of the image with the given fingerprint.
 func (c *ClusterTx) SetImageCachedAndLastUseDate(ctx context.Context, projectName string, fingerprint string, lastUsed time.Time) error {
+	enabled, err := cluster.ProjectHasImages(ctx, c.tx, projectName)
+	if err != nil {
+		return fmt.Errorf("Check if project has images: %w", err)
+	}
+
+	if !enabled {
+		projectName = "default"
```
stgraber: api.ProjectDefaultName

internal/server/db/images.go
```diff
@@ -952,7 +952,11 @@ func (c *ClusterTx) CreateImage(ctx context.Context, project string, fp string,
 						JOIN projects_config AS t2 ON t2.project_id = profiles.project_id
 							AND t2.key = "features.profiles"
 							AND t2.value = "true"
-						WHERE profiles.name = "default"`, id)
+						WHERE profiles.name = "default"
+			UNION
+					SELECT ?, profiles.id FROM profiles
+						JOIN projects ON projects.id = profiles.project_id
+						WHERE projects.name = ? AND profiles.name = "default"`, id, id, imageProject)
```
stgraber: Logic seems good but looking at the existing code, we're not supporting our full range of true/false values. So we should probably pull the full data from `projects_config` so we can properly use IsTrue/IsFalse on the values, then once we know the relevant IDs we can perform the actual INSERT.

internal/server/db/images.go
presztak: Done.

internal/server/db/images.go
presztak: Fixed.

stgraber (approved):

PR #2330, opened 2025-07-31
incusd/instance/qemu: Cleanup volume eject/detach logic

stgraber:
This removes some duplicate logic, avoids failures when volumes come from profiles and adds special-handling for our internal volumes.

### DIFF
--- a/internal/server/instance/drivers/driver_qemu.go
+++ b/internal/server/instance/drivers/driver_qemu.go
@@ -5893,33 +5893,8 @@ func qemuDetachDisk(s *state.State, id int) func(string) error {
 // Detach a disk from the instance.
 func (d *qemu) detachDisk(name string) error {
 	diskName := strings.TrimPrefix(name, qemuDeviceIDPrefix)
-	var id int
-	var devices map[string]dbCluster.Device
-	err := d.state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
-		object, err := dbCluster.GetInstance(ctx, tx.Tx(), d.project.Name, d.name)
-		if err != nil {
-			return err
-		}
 
-		id = object.ID
-		devices, err = dbCluster.GetInstanceDevices(ctx, tx.Tx(), id)
-		if err != nil {
-			return err
-		}
-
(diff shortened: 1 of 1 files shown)


stgraber:
@bensmrs any issue with this?

It simplifies the logic a bit and fixes an issue where RHEL based systems would only boot successfully once as after that `source=agent:config` gets ejected and is no longer available after reboot.

internal/server/instance/drivers/driver_qemu.go
```diff
@@ -5935,8 +5910,27 @@ func (d *qemu) detachDisk(name string) error {
 		return err
 	}
 
+	// Check if it's a special device (we don't store detached state on those).
+	if slices.Contains([]string{"agent:config", "cloud-init:config"}, config["source"]) {
```
bensmrs, line 5914: To me, this would only trigger if the agent is added explicitly in Incus config. It avoids a round of DB transactions leading to an ignored error, though.

internal/server/instance/drivers/driver_qemu.go
```diff
@@ -5935,8 +5910,27 @@ func (d *qemu) detachDisk(name string) error {
...
+		return nil
+	}
+
+	// Find the disk device.
+	_, ok = d.localDevices[diskName]
```
bensmrs, line 5919: I’m not sure how I didn’t think about using `localDevices` in the PR.

bensmrs (commented):
If you tested it and it works well with non-agent/cloud-init drives, then LGTM

internal/server/instance/drivers/driver_qemu.go
stgraber: `incus config device add MY-INSTANCE agent disk source=agent:config` is what you have to do for all RHEL based images. All of those suddenly failed to start the agent with the addition of the `attached` property.

bensmrs: Oh ok, I forgot how RHEL was painful to handle :)

stgraber:
Yeah, I did a quick test both with the special drives and with a regular one, behaved as expected.

PR #2331, opened 2025-07-31
tests: Standardize indentation

stgraber:
This is the result of: shfmt -w -i 4 -sr -l test/

### DIFF
--- a/test/backends/btrfs.sh
+++ b/test/backends/btrfs.sh
@@ -1,29 +1,29 @@
 btrfs_setup() {
-  # shellcheck disable=2039,3043
-  local INCUS_DIR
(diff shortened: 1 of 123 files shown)


bensmrs:
Gonna be really painful to git blame…

stgraber:
@bensmrs yeah, those kind of changes tend to make it a bit harder but it's not that bad.
You typically `git blame` and find that commit, then you `git blame HASH^` to blame past that commit. I pretty frequently have to do that when using git blame anyways.

bensmrs:
I just was looking into `git blame --ignore-rev=XXX`, I’ll have to try

tych0:
there is also a .git-blame-ignore-revs!

PR #2335, opened 2025-07-31
Fix directory unmount

bensmrs:
Closes: #2321


bensmrs:
Not sure if I should keep the same attempt loop as for the mounting code

bensmrs:
In the failing test, we’re checking non-FS devices, and as the path removal logic has been removed, the `/dev/XXX` mount isn’t removed. It’s not beautiful, but should we perform path deletion only if the path starts with `/dev/`?

bensmrs:
We could also pass a magic value to `FSType` in `unixDeviceRemove` or in `(d *disk) Stop` (the latter could be better, as we may be adding other funny device types that behave as UNIX devices) and filter on this in the path removal logic, your call.

stgraber:
> In the failing test, we’re checking non-FS devices, and as the path removal logic has been removed, the `/dev/XXX` mount isn’t removed. It’s not beautiful, but should we perform path deletion only if the path starts with `/dev/`?

That sounds like a nice and quick fix. Given /dev is a mount we manage (as a tmpfs), we know it's fine to delete stuff in there, so special casing it seems fine to me.

bensmrs:
Alright. I removed the warning logic, as the comments suggest it was there only for FS mounts.

bensmrs:
Looks like we don’t need the first `/`

stgraber (approved):

Issue #2323, opened 2025-07-29
Directory mounts declared before boot cannot be hot-detached on VMs


bensmrs:
### Is there an existing issue for this?

- [x] There is no existing issue for this bug

### Is this happening on an up to date version of Incus?

- [x] This is happening on a supported version of Incus

### Incus system details

```yaml
same same
```

### Current behavior

Everything should be clear from the following steps.

### Steps to reproduce

```bash
root@test-incus:~/incus# incus launch images:debian/trixie trixie4 --vm
Launching trixie4
root@test-incus:~/incus# incus config device add trixie4 foo disk source=/tmp path=/srv
Device foo added to trixie4
root@test-incus:~/incus# incus exec trixie4 ls /srv
<host tmp files>
root@test-incus:~/incus# incus config device remove trixie4 foo
Device foo removed from trixie4
root@test-incus:~/incus# incus stop trixie4
root@test-incus:~/incus# incus config device add trixie4 foo disk source=/tmp path=/srv
Device foo added to trixie4
root@test-incus:~/incus# incus start trixie4
root@test-incus:~/incus# incus exec trixie4 ls /srv
<host tmp files>
root@test-incus:~/incus# incus config device remove trixie4 foo
Error: Failed to stop device "foo": Failed to detach path device after 10s
```

bensmrs:
FWIW, manually executing `chardev-remove` on a secondary QMP monitor (after the previous failure) returns that the chardev is busy.
```json
{"QMP": {"version": {"qemu": {"micro": 2, "minor": 2, "major": 9}, "package": "v9.2.2-dirty"}, "capabilities": ["oob"]}}
{"execute": "qmp_capabilities"}
{"return": {}}
{"execute": "chardev-remove", "arguments": {"id": "incus_foo"}}
{"error": {"class": "GenericError", "desc": "Chardev 'incus_foo' is busy"}}
```

Unmounting from the VM beforehand works as expected from the guest PoV, but the chardev stays there in a busy state.

[...]

PR #2360, opened 2025-08-08
Fixes for virtiofs and 9p shares

bensmrs:
Bugs solved:
* Typing regressions caused by #1531
* Non-hot-unpluggable pre-boot virtiofs shares (Closes: #2323)
* 9p post-boot shares being wrongfully mounted as virtiofs
* Useless checks probably caused by confusion because `FSType=9p` server-side even in virtiofs shares (note that it’s different agent-side)

Additional actions:
* Remove pre-boot virtiofs logic
* Remove dual virtiofs+9p logic, enforcing a **single** mount type
* Make the agent a little dumber on path mounts (removing a virtiofs dmesg error when mounting 9p-only shares)

Future directions (to discuss here):
* Maybe add more intelligence to disable virtiofs if `image.os` and related keys match a house-tailored heuristic?
* Clarify `FSType=9p` into `FSType=share` or something like that?
* (more globally) Actually restrict the user / non-internal server operations from setting some volatile keys? (some volatile keys should IMO still be settable by the user, like `hwaddr`)

That was a tough one to do, I tried to split the commits into easily reviewable ones, although some still are pretty big.

### DIFF
--- a/internal/server/device/disk.go
+++ b/internal/server/device/disk.go
@@ -130,8 +130,14 @@ func (d *disk) sourceIsCeph() bool {
 
 // CanHotPlug returns whether the device can be managed whilst the instance is running.
 func (d *disk) CanHotPlug() bool {
-	// All disks can be hot-plugged.
-	return true
+	// 9p mounts cannot be hotplugged. However, with io.bus=auto, we can't know at startup time
+	// if we are dealing with a 9p mount. Still, it's better to fail early. At stop time, we can
+	// extract the info from the volatile key. All other disks can be hotplugged.
+	if d.bus == "" || d.bus == "auto" {
+		return d.volatileGet()["io.bus"] != "9p"
+	}
+
+	return d.bus != "9p"
 }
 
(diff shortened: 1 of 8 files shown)


internal/server/device/disk.go
```diff
@@ -359,8 +368,12 @@ func (d *disk) validateConfig(instConf instance.ConfigReader) error {
 		//
 		// For file systems (shared directories or custom volumes), this is one of:
 		// - `9p`
-		// - `auto` (default) (`virtiofs` + `9p`, just `9p` if `virtiofsd` is missing)
+		// - `auto` (default) (`virtiofs` if possible, else `9p`)
 		// - `virtiofs`
+		//
+		// `9p` doesn't support hotplugging and `virtiofs` doesn't support live migration. `auto` tries
+		// to use `virtiofs` if possible (`migration.stateful` not set to `true` and host support for
+		// virtiofsd) and falls back to `9p` otherwise.
```
stgraber: virtiofsd needs quoting (`virtiofsd`) in that last sentence.

Issue #1833, opened 2025-03-23
ACL should apply allow first when default action is allow


stgraber:
Currently when applying ACLs, Incus will first apply all drop/reject, then apply the allow rules and finally the policy which defaults to reject. 

That's a good order for this particular case as it allows having wide allow rules with narrow reject/drop rules to block a subset of what's allowed, then the policy rejects the rest.

But this apply order isn't a good fit when the default policy is allow as in that situation one would likely have pretty broad reject/drop rules and so want the allow rules applied first to allow a subset of what would otherwise be denied by broader reject/drop rules. 

So we should change our default apply order to match, basically reversing the order if the default action is allow.

AbhinavTiruvee:
@stgraber Hi, I'm part of a group at UT Austin taking a class on virtualization. Could my partner and I take a look at this issue and try to solve it?

stgraber:
Done.

You're going to want to set up Incus with OVN locally. There is some documentation for that in https://linuxcontainers.org/incus/docs/main/howto/network_ovn_setup/.

You should then be able to reproduce the current (wrong) behavior by:
 - Creating an ACL that:
   - Allows connecting to 1.1.1.1
   - Blocks connecting to 1.1.1.0/24
 - Apply the ACL to an instance
 - Set the default action for the ACL to allow

Normally you'd expect this setup to allow talking to 1.1.1.1, but because we currently apply the drop/reject ahead of the allow rules, you'll find that 1.1.1.1 won't be allowed due to the 1.1.1.0/24 reject rule.

Once the issue is fixed, the same setup should allow pinging 1.1.1.1 but reject pinging 1.1.1.2.

stgraber:
@AbhinavTiruvee are you still working on this?

AbhinavTiruvee:
Sorry for not letting you know sooner; No, I am not. Please unassign me from this issue.

PR #2372, opened 2025-08-11
Add directional port groups with allow/reject priority based on default policy

presztak:
This change adds support for prioritizing allow rules when the default policy for a network or interface is set to allow. By default, the policy is reject, so reject rules have higher priority.
To implement this, the original single port group was replaced with four separate port groups:
- Ingress (allow-first)
- Ingress (reject-first)
- Egress (allow-first)
- Egress (reject-first)

These groups store rules in the required order, allowing correct rule evaluation based on the default policy.

Fixes: #1833

### DIFF
--- a/cmd/incusd/patches.go
+++ b/cmd/incusd/patches.go
@@ -17,9 +17,12 @@ import (
 	"github.com/lxc/incus/v6/internal/server/db"
 	dbCluster "github.com/lxc/incus/v6/internal/server/db/cluster"
 	"github.com/lxc/incus/v6/internal/server/db/query"
+	deviceConfig "github.com/lxc/incus/v6/internal/server/device/config"
 	"github.com/lxc/incus/v6/internal/server/instance"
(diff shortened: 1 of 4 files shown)


cmd/incusd/patches.go
```diff
@@ -1343,4 +1347,250 @@ UPDATE networks_load_balancers SET ports="null" WHERE ports="";
...
+
+	if !isLeader {
+		return nil
+	}
+
+	// Check that OVN is available.
+	ovnnb, _, err := s.OVN()
+	if err != nil {
+		return err
+	}
```
stgraber: What happens when this runs on a system without configured OVS/OVN?

presztak: yeah, it will fail with an error saying that it’s unable to connect to OVN. So as solution I'm instantiating ovnnb only if we find ovn network.

stgraber (approved):

PR #2383, opened 2025-08-13
incus/dhcp: Explicitly requesting 51/58/59, derive sane T1, add jitter

xzkutor:
The built-in DHCPv4 client used for OCI containers can overwhelm the DHCP server with renewals every 60 seconds. This occurred when the initial DHCPOFFER lacked renewal/rebind timers (Options 58/59). The renewal loop slept using IPAddressRenewalTime(time.Minute), so the fallback value (1 minute) was used repeatedly.

Additionally, the initial request implicitly asked for a fixed lease time (Option 51) and did not include a Parameter Request List for 51/58/59. Many servers respond by honoring the client’s requested lease (e.g., 3600s) and omitting 58/59 in the OFFER/ACK. Together, these two behaviors produced one-minute renewals even when the server was configured for multi-hour leases.

This change fixes the above by:

1. Not requesting a fixed lease in the client message (WithoutOption(OptionIPAddressLeaseTime)), allowing the server’s default-lease-time to apply.

2. Explicitly requesting critical options via PRL on both REQUEST and RENEW: 1,3,6,15,51,58,59 (and 121 if present). This makes servers include Lease-Time/T1/T2.

3. Deriving a sane T1 for the renewal sleep: prefer timers from the ACK, then OFFER; if still absent, use 50% of the offered lease-time (Option 51); finally fall back to 1m.

4. Adding ±5% jitter to the renewal sleep to avoid herd effects when many containers renew at once.

### DIFF
--- a/cmd/incusd/main_forknet.go
+++ b/cmd/incusd/main_forknet.go
@@ -216,6 +216,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math/rand"
 	"net"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


stgraber:
Thanks!

PR #2392, opened 2025-08-15
doc: Add "Disable the tables" way to "Prevent connectivity issues wit…

anlexN:
…h Incus and Docker", add `nft` command to "Allow egress network traffic flows"

Closes [Why can I not ping and apt update?](https://discuss.linuxcontainers.org/t/why-can-i-not-ping-and-apt-update/24284)


doc/howto/network_bridge_firewalld.md
```diff
@@ -157,6 +321,11 @@ Allow egress network traffic flows
       iptables -I DOCKER-USER -i incusbr0 -j ACCEPT
       iptables -I DOCKER-USER -o incusbr0 -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
 
+  Or the equivalent `nft` commands:
+
+      nft insert rule ip filter DOCKER-USER iifname k8s-net counter accept
+      nft insert rule ip filter DOCKER-USER oifname k8s-net ct state related,established counter accept
```
stgraber: The example uses `incusbr0` so this should be updated to match.

doc/howto/network_bridge_firewalld.md
```diff
@@ -115,6 +115,170 @@ See [Docker on a router](https://docs.docker.com/engine/network/packet-filtering
...
+  When you use Docker, you should enable Docker tables, disable incus table:
+
+  ```shell
+  nft add table ip nat
+  nft add table ip filter
+  nft add table ip6 nat
+  nft add table ip6 filter
+  nft add table inet incus { flags dormant\; }
+  ```
+
```
stgraber: We definitely don't want to put a full `nft` ruleset dump in our documentation, this will never be kept up to date and will make things more confusing for everyone.

I also would prefer not to have those particular `nft` commands in there.
They don't just suspend the Docker rules, they suspend everything which would be very dangerous on a production system.

stgraber: A better option that may be worth mentioning in here is `"ip-forward-no-drop": true` in Docker configuration which should also make it stop interfering on the system.

doc/howto/network_bridge_firewalld.md
anlexN: @stgraber Thanks for your reply. I have updated this pr, please reviewed again.

doc/howto/network_bridge_firewalld.md
anlexN: 1. I have reviewed nftables project from the scratch, nftables team have gived up iptables, recommend to use `nft`. `iptables-translate` can help user translate iptables command to nft command. In the big scale, `nft` is faster than `iptables`.
2. `flags dormant` is table level, can not be chain level. However, in the big scale, docker will modify table level's rulesets. `"ip-forward-no-drop": true` is only focused on `forward` chain, but not others like `nat`. `Incus` also set `nat` rulesets at its ways by itself. Further, "Prevent connectivity issues with Incus and Docker" have many solutions, so what way user select should be decided by user. after all， those way is user command, not incus command. so **You don't need to worry**.
3. many and many fresh people are misdirected to learn `iptables`, it is very unfair to them.

PR #2393, opened 2025-08-15
Fix typo and add clarity for project param. 

AJRepo:
Typo: "the `project` param is are provided" 

I was reading this page and this help question:   https://discuss.linuxcontainers.org/t/problem-with-incus-prometheus-and-multiple-projects/24226  . I was similarly confused by the documentation as an incus-metrics newbie. So in fixing the typo I also changed the note to explicitly state that if you don't use the project param then you get the default project or all projects. 

Todo: Add how to define the default project?

### DIFF
--- a/doc/metrics.md
+++ b/doc/metrics.md
@@ -183,8 +183,8 @@ scrape_configs:
   # to the `/var/lib/incus/cluster.crt` file found on every member of
   # the Incus cluster.
   #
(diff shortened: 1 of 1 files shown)


AJRepo:
oops - I forgot to sign off.  Would you prefer a `git push -f` or deleting and resubmitting?

AJRepo:
Closing and will re-submit with a sign-off.

PR #2402, opened 2025-08-18
Handle USB cdrom drives

stgraber:
USB mass storage with QEMU is pretty weird. There is no way to pass through the `media` argument to have it give us a CD-ROM drive on its own, instead the `usb-storage` driver only does disk devices and generates an internal SCSI device.

To fix this, we need to switch to either `usb-bot` or `usb-uas`. The former still uses the `usb-storage` driver and doesn't need anything special at the guest level, so that's how I went here.

With this in place, we can now have anything marked as an ISO show up as a cdrom device rather than mass storage. That's particularly important for the Windows installer which with this will now function without any need for repacking of the install ISO.

### DIFF
--- a/internal/server/instance/drivers/driver_qemu.go
+++ b/internal/server/instance/drivers/driver_qemu.go
@@ -2629,6 +2629,14 @@ func (d *qemu) deviceDetachBlockDevice(deviceName string, rawConfig deviceConfig
 		return err
 	}
 
+	if rawConfig["io.bus"] == "usb" {
+		// When dealing with USB, remove the intermediate USB device too.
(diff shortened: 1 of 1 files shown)


stgraber:
@bensmrs I don't know if that helps with any of your stuff, but it might. Took be a solid day to figure this stuff out...

bensmrs:
Wow, great catch!

> USB mass storage with QEMU is pretty weird

At this point, what surprises me is when QEMU isn’t weird.

> I don't know if that helps with any of your stuff, but it might

Nah, my stuff is mostly sorted out, I’m just lacking some time to work on it. But hey, we never know :)

hallyn (approved):

PR #2412, opened 2025-08-20
github: Downgrade Linstor

### LINKED ISSUE #2409: Linstor tests failing

### Incus system details

```yaml
Not relevant to this issue
```

### Instance details

_No response_

### Instance log

_No response_

### Current behavior

_No response_

### Expected behavior

_No response_

### Steps to reproduce

Run the current testsuite on Github.

https://github.com/lxc/incus/actions/runs/17106160968/job/48515482675?pr=2407

Error: Failed creating instance from image: Clone operation failed

stgraber:
Let's not test on pre-release software.

Closes #2409

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -408,9 +408,9 @@ jobs:
           # Install everything required to compile DRBD and run LINSTOR tools.
           sudo apt-get install --no-install-recommends -y \
             drbd-dkms \
-            linstor-client \
-            linstor-controller \
-            linstor-satellite \
+            linstor-client=1.25.4-1ppa1~noble1 \
+            linstor-controller=1.31.3-1ppa1~noble1 \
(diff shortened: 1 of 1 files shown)


luissimas:
AFAIK Linbit only makes the latest version available in the PPA (this includes rc versions). We can't downgrade the package using the PPA because there's no version available other than the latest :'(.

@bensmrs did you find any alternative repository that is more stable than the PPA?

bensmrs:
At work we’re using Debian, with LINBIT’s Proxmox repositories (aligned to Debian versions), which are stable. I still have in a corner of my head a project to actually package LINSTOR for Debian. So if we’re planning on moving the CI to Debian (which is obviously a good idea, *wink wink*), we can use the Proxmox repos without any problem.

stgraber:
Annoyingly Launchpad has the older builds still recorded in the PPA but they don't get included in the index which is why that PR didn't work...

PR #2413, opened 2025-08-21
incus/dhcp: default routes and multiple DHCP clients for OCI containers

xzkutor:
What changed:

1. internal/server/ip
  - Made Route.List device-optional: when DevName is empty or unresolved, the OIF filter is cleared and routes across all devices are returned.
  - Introduced a single builder netlinkRoute(mode) to construct netlink.Route for both apply (add/replace/delete) and query (list) paths.
 2. Implemented a DHCPv4 default route policy for OCI containers (incusd/main_forknet.go):
 - Keep a single default route in the main table, owned by the lexicographically smallest interface name.
 - Use Route.Replace to avoid EEXIST and make route updates idempotent.
 - Mark installed routes with Proto="dhcp".
 
Behavioral impact:
1. Route.List now returns routes across all devices when DevName is empty (previously, it always filtered by device).
2. Add, Replace, Delete, and Flush continue to require a valid device and behave as before.

### DIFF
--- a/cmd/incusd/networks.go
+++ b/cmd/incusd/networks.go
@@ -520,6 +520,11 @@ func networksPost(d *Daemon, r *http.Request) response.Response {
 			}
 		}
 
+		// Make sure that no description is set through the member-specific path.
+		if req.Description != "" {
+			return response.BadRequest(errors.New("The network description cannot be set for a specific member"))
+		}
+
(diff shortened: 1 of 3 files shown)


xzkutor:
will add it as a new pull-request with a single commit.

Issue #1822, opened 2025-03-21
Add tmpfs support for disk devices


huoxingdawang:
I noticed that Incus has recently supported OCI images, like Docker, and I really like this feature. 
However, I noticed that many Docker configurations include something like `--mount type=tmpfs,target=/tmp/cache,tmpfs-size=1000000000` to create a temporary directory within the container.

Currently, a feasible alternative may be to map all containers to the host's tmpfs using the "path" mode. 
But if all containers share the same directory on the host, some containers may not work properly. For example, multiple copies of the same application would try to access the same directory. 
If each container has its own tmpfs on the host, I need to manually create all the directories, which is inconvenient to maintain.
And for both methods, after restart the containe, the files will be kept rather than deleted.

For system containers with complete distributions, I can create tmpfs within the container, such as editing `/etc/fstab`. But this approach doesn't seem effective for applications, as they do not handle the content in `/etc/fstab`.

I'm not sure if there's already a solution for this, but if not, I suggest adding tmpfs support for disk devices. For example, the configuration could look like this, and we can get a 8GB temporary directory within the container:

```yaml
devices:
  tmpfs:
    path: /tmp
    source: tmpfs
    size: 8GB
    type: disk
```

stgraber:
Yeah, it's been a bit of a recurring problem with OCI containers.

The problem with any work on this is around memory accounting.

If you pass in a path on a tmpfs that's been mounted by a user on the host system.
The memory usage caused by the container writting to that tmpfs will not be counted as part of the memory limit of the container. This effectively allows the container to escape its limit and potentially cause a DoS attack on the host.

What we need is for the container itself to mount a tmpfs in the correct location.
Instruction liblxc to do it through the monitor process may be sufficient, but we'll need to do some testing there...


Actually, the issue is slightly worse than this as the actual behavior that Docker implements is:
 - Setup a tmpfs at some location
 - Setup overlayfs between the existing path in the image (underlay), tmpfs (workdir) and the path itself (overlay)

This then allows access and modification of the existing data with all changes being lost on container restart.


That's going to be quite tricky to implement in a way that's safe from DoS attacks.

The pure tmpfs is quite a bit easier, we can tell liblxc to do `tmpfs tmpfs /some/path create=dir,size=8G` and it should do what we want.

The overlay case is going to be quite a bit weirder. I don't recall if we have good support for mount-move in liblxc. If we do, we could abuse it with something like:
[...]

PR #2430, opened 2025-08-27
 Add tmpfs support for disk devices

presztak:
Closes: #1822

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2873,3 +2873,10 @@ The webhook data matches what's sent over the existing events API.
 ## `storage_driver_truenas`
 
 This adds a TrueNAS storage driver.
+
+## `container_disk_tmpfs`
+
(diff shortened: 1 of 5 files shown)


doc/api-extensions.md
```diff
@@ -2873,3 +2873,10 @@ The webhook data matches what's sent over the existing events API.
 ## `storage_driver_truenas`
 
 This adds a TrueNAS storage driver.
+
+## `container_disk_tmpfs`
+
+This adds tmpfs support for disk devices.
+
+* `source=tmpfs:` mounts a tmpfs filesystem, respecting size, uid, gid and mode options
```
stgraber: Should be "file system" and, then size, uid,  gid and mode should all be between "``" to avoid validation errors with the doc.

stgraber:
@presztak looks good. Can you add a couple of basic tests to make sure the feature works?

PR #2442, opened 2025-08-29
incusd/network: Exclude physical networks with parent=none from OVN chassis enablement

ViniRodrig:
- Updated OVN uplink and chassis detection logic to skip networks of type "physical" with parent set to "none".
- Ensured physical network driver setup permits parent="none" and checks interface existence accordingly.
- Nodes with physical uplinks that specify a valid parent (not "none") will participate in the OVN HA chassis group; nodes with parent="none" will be excluded.
- Behaviour tested by creating network with both parent=enp6s0 and parent=none on multiple nodes using Incus deploy.
- Only nodes with a valid physical uplink parent are added to the HA group, matching expected OVN behaviour.

### DIFF
--- a/internal/server/network/driver_ovn.go
+++ b/internal/server/network/driver_ovn.go
@@ -1554,6 +1554,11 @@ func (n *ovn) startUplinkPort() error {
 		return fmt.Errorf("Failed loading uplink network %q: %w", n.config["network"], err)
 	}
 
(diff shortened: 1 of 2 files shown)


stgraber:
Sorry for the delay, had a few busy weeks here, reviewing this one now!

PR #2450, opened 2025-09-02
Add certificate time validation to incus-agent

davidvasandani:
## Summary

This PR adds certificate time validation to the incus-agent to prevent confusing "not authorized" errors that occur when a VM's system clock is out of sync with the certificate validity period.

## Problem

When a Windows VM's system clock is behind (common after VM snapshots/restores), the agent fails with cryptic "not authorized" errors during TLS authentication. The root cause is that certificates appear "not yet valid" from the VM's perspective, but this fails silently within the TLS stack without clear diagnostics.

## Solution

Added proactive time validation checks at agent startup that:
- Validate system time is within certificate validity windows
- Provide clear, actionable error messages
- Show exact time differences to help administrators

### 1. Client Certificate Validation (`api_1.0.go`)
- Checks system time against `server.crt` validity period
- Reports time difference in minutes if outside valid window
- Provides clear "CRITICAL" error messages

### 2. Agent Certificate Validation (`network.go`)  
- Validates agent's own certificate before starting TLS server
- Ensures both certificates are valid before accepting connections

### 3. Enhanced Authentication Logging (`server.go`)
- Added detailed certificate fingerprint logging
- Trust store debugging information
- Improved troubleshooting capabilities

### When Clock is Behind:
```
System time is 166 minutes before certificate validity period
CRITICAL: System clock appears to be behind. Certificate is not yet valid.
Please sync system time or wait for certificate to become valid.
Error: System time (2025-09-02T05:00:00Z) is before certificate validity (2025-09-02T07:46:02Z)
```

### When Validation Passes:
```
Certificate time validation passed
Agent certificate time validation passed
```

## Testing

Tested on Windows 11 VM by:
1. Setting system clock 3 hours behind
2. Verifying clear error messages appear
3. Fixing time and confirming agent starts successfully
4. Validating Incus can connect after time sync

## Benefits

- **Early Detection**: Issues caught at startup vs during connection attempts
- **Clear Diagnostics**: Administrators immediately know it's a time issue
- **Reduced Support**: No more debugging mysterious "not authorized" errors
- **Better UX**: Actionable error messages that explain how to fix the problem

## Documentation

Added `TIME_VALIDATION.md` with:
- Implementation details
- Testing instructions  
- Windows time sync troubleshooting
- Common causes and fixes

This change is backward compatible and adds no new dependencies.

### DIFF
--- a/TIME_VALIDATION.md
+++ b/TIME_VALIDATION.md
@@ -1,0 +1,107 @@
+# Incus Agent Certificate Time Validation
+
+## Overview
+This version of the Incus agent includes certificate time validation checks to prevent authentication failures caused by system clock synchronization issues.
+
+## Problem Solved
+Previously, when a Windows VM's system clock was out of sync (particularly when behind), the agent would fail with cryptic "not authorized" errors during TLS authentication. This was because the certificate validity check would fail silently within the TLS stack.
+
+## Implementation
+
+### Changes Made
+
+1. **api_1.0.go** (lines 222-245)
+   - Added validation of system time against client certificate validity window
+   - Checks both NotBefore and NotAfter times
+   - Provides clear error messages with time differences in minutes
+
+2. **network.go** (lines 76-94)
+   - Added validation for the agent's own certificate
+   - Ensures agent certificate is valid before starting TLS server
+
+3. **server.go** (lines 108-129)
+   - Enhanced authentication logging with certificate fingerprints
+   - Added trust store debugging information
+   - Improved certificate comparison logging
+
+## Error Messages
+
+### When System Clock is Behind
+```
+System time is 166 minutes before certificate validity period
+CRITICAL: System clock appears to be behind. Certificate is not yet valid.
+Please sync system time or wait for certificate to become valid.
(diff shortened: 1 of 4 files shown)


stgraber:
Missing DCO sign-off, but in general I also don't think we'd merge this anyways.

The PR as it stands has a bunch of hardcoded logic which isn't suitable for inclusion, the extra document at the top-level of the tree doesn't really make sense either and the agent is usually meant to be run in the background as a service (just hasn't been packaged that way on Windows yet), so anything it logs will typically never actually be seen.

To me it feels like a better option would be a one-line NOTE/WARNING in the VM install instructions for the agent to mention that the agent uses TLS certificates for authentication and that the guest therefore needs to have its clock be on time.

davidvasandani:
Thanks for the feedback! For what it's worth, I was running the agent as a service and couldn't connect. Ran it manually with `--debug --verbose`, and the only error was `not authenticated`. If you strongly prefer not to modify the PR (perhaps to log the cert error only in `--debug` and/or `--verbose` mode), I'll update it to a doc-only change as you suggested.

stgraber:
I sent https://github.com/lxc/incus/pull/2452 which adds a warning to the page covering agent installation in VMs.

PR #2451, opened 2025-09-02
incusd: Deprecate gorilla/mux

0xk1f0:
As already determined in #2259, it would make sense to start phasing out [`gorilla/mux`](https://github.com/gorilla/mux) in all of incus, this PR aims to do just that.

Worth noting that `http.ServeMux` does not implement a `NotFoundHandler`, but since we already overwrite `ServeHTTP`, I tried to do a trivial reimplementation of it, which just checks if a `Handler` for a given route exists and executes a supplied function if not.

Most of the other changes here consist of replacing `mux.Vars(r)[]` with `r.PathValue()`. This all builds fine, but I have not extensively tested this yet, which is on my to-do list for the next coming days.


stgraber:
Mark as Draft until the regressions can be sorted out :)

PR #2457, opened 2025-09-04
Introduce initial CLI for IncusOS

stgraber:


cmd/incus/utils.go
```diff
@@ -839,3 +839,21 @@ func sshSFTPServer(ctx context.Context, sftpConn func() (net.Conn, error), entit
 		}()
 	}
 }
+
+func makeJsonable(data any) any {
```
jarrodu, line 843: It looks like this was the solution you were looking for. It seems like there might be more to the story though.

cmd/incus/admin_os.go
```diff
@@ -0,0 +1,837 @@
...
+
+		ts := time.UnixMicro(timeInt)
+
+		fmt.Printf("[%s] %s: %s\n", ts.Format(dateLayout), line["SYSLOG_IDENTIFIER"], line["MESSAGE"])
+	}
+
+	return nil
+}
+
+// IncusOS system configuration command.
```
jarrodu: This does not look like a command to me.

cmd/incus/admin_os.go
```diff
@@ -0,0 +1,837 @@
...
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("%s", data)
+
+	return nil
+}
+
+// IncusOS system configuration command.
```
jarrodu: Same. It does not look like a command.

cmd/incus/utils.go
stgraber: Yeah, that's what I had to use to get the JSON marshaller to function.
It won't handle all possible cases, but it does work with everything in our API right now.

cmd/incus/admin_os.go
stgraber: Those comments were indeed wrong (copy/pasted) though they'll still mentioned "command" as the struct is for what we internally refer to as a command (technically a sub-sub-sub-command in this case :)).

PR #2458, opened 2025-09-05
Clarify comment for cluster rebalancing

presztak:
Follow-up of #2454 , clarifies the comment for the `Rebalance` function.


internal/server/cluster/membership.go
```diff
@@ -14,7 +14,6 @@ import (
 
 	"github.com/cowsql/go-cowsql/app"
 	"github.com/cowsql/go-cowsql/client"
-
```
stgraber: Hmm, that one line change looks wrong, the line break is there to separate external and local imports.

presztak: Ah, sorry, I guess this happened after running `gci` on my side.

PR #2474, opened 2025-09-09
incus export. Fail fast if [target] is a directory

AJRepo:
I ran the command 

`incus export INSTANCE /path/to/dir/  --instance-only --optimized-storage`

which then ran the backup 

```
Backing up instance: ###.##MB (###MB/s)
```

and then after much waiting for the backup to complete  ... incus failed with the error

```
Error: open /path/to/dir/: is a directory
```

I would have preferred that it fail immediately instead of doing the backup and then failing for lack of a filename. 

This makes `incus export` fail immediately if [target] is a directory. 

For future use, it might be good to be able to specify [target] as a directory and then autoname the backup files, check for disk space, etc.

### DIFF
--- a/cmd/incus/export.go
+++ b/cmd/incus/export.go
@@ -67,6 +67,18 @@ func (c *cmdExport) Run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	var targetName string
+	if len(args) > 1 {
+		targetName = args[1]
+	} else {
+		targetName = name + ".backup"
+	}
+
+	// Test if target is a directory and fail if so
+	if fi, err := os.Stat(targetName); err == nil && fi.IsDir() {
+		return fmt.Errorf("Target path %q is a directory; please provide a file name", targetName)
+	}
+
 	instanceOnly := c.flagInstanceOnly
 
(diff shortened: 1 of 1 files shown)


AJRepo:
I see there are some linting errors from custom checks `test/lint/no-oneline-assign-and-test.sh`  - I'll rewrite and resubmit

AJRepo:
The last revision was created by  `gci write cmd/incus/export.go` and yet it created a rewritten file that fails the same linting checks it ran . Undoing its change fixes the failure in `golangci-lint run --timeout 5m --new --new-from-rev origin/HEAD`

AJRepo:
It's now complaining about i18n updates. I see In cmd/incus/export.go:35 I changed formatting from spaces to tabs
on one line to match other lines.  msgid ""
 "incus export u1 backup0.tar.gz\n"
-"    Download a backup tarball of the u1 instance."
+"\tDownload a backup tarball of the u1 instance."
 msgstr ""

Will run `make i18n` and resubmit. Sorry for my learning curve here.

stgraber:
Sorry for the delay, I've had a pretty busy couple of weeks. I'm now going to look into this one.

PR #2475, opened 2025-09-10
Feat/enable override config and device for import backup

heldergomes:


heldergomes:
well, I'd like to add few key points here.

- I read on incus documentation the following sentence: `When updating strings in the CLI tool (cmd/), you may need a commit to update the templates`, so I assumed that I should run make i18n, is it right?

- The device/config parser, which transform string[] into map[string], are basically the same as copy feature, is there any place on the project where can I move on, in order to reuse the code?

- Wasn't clear for me if is there any unit test which I have to implement, if so, let me know!

heldergomes:
I was thinking, what if we add a new file = `/shared/util/cmd.go` and move the configMap/deviceMap parser, which is shared in incus/copy and incusd/instances_post, is it reasonable?

stgraber:
> I was thinking, what if we add a new file = `/shared/util/cmd.go` and move the configMap/deviceMap parser, which is shared in incus/copy and incusd/instances_post, is it reasonable?

I missed that comment. Yeah, we can do that. 
I'm happy with the PR as it is now so we can merge it, but removing the duplication in a follow-up would be good. 

Package wise that may make more sense in internal/instance.

PR #2496, opened 2025-09-22
Replace map[string]string with ConfigMap for more user friendly handling of bool and numbers (with defaults)

breml:
If we agree to continue with this approach, the same might be applied to the `map[string]map[string]string` used for devices.

Continuation of #2940


stgraber:
Looks good, we can merge this one right after 6.17 is out so we have more time to catch any issue with it.

PR #2502, opened 2025-09-24
incusd/instance/qemu: Re-enable vsock on Windows

stgraber:
The latest virtio Windows drivers now include a signed vsock driver so let's re-enable the device.


stgraber:
I've opened https://github.com/mdlayher/vsock/issues/62

We may be able to use the provided TCP bridge in the meantime by effectively having the agent listen on a loopback address and then have the bridge send the traffic to it. Though the agent also generally likes being able to connect to the host, which wouldn't be possible here until we get native support.

hallyn (approved):

PR #2514, opened 2025-09-27
Change snapshot deletion command to `storage volume snapshot delete`

robbystk:
The current command doesn't work for me on incus 6.16.  The error message is `Error: Invalid storage volume "<volume name>/<snapshot name>"`.  It seems like the command should be `storage volume snapshot delete` rather than `storage volume delete` for a snapshot.

Please correct me if I'm wrong.


stgraber (approved):

stgraber:
Thanks!

Issue #2484, opened 2025-09-14
Introduce a `storage.logs_volume` config option


stgraber:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

Similar to the existing `storage.backups_volume` and storage.images_volume`, this will allow moving what normally would be in `/var/log/incus` over to a dedicated Incus-managed volume.

This is motivated by a recent stress test where after running over 7000 VMs on a server, our usage in /var/log/incus ended up filling the entire root disk. Nothing too bad, each VM was using just 3-4MB of storage, but that then would quickly add up.

Unlike what we do with backups and images, I don't expect us to move all data away from /var/log/incus and move it into the volume as that would interfere with /var/log/incus/incusd.log which is created prior to us setting up volume mounts.

Instead when the config key is set, we should relocate any instance folder from under /var/log/incus/ onto the newly created volume, then for any newly created instance log folder, we should do it under the configured volume, effectively only leaving /var/log/incus/incusd.log at the system path.

c24-tj:
Hi, may I try my hand at this?

stgraber:
Sure! I don't think @presztak started work on this one yet and he's got other things to work on so I changed the assignee over to you now.

PR #2525, opened 2025-10-02
cmd/incusd: add logs volume config

c24-tj:
This PR implements #2484


stgraber:
Logic for setting and handling of the new config key looks good but I think this is missing logic to actually have the instances write their logs to that path.

Most systems run with the default of using /var/log/incus/ so having a path now ready at /var/lib/incus/logs/ with the pre-existing data moved there will not cause new instances from logging at the new location.

Also another small wrinkle, running instances will not like their existing log files being moved across mounts, that will effectively break logging for those. So we should add a check so the config key can only be changed if the system has no running instances.

Issue #2506, opened 2025-09-25
Add ability to export ISO volumes


bensmrs:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

I’m able to import ISOs on a server, but I’d like to also export them. My current use case (it’s fine, I’ll do it another way) is that I have a custom installation ISO on a remote, and I’d like to upload it to another remote (but first, I need to download it).

### What do you think would need to be added

The ability to export ISOs with `incus storage volume export`, instead of getting `Error: Backup create: Volume of content type "iso" cannot be backed up`

stgraber:
Yeah, I ran into that one yesterday too.
It would certainly make sense to allow retrieving an ISO image that way. 

The "backup" would basically just be a raw dump of the ISO (no tarball involved) so you get back the same thing that was imported.

stgraber:
Let me know if you want to take this one.

bensmrs:
Yeah, I can take this one

bensmrs:
Well that one is a bit harder than I expected. The way I’m currently exploring is to have `BackupCustomVolume` use a generic interface instead of a TAR writer, and pass it either a custom IO writer or a TAR writer in `volumeBackupCreate` (and bypassing `volumeBackupWriteIndex` for ISOs). The custom IO writer will have no-ops for every public method except `WriteFileFromReader`, which will just dump the file content.
A bit dirty, but honestly the only way I see without rewiring the whole backup codebase.

PR #2526, opened 2025-10-02
Implement ISO export

bensmrs:
Closes: #2506

### DIFF
--- a/internal/server/storage/pool_interface.go
+++ b/internal/server/storage/pool_interface.go
@@ -143,7 +143,7 @@ type Pool interface {
 	MigrateCustomVolume(projectName string, conn io.ReadWriteCloser, args *migration.VolumeSourceArgs, op *operations.Operation) error
 
 	// Custom volume backups.
-	BackupCustomVolume(projectName string, volName string, tarWriter *instancewriter.InstanceTarWriter, optimized bool, snapshots bool, op *operations.Operation) error
+	BackupCustomVolume(projectName string, volName string, writer instancewriter.InstanceWriter, optimized bool, snapshots bool, op *operations.Operation) error
 	CreateCustomVolumeFromBackup(srcBackup backup.Info, srcData io.ReadSeeker, op *operations.Operation) error
 
 	// Storage volume recovery.
--- a/cmd/incus/storage_volume.go
+++ b/cmd/incus/storage_volume.go
@@ -3221,14 +3221,14 @@ type cmdStorageVolumeExport struct {
 func (c *cmdStorageVolumeExport) Command() *cobra.Command {
 	cmd := &cobra.Command{}
 	cmd.Use = usage("export", i18n.G("[<remote>:]<pool> <volume> [<path>]"))
-	cmd.Short = i18n.G("Export custom storage volume")
+	cmd.Short = i18n.G("Export custom storage volumes")
 	cmd.Long = cli.FormatSection(i18n.G("Description"), i18n.G(
-		`Export custom storage volume`))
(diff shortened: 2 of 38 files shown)


bensmrs:
Not gonna lie, this one was quite painful

bensmrs:
Oh, LVM seems to be behaving differently, but the behavior didn’t appear on my machine.
Edit: turns out on my machine, the ISO was 4M-aligned…

bensmrs:
So two questions come up:
- Should I drop the `import > export` idempotence or should I rather query the DB for the ISO size? (edit: I found a way to keep idempotence)
- Why in the current code don’t we use the actual volume size in the export rather than the on-disk size?

internal/server/storage/pool_interface.go
```diff
@@ -143,7 +143,7 @@ type Pool interface {
 	MigrateCustomVolume(projectName string, conn io.ReadWriteCloser, args *migration.VolumeSourceArgs, op *operations.Operation) error
 
 	// Custom volume backups.
-	BackupCustomVolume(projectName string, volName string, writer instancewriter.InstanceWriter, optimized bool, snapshots bool, op *operations.Operation) error
+	BackupCustomVolume(projectName string, volName string, contentType drivers.ContentType, writer instancewriter.InstanceWriter, optimized bool, snapshots bool, op *operations.Operation) error
```
stgraber: It feels pretty weird from an API point of view having to pass in the contentType.

I understand that we technically already have it in the caller and that saves us a second lookup, but I think I'd keep this function signature the way it was to keep that (internal) API clean.

bensmrs: Alright

stgraber:
> * Should I drop the `import > export` idempotence or should I rather query the DB for the ISO size?
> 
> * Why in the current code don’t we use the actual volume size in the export rather than the on-disk size?

In the case of an ISO which is write-once, it'd be safe to truncate it to the size we have in the database. For any other volume type, not so much as actual data (GPT table mostly) may have been written all the way at the end of the volume (post-rounding/alignment).

bensmrs:
> In the case of an ISO which is write-once, it'd be safe to truncate it to the size we have in the database. For any other volume type, not so much as actual data (GPT table mostly) may have been written all the way at the end of the volume (post-rounding/alignment).

Oh I see. But doesn’t it mean that we’re exposing in the API a size that may be wrong? (thinking about vol.config["size"])

bensmrs:
Nah, it really looks like `vol.config["size"]` is only exposed for ISO volumes, correct me if I’m wrong.

bensmrs:
I’ll add a commit to patch the command description tomorrow

stgraber:
@bensmrs can you add an API extension for this one?

bensmrs:
I’m always confused whether we need an extension or not… Here, clients won’t behave differently (as in: it’s the server that tells whether things worked; the clients don’t have to adapt) with or without this extension, so what’s exactly the point?

stgraber:
@bensmrs sorry, needs a rebase because of the API extension.

stgraber:
> I’m always confused whether we need an extension or not… Here, clients won’t behave differently (as in: it’s the server that tells whether things worked; the clients don’t have to adapt) with or without this extension, so what’s exactly the point?

The reason for this one is so for example a web UI can tell whether to show a download button next to that ISO image or not.

Doesn't really matter for the CLI as we had a clear error, but it does allow for providing a better user experience in some situations.

bensmrs:
Ok thanks for the clarification

PR #2529, opened 2025-10-03
Add support for SPICE audio

bensmrs:
As discussed in https://discuss.linuxcontainers.org/t/incus-are-we-macos-yet/22117/14


stgraber:
@bensmrs slight wrinkle on this one is that I think it will break all live migrations of existing VMs that don't yet have the device.

We've done this kind of breakage in the past but typically limited to just specific Windows versions so didn't have to care too much, but this one is going to break everyone.

I see two ways to handle that:

 1) We go the lazy route and we add a hardware revision number type thing as a volatile key, then we can check against that on migration receive and skip some devices based on that. The problem is that backports will get tricky and we'll have to keep backward compatibility code around potentially for a long time.
 2) We instead dump a text representation of the hardware used for the VM as a JSON file in the instance config drive. We can then pick that back up on the target and apply that back on the target. In practice, that basically means saving the qemu.conf file as part of a JSON file and using that instead of a newly generated one.

stgraber:
Obviously, I'd lean towards 2). We'd need to check what kind of variable data we have in the generated qemu.conf right now that would need to be changed on the migration target.

That won't include anything injected through QMP, but that's probably fine for now as generally, changes of devices breaking migration have all been on devices in qemu.conf

bensmrs:
Ok, I’ll have a look. Although I’m not sure that would break stuff as the device shouldn’t take a PCI slot otherwise taken (as it’s added after those on `qemu_pcie0`). What exactly could break?

stgraber:
Would be worth testing, but in my experience, you get a nasty QEMU error on migration receive if the PCIe layout differs as the guest would otherwise get restored on different hardware causing an immediate kernel panic.

stgraber:
Ah, we'll also need an API extension on this one once we sort out the live-migration between version issue.

Btw, you can probably test this with:
 - Start the VM on a version without the new device
 - Run `incus stop --stateful` on it
 - Update Incus to the version with the new device
 - Run `incus start` to restore the state

bensmrs:
Ok interesting, I’d have expected nasty stuff for different enumerations and other weird remappings, not for a single device present or absent in the end. But yeah, let’s not play with fire here.

bensmrs:
Well that’s sad…
```
root@incus-dev-2:~/incus# incus stop --stateful trixie
Error: Failed initializing state save to "/var/lib/incus/virtual-machines/trixie/state": GenericError: State blocked by non-migratable device '0000:00:01.0:00.7/virtio-sound-device'
```

bensmrs:
How about introducing a mechanism to hot-unplug some devices before migration, and hot-plug them back after? I’d eyeball that for this particular audio device, it shouldn’t bite us. And in case of a version mismatch, the device just won’t get plugged back and that’s it.

stgraber:
> How about introducing a mechanism to hot-unplug some devices before migration, and hot-plug them back after? I’d eyeball that for this particular audio device, it shouldn’t bite us. And in case of a version mismatch, the device just won’t get plugged back and that’s it.

That comes with a rather high risk of kernel panics in the guest though.
Removing a sound card that's in use, or the vsock device that's currently being used, or even worse for some of our previous changes, the IOMMU controller, I don't expect many of those to be too successful.

bensmrs:
So much for PCIe being “hot-pluggable” :)
Well, at least in my tests, I didn’t crash my Trixie VM when doing `device_del`, but I can’t speak for other distros and OSes.
What about only enabling SPICE audio when `migration.stateful` is not `true` then? If you think that would be too specific, we can close this PR for now and wait/pray for the QEMU devs to implement migration of `virtio-sound-pci` devices.

stgraber:
Yeah, that's fine, we can add it only to instances with migration.stateful=false.

This also currently avoids having to deal with the rest of the migration problems, convenient ;)

bensmrs:
Alright. Do we still need an API extension given all these restrictions?

stgraber:
> Alright. Do we still need an API extension given all these restrictions?

I don't think so given there's not much use for one from a client perspective here.

Issue #2505, opened 2025-09-25
Add a "detached" property for devices


nodens:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

Hi,

I'm often using a VM to update usb devices that would usually be connected to the host instead.
It's a bit cumbersome to find out the properties of the device via incus info --resources every time to then do `incus config device add <vm> usb vendorid=XX productid=YY` for the upgrade, and then remove it to make it available to the host again, only to have to add it again next time a config/firmware update tool need to be launched from the VM.

### What do you think would need to be added

A device property like `detached: <bool>` would allow to attach/detach the devicer without looking up every time.

Of course, one could script this, but it'd be nice to not have to :)

Alternatively, being able to attach/detach usb device via the web UI would make things easier as well.

Thanks for considering!

bensmrs:
Salut nodens,

Yeah, I think that’d be a great addition, and it echoes nicely with the `attached` property that I’ve already implemented for `disk` devices. @stgraber, I can take this one (only for USB devices first, maybe others if it’s not too cumbersome, with a property named `attached` instead of `detached`).

nodens:
Salut Ben!

attached definitely make even more sense, I was thinking backward. Thanks! :)

stgraber:
> Salut nodens,
> 
> Yeah, I think that’d be a great addition, and it echoes nicely with the `attached` property that I’ve already implemented for `disk` devices. @stgraber, I can take this one (only for USB devices first, maybe others if it’s not too cumbersome, with a property named `attached` instead of `detached`).

Yeah, that sounds good to me.

PR #2530, opened 2025-10-03
Add `attached` key for `usb` devices

bensmrs:
Closes: #2505


bensmrs:
While writing the PR, I noticed that https://github.com/lxc/incus/blob/9457a01ef564e6ef16fb9a3fd8da329fa3575526/internal/server/device/disk.go#L1777-L1789 should never actually be executed, as `attached` is not in `UpdatableFields`. I’ll investigate more.

stgraber:
@bensmrs we need an API extension on this one

stgraber (approved):

Issue #2486, opened 2025-09-16
Move documentation for storage-pool configs to gendoc


breml:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

Information about config keys for storage in general and storage pools in particular is not available on the metadata API `/1.0/metadata/configuration`. In particular I need the information, which config keys do have `local` scope and which have `global` scope.

### What do you think would need to be added

Based on conversation in https://github.com/FuturFusion/operations-center/pull/214/files#r2348871369, moving the documentation for storage pools to gendoc would resolve this issue.

PR #2539, opened 2025-10-04
Add gendoc for storage config keys

presztak:
Closes: #2486

### DIFF
--- a/doc/config_options.txt
+++ b/doc/config_options.txt
@@ -5177,3 +5177,1515 @@ This value is required by some providers.
 ```
 
 <!-- config group server-openfga end -->
+<!-- config group storage_btrfs-common start -->
+```{config:option} btrfs.mount_options storage_btrfs-common
+:default: "`user_subvol_rm_allowed`"
+:scope: "global"
+:shortdesc: "Mount options for block devices"
+:type: "string"
+
+```
+
+```{config:option} size storage_btrfs-common
+:default: "auto (20% of free disk space, >= 5 GiB and <= 30 GiB)"
+:scope: "local"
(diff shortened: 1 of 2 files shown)


internal/server/storage/drivers/driver_ceph_volumes.go
```diff
@@ -837,13 +837,116 @@ func (d *ceph) FillVolumeConfig(vol Volume) error {
...
 func (d *ceph) commonVolumeRules() map[string]func(value string) error {
 	return map[string]func(value string) error{
-		"block.filesystem":    validate.Optional(validate.IsOneOf(blockBackedAllowedFilesystems...)),
+		// gendoc:generate(entity=storage_volume_ceph, group=common, key=block.filesystem)
+		//
+		// ---
+		//  type: string
+		//  condition: block-based volume with content type `filesystem`
+		//  default: same as `volume.block.filesystem`
+		//  shortdesc: {{block_filesystem}}
```
stgraber, line 846: What does that do?

stgraber, line 846: Were you able to confirm that this still renders correctly in Sphinx?

stgraber:
@presztak looks good but we also need to add an extra field on all the keys that are server-specific as that's the main thing that @breml needs out of this.

We have that already done for the server configs where we use `scope: global` or `scope: local` to indicate whether they are server specific or not.

internal/server/storage/drivers/driver_ceph_volumes.go
presztak: Yup, still renders correctly.

presztak:
@stgraber , `scope` key was added.

Issue #2182, opened 2025-06-07
systemd credentials for containers


gwenya:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

[Systemd credentials](https://systemd.io/CREDENTIALS/) are key/value pairs that can be securely passed to systemd services. They are suitable both for actual secrets and for general configuration. 
In VMs, credentials can be passed from the host to the guest systemd via SMBIOS type 11 strings, which incus already supports nicely.
For containers, passing credentials from the host to the guest systemd works by mounting them into the container and passing the mount path in the environment variable `$CREDENTIALS_DIRECTORY`. This is somewhat supported by incus since arbitrary environment variables and mounts are possible, but it requires the user to manually prepare the credentials in a host directory or volume which is rather inconvenient.

### What do you think would need to be added

Add a new disk device source:

```yaml
devices: 
  credentials:
    type: disk
    source: credentials
    credentials-binary.foo: Zm9vCg==
    credentials.password: hunter2
    path: /run/incus-credentials
```

The user could then configure the `CREDENTIALS_DIRECTORY` environment variable to be `"/run/incus-credentials"` so that systemd can pick them up.

These credentials would only be as secure as the instance config is, i.e. anyone with read access via incus or directly to the incus database can extract them. At a later point they could be extended with (more secret) credentials from some external store (e.g. systemd on the host itself, or something like Hashicorp Vault / OpenBao).

stgraber:
I think I'd prefer something like `systemd.credential.foo: bar` or `systemd.credential-binary.foo: 123==` as instance config keys.

For containers, we'd handle passing a read-only bind-mount into a safe path, typically we abuse /dev for that as that's the only writable mountpoint we can guarantee will be present (/run is mounted by the container during boot, we can't use it safely).

So probably something like `/dev/.incus_systemd_credentials`.

For VMs, we can automatically turn those into the relevant SMBIOS11 so that they work there too.

gwenya:
Ah yeah for sure! I had thought about it more generic so that it’s not tied to the systemd name and maybe more useful with non-systemd containers, but this would certainly make for a nicer UI.

I haven’t looked at how we do mounts in general, but I was thinking of it as just another mount (with the target being a ramfs created outside the container), and those seem to work with paths that don’t exist before boot, though I suppose this one would have to be there early enough in the boot process for systemd to pick it up, so that might be tricky. I think I’ll take a closer look at how nspawn does it.

[...]

PR #2545, opened 2025-10-06
Add systemd credentials

bensmrs:
Closes: #2182

This PR is not ready yet, but I need some help to wrap it up. First, does my LXC attempt look reasonable? Second, as soon as I edit the LXC instance config, the bind mount is cleared:
```
root@trixie:~# ls -al /dev/.incus-systemd-credentials/
total 12
d--x------ 1 root root  18 Oct  6 23:23 .
drwxr-xr-x 9 root root 540 Oct  6 23:23 ..
-r-------- 1 root root  14 Oct  6 23:23 bar
-r-------- 1 root root   3 Oct  6 23:23 baz
-r-------- 1 root root   4 Oct  6 23:23 foo
root@trixie:~#
logout
root@incus-dev:~/incus# incus config set trixie user.xxx yyy
root@incus-dev:~/incus# incus shell trixie
root@trixie:~# ls -al /dev/.incus-systemd-credentials/
total 0
```
Any idea why, @stgraber? (the files are still there on the host)

Once I have a clearer view on that, I can start looking at the live update.

### DIFF
--- a/test/main.sh
+++ b/test/main.sh
@@ -391,6 +391,7 @@ if [ "${1:-"all"}" != "cluster" ]; then
     run_test test_storage_volume_recover "Recover storage volumes"
     run_test test_syslog_socket "Syslog socket"
     run_test test_incus_user "incus-user"
+    run_test test_systemd "systemd"
 fi
 
 # shellcheck disable=SC2034
--- a/test/suites/systemd.sh
+++ b/test/suites/systemd.sh
@@ -1,0 +1,63 @@
+assert_systemd_credentials_entries() {
+    entries=$(incus exec foo -- ls -A1 /dev/.incus-systemd-credentials | wc -l)
+    if [ "$entries" != "$1" ]; then
+        printf "Expected %s entries in systemd credentials directory; got %s\n" "$1" "$entries"
+        false
+    fi
+}
+
+assert_systemd_credentials_value() {
+    value=$(incus exec foo cat "/dev/.incus-systemd-credentials/$1")
+    if [ "$value" != "$2" ]; then
+        printf "Expected %s for systemd credential %s; got %s\n" "$2" "$1" "$value"
+        false
+    fi
+}
+
+test_systemd() {
+    ensure_import_testimage
+    ensure_has_localhost_remote "${INCUS_ADDR}"
+
+    incus launch testimage foo
+    stat=$(incus exec foo -- stat -c%a /dev/.incus-systemd-credentials)
+    if [ "$stat" != "100" ]; then
+        echo "Wrong permissions on systemd credentials directory"
+        false
+    fi
+
+    assert_systemd_credentials_entries 0
+
+    # Regular credential
+    incus config set foo systemd.credential.foo bar
+    assert_systemd_credentials_entries 1
+    assert_systemd_credentials_value foo bar
+
+    # Base64 credential
+    incus config set foo systemd.credential-binary.xxx eXl5
+    assert_systemd_credentials_entries 2
+    assert_systemd_credentials_value foo bar
+    assert_systemd_credentials_value xxx yyy
+
+    # Mutually exclusive credential and credential-binary keys
+    ! incus config set foo systemd.credential-binary.foo YmF6========= || false
+    incus config unset foo systemd.credential.foo
+
+    # Base64 credential with superfluous padding
+    incus config set foo systemd.credential-binary.foo YmF6=========
+    assert_systemd_credentials_entries 2
+    assert_systemd_credentials_value foo baz
+    assert_systemd_credentials_value xxx yyy
+
+    # Consistency after reboot
+    incus restart foo --force
+    assert_systemd_credentials_entries 2
+    assert_systemd_credentials_value foo baz
+    assert_systemd_credentials_value xxx yyy
+
+    # Credential deletion
+    incus config unset foo systemd.credential-binary.foo
+    assert_systemd_credentials_entries 1
+    assert_systemd_credentials_value xxx yyy
+
+    incus rm -f foo
+}


stgraber:
That's pretty odd. The only thing I can really think of is that we may have some logic on Update() which attempts to activate/deactivate the instance volume and is somehow causing the unmount, but that's pretty weird.

bensmrs:
So, the problem is indeed in `Update`: https://github.com/lxc/incus/blob/3d48ab18b9486feedb0bf36b4ce62ee48612265e/internal/server/instance/drivers/driver_lxc.go#L4872

I think because I’m deleting the directory and recreating it, the bindmount still points to the deleted entry.

So, not sure if I should continue doing my initialization in `initLXC`, and if so, if I can **really** rely on `cConfig` to tell me that the initialization has already been done (because the reverter in `Update` resets it)…

bensmrs:
At least now it works. I don’t know how that would play with live-migration, but I also don’t think I should really bother, as live-migrating a systemd container is a very bad idea…

bensmrs:
I’m not sure all those failing tests are related to the PR

stgraber:
I think they are, they're all failing in the same way and the only stuff that's passing is cluster tests where no mount is involved (dir/btrfs).

bensmrs:
Hmmm ok, I’ll run the test suite locally then.

bensmrs:
Ok there’s something:
```
root@incus-dev:~/incus/test# grep -r .incus-systemd-credentials tmp.oJD/SuiwKDXIw
tmp.oJD/SuiwKDXIw/run/c1/lxc.conf:lxc.environment = CREDENTIALS_DIRECTORY=/dev/.incus-systemd-credentials
tmp.oJD/SuiwKDXIw/run/c1/lxc.conf:lxc.mount.entry = /root/incus/test/tmp.oJD/SuiwKDXIw/containers/c1/credentials dev/.incus-systemd-credentials none bind,ro,create=dir 0 0
tmp.oJD/SuiwKDXIw/logs/c1/lxc.log:lxc c1 20251007192936.694 ERROR    utils - ../src/lxc/utils.c:safe_mount:1330 - No such file or directory - Failed to mount "/root/incus/test/tmp.oJD/SuiwKDXIw/containers/c1/credentials" onto "/usr/lib/x86_64-linux-gnu/lxc/rootfs/dev/.incus-systemd-credentials"
tmp.oJD/SuiwKDXIw/logs/c1/lxc.log:lxc c1 20251007192936.694 ERROR    conf - ../src/lxc/conf.c:mount_entry:2217 - No such file or directory - Failed to mount "/root/incus/test/tmp.oJD/SuiwKDXIw/containers/c1/credentials" on "/usr/lib/x86_64-linux-gnu/lxc/rootfs/dev/.incus-systemd-credentials"
```
But
```
root@incus-dev:~/incus/test# ls -al /root/incus/test/tmp.oJD/SuiwKDXIw/containers/c1/credentials
total 0
d--x------ 2 1000000 1000000 40 Oct  7 19:29 .
d--x------ 3 root    root    60 Oct  7 19:29 ..
```

This part however, `/usr/lib/x86_64-linux-gnu/lxc/rootfs/dev/.incus-systemd-credentials` is very intriguing (well, to me at least).

stgraber:
> This part however, `/usr/lib/x86_64-linux-gnu/lxc/rootfs/dev/.incus-systemd-credentials` is very intriguing (well, to me at least).

That's normal, /usr/lib/x86_64-linux-gnu/lxc/rootfs/ is the temporary mount point of the instance's root disk before pivot_root and such.

stgraber:
The error could happen on either side of the mount, so it could be because the source of the mount is somehow unavailable at the time the mount is attempted or because the mount target doesn't exist.

Now `create=dir` should in theory take care of the target part, making the source more likely to be an issue though I'm not too sure what may be going on there.

bensmrs:
`,optional` seems to help quite a bit, but to be honest, I don’t really understand what I’m doing on that mount line…
I’m looking at the new errors.

stgraber:
Definitely shouldn't need `optional` as that would make it skip the missing source path which isn't really what we want here, we always want that mount.

Did you figure out a simple reproducer for the failure? Then I can take a look at it.

bensmrs:
Ok I’ll revert the recent changes and I’ll try to produce a MWE. This is probably a rather trivial oversight; I’m quite new to the LXC driver codebase, and I’m still figuring out its general structure.

Edit: looks like it fails with tests using ZFS

bensmrs:
I can reproduce the problem on `lvm` and `zfs` drivers, but not `dir` or `btrfs`:
```
root@incus-dev:~/incus# incus launch testimage test-1 --profile dir
Launching test-1
root@incus-dev:~/incus# incus launch testimage test-2 --profile zfs
Launching test-2
Error: Failed to run: /root/go/bin/incusd forkstart test-2 /var/lib/incus/containers /run/incus/test-2/lxc.conf: exit status 1
```

So it looks like something is mounted too late when working with logical volume managers, but I’m having some trouble understanding the whole LXC driver dance for now.

stgraber:
Okay, I should be able to easily test it here on ZFS then.

bensmrs:
Nice. Please use c87d66686c24e08e61a30f4c214409975e7a810c, as it doesn’t include my desperate attempts :)

stgraber:
I found the problem:

```
root@dakara:~# ls /var/lib/incus/containers/d13/
backup.yaml  metadata.yaml  rootfs  templates
root@dakara:~# umount /var/lib/incus/containers/d13/
root@dakara:~# ls /var/lib/incus/containers/d13/
credentials
root@dakara:~# 
```

So basically the `credentials` folder was created prior to the instance's volume being mounted.
So when the instance volume does get mounted, it hides the `credentials` folder, causing the error.

internal/server/instance/drivers/driver_lxc.go
```diff
@@ -1068,9 +1071,81 @@ func (d *lxc) initLXC(config bool) (*liblxc.Container, error) {
...
+	}
+
+	for oldCredential := range oldCredentials {
+		err = os.Remove(filepath.Join(credentialsDir, oldCredential))
+		if err != nil {
+			return nil, fmt.Errorf("Failed to remove credential %q: %w", oldCredential, err)
 		}
 	}
 
+	err = lxcSetConfigItem(cc, "lxc.mount.entry", fmt.Sprintf("%s dev/.incus-systemd-credentials none bind,ro,create=dir,optional 0 0", credentialsDir))
```
stgraber: We'll want to drop the `optional` from there once the issue is sorted out.

internal/server/instance/drivers/driver_lxc.go
```diff
@@ -1068,9 +1071,81 @@ func (d *lxc) initLXC(config bool) (*liblxc.Container, error) {
...
+			credentials[after] = data
+		}
+	}
+
+	err = lxcSetConfigItem(cc, "lxc.environment", "CREDENTIALS_DIRECTORY=/dev/.incus-systemd-credentials")
+	if err != nil {
+		return nil, err
+	}
+
+	credentialsDir := filepath.Join(d.Path(), "credentials")
```
stgraber: Basically the whole `credentials` folder interaction should be moved to `startCommon`.
The lxcSetConfigItem calls are still fine to stay where they are, but all the filesystem interactions must be run from a spot where we're confident that everything is mounted.

bensmrs:
Oh, makes perfect sense! Thanks a lot, I’ll review that tomorrow.

internal/server/instance/drivers/driver_lxc.go
bensmrs: So this means that `Update` will have additional code to handle credentials updates. I thought I was being clever by stashing everything in `initLXC`, thus getting updates for free. It unfortunately also came with bugs for free. Oh well.

internal/server/instance/drivers/driver_lxc.go
bensmrs: Sure

internal/server/instance/drivers/driver_lxc.go
stgraber: Ah yeah, you may need to move the logic into a separate function and call it from both places.

bensmrs:
Aaaaah it looks good now.
I’m not 100% sure about the idmap, and I don’t fully grasp the implications of live changes of the idmap on that mounted directory. Any pointers?
I’ll write a few tests in the meantime.

bensmrs:
~~The tests are showing some rather strange race conditions on update (or maybe `setupCredentials` not getting called at all) where the DB gets updated but not the FS files~~

Edit: the tests have shown something that I completely overlooked: if a user sets both `systemd.credential.foo` and `systemd.credential-binary.foo`.

bensmrs:
Great, should be ready to review now

PR #2554, opened 2025-10-12
mount sysfs as read-only

rossmeier:
With a recent update to systemd 258 in my containers, I noticed `console-getty.service` not working anymore with some kind of spawn error. With help from the lovely people over at systemd (https://github.com/systemd/systemd/issues/39036), I tracked this down to incus mounting `/sys/` as read-write which does not comply with systemd's specified container interface described in https://systemd.io/CONTAINER_INTERFACE/.

For this PR, I found all places (hopefully) that mount /sys as read-write and changed them to readonly which has fixed the issues with systemd for me so far. I personally never used either the hidden `forknet` command or the apparmor feature so I can't tell if those respective changes even work. Some more testing is definitely required.

### DIFF
--- a/cmd/incusd/main_forknet.go
+++ b/cmd/incusd/main_forknet.go
@@ -64,7 +64,7 @@ static void forkdonetdetach(char *file) {
 		_exit(1);
 	}
 
-	if (mount("sysfs", "/sys", "sysfs", 0, NULL) < 0) {
+	if (mount("sysfs", "/sys", "sysfs", MS_RDONLY, NULL) < 0) {
 		fprintf(stderr, "Failed mounting new sysfs: %s\n", strerror(errno));
 		_exit(1);
 	}
--- a/internal/server/apparmor/instance_lxc.profile.go
+++ b/internal/server/apparmor/instance_lxc.profile.go
@@ -88,7 +88,7 @@ profile "{{ .name }}" flags=(attach_disconnected,mediate_deleted) {
 
   # Handle sysfs (access handled below)
   mount fstype=sysfs -> /sys/,
-  mount options=(rw,nosuid,nodev,noexec,remount) -> /sys/,
+  mount options=(ro,nosuid,nodev,noexec,remount) -> /sys/,
 
   # Handle /run remounts.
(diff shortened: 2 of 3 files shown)


stgraber:
We're aware of the systemd recommendations but they are not fit for the kind of system containers that we run. 

Our containers run udev, they run network management tools, they run a whole bunch of other software that very much require a writable /sys.

The issue you mention with the console unit has already been worked around in our newer images through a systemd override on that unit.

stgraber:
I'll have to check if the issue happens consistently on systemd 258 across all distros as that would then make it much easier for us to work around. 

We always ship a systemd generator script in our images which amongst other things deals with those kind of issues by generating overrides on the fly. 

The console-getty workaround is currently done in the image build script because it only affected a couple of distros at the time, but if it's going to hit everyone with systemd 258, then we should move it to the generator script instead.

rossmeier:
Thanks for the clarification, sorry for the PR in that case. I couldn't find any hint on the console-getty issue though. Could you point to the change so I can update my deployments? Also do those generator scripts get updated in a way for existing containers?

stgraber:
The generator script doesn't get automatically updated but it is at least easy to manually update it by grabbing the current version from the distrobuilder GitHub. 

The workaround for the getty-console issue can be found here: 

https://github.com/lxc/lxc-ci/commit/db6a7e40c8fc1d8ee2f895ea10030df2a9cec156

rossmeier:
Thanks for the link. I tried deploying this and couldn't get it to work. It actually finally worked with this file:
```
[Service]
StandardInput=null
StandardOutput=null
ExecStart=
ExecStart=-/sbin/agetty --noreset --noclear --issue-file=/etc/issue:/etc/issue.d:/run/issue.d:/usr/lib/issue.d --keep-baud 115200,57600,38400,9600 console
```
Verifying by launching a fresh archlinux instance showed that `console-getty` is actually broken in the current `archlinux` image...
In any way, since the `console-getty` seems to be a standard systemd feature using the getty generator, I assume that more distros using systemd will run into this problem as they get updated to 258

stgraber:
https://github.com/lxc/distrobuilder/pull/949

PR #2561, opened 2025-10-14
internal/server/network: Update libovsdb import path

gibmat:
Should there also be a commit with the results of running `make update-ovsdb`? (It's pretty large.)


stgraber:
@gibmat Yeah, I'd do it in 4 commits:
 - Switch import paths in Go
 - Update Makefile
 - Update gomod
 - Update generated ovsdb stuff

PR #2564, opened 2025-10-16
More work on Incus agents

bensmrs:


bensmrs:
@stgraber, I’m gonna need some help in the agent selection logic. For ISO mounts, the agent is selected in an architecture-dependent way, which is not used for 9p mounts. Could you please have a look?

stgraber:
> @stgraber, I’m gonna need some help in the agent selection logic. For ISO mounts, the agent is selected in an architecture-dependent way, which is not used for 9p mounts. Could you please have a look?

Yeah, the reason for that is pretty simple. In the 9p case, we can just share a folder on the host containing all the agent builds (`/opt/incus/agent/` in Zabbly packages). That's cheap as it's the same data available for everyone through QEMU's built-in sharing.

For ISOs, now we need to actually generate a .iso for every instance using that. We don't want to waste space as each agent build can be around 20MB, so instead we expect the ISO to only contain the binary the instance needs. We should know the architecture of the instance already (d.architecture) and we can tell the OS through `image.os`.

internal/server/device/disk.go
```diff
@@ -2886,14 +2886,16 @@ func (d *disk) generateVMAgentDrive() (string, error) {
...
-		var dstFilename string
+		dstFilename := "incus-agent"
+		imageOS := strings.ToLower(d.inst.ExpandedConfig()["image.os"])
 
-		if strings.Contains(strings.ToLower(d.inst.ExpandedConfig()["image.os"]), "windows") {
+		if strings.Contains(imageOS, "windows") {
 			srcFilename = fmt.Sprintf("incus-agent.windows.%s", d.state.OS.Uname.Machine)
 			dstFilename = "incus-agent.exe"
+		} else if strings.Contains(imageOS, "darwin") || strings.Contains(imageOS, "macos") || strings.Contains(imageOS, "mac os") {
+			srcFilename = fmt.Sprintf("incus-agent.darwin.%s", d.state.OS.Uname.Machine)
```
stgraber: This will be somewhat user visible, should we just use `macos` here instead of `darwin`?
I don't really expect anyone to set `image.os` to something other than `Mac OS` or `MacOS` and while technically the underlying OS name is Darwin, I think it will just confuse most folks :)

internal/server/device/disk.go
```diff
@@ -2886,14 +2886,16 @@ func (d *disk) generateVMAgentDrive() (string, error) {
 	// Include the most likely agent.
 	if util.PathExists(os.Getenv("INCUS_AGENT_PATH")) {
 		var srcFilename string
-		var dstFilename string
+		dstFilename := "incus-agent"
+		imageOS := strings.ToLower(d.inst.ExpandedConfig()["image.os"])
 
-		if strings.Contains(strings.ToLower(d.inst.ExpandedConfig()["image.os"]), "windows") {
+		if strings.Contains(imageOS, "windows") {
 			srcFilename = fmt.Sprintf("incus-agent.windows.%s", d.state.OS.Uname.Machine)
```
stgraber: I see what you mentioned about architecture.

We indeed shouldn't be using `d.state.OS.Uname.Machine` here but instead use `d.inst.Architecture()` to get the correct architecture name for the instance.

internal/server/instance/drivers/driver_qemu.go
```diff
@@ -3473,17 +3514,22 @@ func (d *qemu) isWindows() bool {
...
+			return true
 		}
 	}
 
-	return true
+	return false
+}
+
+// isLinux returns whether the VM is Linux. This is the default if no `image.os` is provided.
+func (d *qemu) isLinux() bool {
```
stgraber: I wonder if it would make sense to have something more format about the Guest OS.

Basically a `inst.GuestOS()` function in `driver_common.go` which does the `image.os` parsing and gives us one of:
- `linux`
- `macos`
- `windows`
- `unknown`

So we can then rely on that everywhere we need to figure out what the OS is both here and in the devices.

bensmrs:
> Yeah, the reason for that is pretty simple. In the 9p case, we can just share a folder on the host containing all the agent builds (`/opt/incus/agent/` in Zabbly packages). That's cheap as it's the same data available for everyone through QEMU's built-in sharing.

Oh I actually missed that `incus-agent` is a script to call the right agent. In my tests, I just injected the macOS agent binary. I’ll look into it.

internal/server/device/disk.go
bensmrs: I have no strong opinion. I think that techies wouldn’t mind reading `darwin` there, and the agent would technically work on **all** (*cough cough*) Darwin distributions, but I get that it may also be pedantic purism.

internal/server/instance/drivers/driver_qemu.go
bensmrs: Except in the driver’s logic, `linux` and `unknown` are the same thing, and it’s not reasonable to keep a list of distros. If you decide to package Rebecca Black OS on the image server, we’d have to also patch Incus to include it, which is not really fun for the one having to maintain this mess.

stgraber: Yeah, you're right, we don't really have a single flag to tell us we're dealing with Linux currently as the images give the distro as the OS name, so that's a bit impractical.

Though we could still use that logic in general. For containers we could always confidently return `linux`. For VMs, they'd get the `unknown` value with the code then treating `linux` and `unknown` as equivalent.

bensmrs:
Oooh I actually got lost in the agent dance and overlooked that the `config` and `agent` shares are actually different!
So there’s a little more work to do, and an additional splitting for the `incus-agent` script between a Linux-specific one and a Darwin-specific one.
I placed myself in the very convenient case where `/Volumes/config/incus-agent` is the actual agent binary (and it works fine).

internal/server/device/disk.go
stgraber: Yeah, I'd probably use `macos`, that would line up with the name we use for all our binaries in releases and such. If there were a bunch of Darwin based OSes out there, that'd be different, but that's not really a thing ;)

bensmrs:
Ok that should be good now, both `config` and `config+agent` setups should be supported. I haven’t taken any care to handle CDROM installation, because macOS supports 9p, so not really worth the time I think.

PR #2577, opened 2025-10-22
incusd/instance/lxc: Fix handling of credentials on existing instances

stgraber:
This fixes an issue where profile updates fail on instances that were started prior to the introduction of the credentials feature.


bensmrs:
Oh good catch!

Issue #2582, opened 2025-10-23
Allow custom MAC address prefixes


bensmrs:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

Companies having their own OUIs or managing/namespacing resources at the MAC level, and wanting to use those MAC addresses in Incus, have to systematically add overrides so that they don’t use Zabbly’s OUI.

### What do you think would need to be added

It makes sense to have custom MAC address templates defined globally, per project, and per network (per node doesn’t sound right). So I suggest adding a `network.hwaddr.template` configuration key to global, project and network config (with network overriding project, itself overriding global), defaulting to `10:66:6a:xx:xx:xx`. If that looks ok to you, I can take this one.

stgraber:
Should be `instances.nic.hwaddr` to match the existing `instances.nic.host_name`.
Default value would then be the Zabbly OUI that's reserved for this use (`10:66:6a:xx:xx:xx`) with administrators being able to change it for future MAC allocations.

bensmrs:
> Should be `instances.nic.hwaddr` to match the existing `instances.nic.host_name`.

Ok I’m definitely learning about this one only now :)
Would it also be prefixed with `instances.` for project and network config? That only sounds right for server config. Then again, without the `instances.` prefix, `nic.hwaddr` would only sound right for project config. OTOH, having a config key named differently but that behaves the same would be a semantic nightmare.

stgraber:
Oh right, I kinda forgot about the network case too and wherever else we may be using MACs.
So yeah, we may need to put that one under `network.hwaddr_pattern` to have something more generic.

stgraber:
So your initial suggestion was right but it reminded me of the other similar case we have with the host device so I was hoping it'd fit the same pattern. Sadly, that won't work given that we also give MAC addresses to things other than instances :)

bensmrs:
Currently working on it. I’m wondering, in the case of `network.hwaddr_pattern` set on network, if it should impact the network’s hwaddr (i.e. in the case of a bridge network) or not. Setting the property means “every network device plugged to this network should use the following pattern”, but it’s ambiguous whether the network itself should be included in the rule or not. What’s your opinion on this?

stgraber:
Yeah, I think it should affect anything that gets a generate MAC, including networks.

PR #2589, opened 2025-10-28
Allow custom MAC address prefixes

bensmrs:
Closes: #2582

I’m not sure if additional work is needed to properly include the keys in the documentation. Also, if `network.` is superfluous in network configuration, I can remove it. I kept it for the sake of consistency.
I went into a bit of detail regarding the birthday paradox; I can remove it if you want, although I think it’s important to mention.

### DIFF
--- a/test/suites/network_hwaddr_pattern.sh
+++ b/test/suites/network_hwaddr_pattern.sh
@@ -1,0 +1,53 @@
+test_network_hwaddr_pattern() {
+    ensure_import_testimage
+    ensure_has_localhost_remote "${INCUS_ADDR}"
+
+    incus network create "inct$$"
+    incus init testimage nettest -n "inct$$"
+    incus start nettest
+
+    # First, check for the default Zabbly MAC
+    ip link show "inct$$" | grep -q 'link/ether 10:66:6a:'
+    incus exec nettest -- ip link show eth0 | grep -q 'link/ether 10:66:6a:'
+    incus stop -f nettest
+
+    # Then, override the MAC pattern globally
+    incus config set network.hwaddr_pattern=00:11:22:xx:xx:xx
+    incus config unset nettest volatile.eth0.hwaddr
+    # Refresh the bridge
+    incus network set "inct$$" bridge.mtu=1500
+    ip link show "inct$$" | grep -q 'link/ether 00:11:22:'
+    incus start nettest
+    incus exec nettest -- ip link show eth0 | grep -q 'link/ether 00:11:22:'
+    incus stop -f nettest
+
+    # Finally, create a project and override its MAC pattern
+    incus project create foo
+    incus project switch foo
+    ensure_import_testimage
+    incus profile device add default root disk path="/" pool="incustest-$(basename "${INCUS_DIR}")"
+    incus project switch default
+    incus project set foo network.hwaddr_pattern=00:33:44:xx:xx:xx
+    incus config unset nettest volatile.eth0.hwaddr
+    incus start nettest
+    incus exec nettest -- ip link show eth0 | grep -q 'link/ether 00:11:22:'
+    incus stop -f nettest
+    incus init --project foo testimage nettest-foo -n "inct$$"
+    incus start --project foo nettest-foo
+    incus exec --project foo nettest-foo -- ip link show eth0 | grep -q 'link/ether 00:33:44:'
+    incus stop --project foo -f nettest-foo
+    incus delete --project foo -f nettest-foo
+
+    incus project set foo features.networks=true
+    if incus network create --project foo "inct$$-foo"; then
+        ip link show "inct$$-foo" | grep -q 'link/ether 00:33:44:'
+        incus network delete --project foo "inct$$-foo"
+    else
+        echo "==> SKIP: Skipping OVN tests"
+    fi
+
+    incus image remove --project foo testimage
+    incus project delete foo
+    incus delete -f nettest
+    incus network delete "inct$$"
+}
--- a/test/main.sh
+++ b/test/main.sh
@@ -337,6 +337,7 @@ if [ "${1:-"all"}" != "cluster" ]; then
     run_test test_address_set "network address set"
     run_test test_network_forward "network address forwards"
     run_test test_network_zone "network DNS zones"
+    run_test test_network_hwaddr_pattern "network MAC address pattern"
     run_test test_idmap "id mapping"
     run_test test_template "file templating"
     run_test test_pki "PKI mode"


test/suites/network_hwaddr_pattern.sh
```diff
@@ -0,0 +1,55 @@
...
+
+    # First, check for the default Zabbly MAC
+    ip link show "inct$$" | grep -q 'link/ether 10:66:6a:'
+    incus exec nettest -- ip link show eth0 | grep -q 'link/ether 10:66:6a:'
+    incus stop -f nettest
+
+    # Then, override the MAC pattern globally
+    incus config set network.hwaddr_pattern=00:11:22:xx:xx:xx
+    incus config unset nettest volatile.eth0.hwaddr
+    # FIXME: ip link show "inct$$" | grep -q 'link/ether 00:11:22:'
```
bensmrs: This doesn’t work because I can’t find a way to signal to the network that the server config has changed. Is it possible?
(Here, the bridge network should change its MAC address, as the parent MAC allocation pattern has changed.)

Similar question on project-specific networks when overriding a project’s config, although the current code doesn’t handle inheriting the project config for now (I missed this case).

bensmrs: We could also put that under the rug, as the volatile instance addresses don’t get changed. It’s just that adding `network.hwaddr_pattern` on a network triggers its config reload, thus a MAC address change, so the behavior is not consistent in this very case. But again, it’s not (and can’t be) consistent between networks and instances…

stgraber: My usual trick to trigger a full network restart is to mess with `bridge.mtu`.

Though from a testing point of view, I think checking that a new bridge has the correct MAC pattern is sufficient. Mangling of existing bridges by clearing the volatile key isn't exactly something we recommend users doing (we usually tell everyone not to touch volatile).

shared/validate/validate.go
```diff
@@ -254,12 +254,19 @@ func IsInterfaceName(value string) error {
...
+	if !match {
+		return errors.New("Invalid MAC address pattern, must be 6 bytes (hex characters or `x` to denote a wildcard) separated by colons")
+	}
+
+	return nil
+}
+
 // IsNetworkMAC validates an Ethernet MAC address. e.g. "00:00:5e:00:53:01".
 func IsNetworkMAC(value string) error {
-	_, err := net.ParseMAC(value)
```
stgraber, line 259: Hmm, please no, ParseMAC does more than validate formatting, it checks that the MAC is actually valid.

There are various reserved bit patterns in MAC addresses which lead to invalid MAC addresses, that's what we use ParseMAC to catch here.

internal/server/network/driver_common.go
```diff
@@ -122,7 +122,17 @@ func (n *common) FillConfig(config map[string]string) error {
 
 // validationRules returns a map of config rules common to all drivers.
 func (n *common) validationRules() map[string]func(string) error {
-	return map[string]func(string) error{}
+	return map[string]func(string) error{
+		// gendoc:generate(entity=network, group=common, key=network.hwaddr_pattern)
```
stgraber: @bensmrs do you have a strong need for this one?

My gut feeling is that MAC address pattern will in 90% of cases be an organizational decision, so it would go into the cluster config. I can see a case for multi-tenant cluster wanting the pattern changed on a per-project basis.

But I really don't see a use case for wanting to specify the pattern on a per-network basis. At that point, just set bridge.hwaddr to the MAC you want ;)

shared/validate/validate.go
bensmrs: Ok I didn’t look at all the things ParseMAC did

internal/server/network/driver_common.go
bensmrs: Sure but if you do that, it won’t impact the NICs within this network.
I have absolutely no need for this one, but we could imagine a cloud provider giving a MAC range to a customer (some providers do dark magic on L2), and they’d then want to configure a single network to use these, and another to use another one. No strong opinion.

shared/validate/validate.go
stgraber: I just looked now and annoyingly it doesn't appear to actually do the validation.

It supports the various syntaxes for MAC addresses which is nice but it doesn't actually check for the reserved bits. We should still use it though as validating those magic bits is something that they may/should add to that function...

stgraber: I'm pretty sure I got a nice clean error before from an invalid MAC but it may have come out of the kernel then.

`11:22:33:44:55:66` is a convenient invalid MAC, it has the multicast bit set so isn't suitable for a single device but only as a target for multiple devices.

test/suites/network_hwaddr_pattern.sh
bensmrs: Well, the bridge MAC address is not stored as a volatile key; I’d have messed with it in the tests else. I’ll change the MTU in the test to trigger the change, and we’ll just not perform the refresh on already-existing networks.

shared/validate/validate.go
bensmrs: > it may have come out of the kernel then

I think it’s this, yep (already had that one)

internal/server/network/driver_common.go
stgraber: Oh, I see, the hwaddr pattern you set on the network isn't really for the network itself, it's for the NICs that get attached to that network in the future.

That's interesting, I didn't consider that use case. My gut feeling is that this is extremely unlikely to be needed and more likely to cause confusion (as it did for me) ;)

I'd say, let's skip that bit and just keep it global or per-project, then if someone comes with an actual requirement for having a different pattern just on a specific network, we can add it back :)

bensmrs: Makes sense. I have a few patches to support project config in networks, so you can stop the review for now.

shared/validate/validate.go
bensmrs: > It supports the various syntaxes for MAC addresses which is nice

Yes, but we don’t. We could, though.
Anyway, I’m reverting this change.

bensmrs:
Ok that should be good now. I can’t test OVN on my environment, so I’ll wait for the GitHub tests to check if project overrides for project-specific networks work.

bensmrs:
Well that’s bad. Does OVN work in the testing environment?

stgraber:
The tests look quite sad :)

bensmrs:
Well for most of them I don’t think it’s my fault. I’m seeing `==> SKIP: Skipping OVN tests` in previous tests, so I guess that doesn’t really work in the CI…

stgraber:
```
==> Please update the metadata in your commit (make update-metadata)
==> Differences detected:
--- /tmp/tmp.Ska3eyFbAK/metadata-before.txt	2025-10-28 15:14:13.629821071 +0000
+++ /tmp/tmp.Ska3eyFbAK/metadata-after.txt	2025-10-28 15:14:21.150894383 +0000
@@ -1,2 +1,2 @@
-f8f4789874dffc0fef998148cf7837c3  doc/config_options.txt
-9ab581c78ed9a371732f30d475f6bf8d  internal/server/metadata/configuration.json
+56a4e28d6aa1fb055f67ce9bbfde2892  doc/config_options.txt
+4e151ff0798217066c18a04b3c326469  internal/server/metadata/configuration.json
```

bensmrs:
Yup **that one**’s on me

stgraber:
Right, `features.networks=true` won't work in the testsuite as we don't have OVN available in there. We can add a test to https://github.com/lxc/lxc-ci/blob/main/bin/test-incus-network-ovn to have it test as part of the OVN tests though.

Issue #2596, opened 2025-10-30
Automatically bring MACVLAN parents up


bensmrs:
### Is there an existing issue for this?

- [x] There is no existing issue for this bug

### Is this happening on an up to date version of Incus?

- [x] This is happening on a supported version of Incus

### Incus system details

```yaml
Irrelevant
```

### Current behavior

I’ve started playing with MACVLAN interfaces to remove the bridge overhead, and am facing an issue similar to #1281. My parent interface is configured with `systemd-networkd` with a link unit, but no network unit (I really don’t need / want my interface to have an IP). Doing so, systemd has no reason to bring the interface up, and surely enough, Incus instances cannot access the network without a manual `ip link set XXX up`.

### Expected behavior

As with bridge external interfaces, Incus should bring MACVLAN parent interfaces up.

### Steps to reproduce

Irrelevant

stgraber:
Yeah, would be fine to bring the parent up when starting a child device.

stgraber:
@bensmrs want to prepare a fix for it?

bensmrs:
How many hours until it’s no longer considered for the next release?

stgraber:
You probably got another 4-5 hours I'd guess. I'm dealing with some other issues and then need to figure out what we want to do about #2562 in this release...

bensmrs:
Yeah that one looks tough :)
Well 4 hours should be fine for me to take this one, should be a pretty trivial fix.

bensmrs:
Nice, I also found a little bug :)

PR #2597, opened 2025-10-30
Bring up parent interface and check existence on update

bensmrs:
Closes: #2596
Also adds a check for parent interface existence on config change


bensmrs:
Hmm looks like I’ve done too much JS and C recently

Issue #1037, opened 2024-07-23
Set correct SELinux context options on container & VM mounts


stgraber:
Per https://discuss.linuxcontainers.org/t/unable-to-add-lxc-rootfs-mount-options-context/20943/14 we should try and at least set the correct `context` mount options whenever we mount a container or VM data directory.

Ideally we'd have something more robust where we can handle per-instance labeling but that would require someone with more SELinux knowledge than I have to get this done properly, put testing in place and keep an eye on it so it doesn't regress :)

So for now, we'll try to go for low hanging fruits and just make things better where feasible.

mschiff:
Hi Stéphane,

in the meantime I have developed a (mostly) working SELinux-Module for incus! Getting containers and VMs running  on a SELinux system is not yet straight forward, but it is doable and working with one major issue which I am not able to solve without some basic incus support here.

It would we great to have some very basic support for this in incus as it would make running qemu VMs much better: 

The missing piece is the SELinux-labels of all related files for a specific VM.

A possibility to pass a context to incus which it will use to label VM specific dirs before creating supporting files and starting a VM would eliminate this last big roadblock.

Currently its like that:

after creating/starting a VM I get:
```
(sysadm_r)@host ~ # ls -lZ /var/lib/incus/devices/vm1/
dr-x------. 6 nobody root system_u:object_r:unlabeled_t:s0 14  6. Dez 17:59 config.mount

(sysadm_r)@host ~ # ls -lZ /var/lib/incus/devices/vm1/config.mount/
-r--------. 1 nobody root system_u:object_r:unlabeled_t:s0        2  6. Dez 17:59 agent-mounts.json
-rw-r--r--. 1 nobody root system_u:object_r:unlabeled_t:s0      819  6. Dez 17:59 agent.conf
-r--------. 1 nobody root system_u:object_r:unlabeled_t:s0      725  6. Dez 17:59 agent.crt
-r--------. 1 nobody root system_u:object_r:unlabeled_t:s0      288  6. Dez 17:59 agent.key
dr-x------. 2 nobody root system_u:object_r:unlabeled_t:s0        3  6. Dez 17:59 files
-r-x------. 1 nobody root system_u:object_r:unlabeled_t:s0 21721508 19. Nov 14:24 incus-agent
-rwx------. 1 nobody root system_u:object_r:unlabeled_t:s0     1302  6. Dez 17:59 install.sh
[...]

PR #2599, opened 2025-10-31
Initial SELinux detection and handling

stgraber:
Related to #1037 

This introduces some SELinux logic and makes it be mostly in line with AppArmor:
 - Adds a new `INCUS_SECURITY_SELINUX` environment variable
 - Adds new fields to the OS struct
 - Adds some basic detection logic to see if SELinux is present on the system
 - Adds some basic detection logic for what kind of SELinux policy/labels are in use
 - Detect the current Fedora-style policy
 - Configure LXC to set the correct context including instance-specific category

For now the logic is opt-in, so `INCUS_SECURITY_SELINUX` needs to be set to `true` for the logic to kick-in. This should let us progressively test and expand this logic before we flip it on by default.

I've done basic testing on Fedora 43 at this point. Unfortunately this doesn't use the new Incus module from the refpolicy (that @mschiff contributed upstream), so we'll need to add support for that too and see if we can't have the existing SELinux-enabled distros shipping Incus to use the refpolicy logic instead.

Incus refpolicy PR: https://github.com/SELinuxProject/refpolicy/pull/951/files

### DIFF
--- a/doc/.wordlist.txt
+++ b/doc/.wordlist.txt
@@ -262,6 +262,7 @@ scalable
 scriptlet
 SDN
 Seccomp
+SELinux
 SEV
 SFTP
 SHA


stgraber:
Because I don't have a system using the new refpolicy stuff and I'm just dealing with the basic logic that comes from the `incus-selinux` package on Fedora, the only useful thing I could do at this point was adding the category to the container's context.

With a better structured policy, we should be able to extend the OS logic to determine the correct context for both containers and VMS and also give us the expected labels to apply to the filesystems we initialize or mount.

The OS struct is convenient because it's available throughout the Incus codebase (part of the State struct), so we can get to it both from the instance drivers and from the storage side too.

PR #2613, opened 2025-11-04
client: Add GetEventsByType and GetEventsAllProjectsByType

breml:


stgraber:
@breml 

```
 Error: client/incus_events.go:67:2: import-shadowing: The name 'url' shadows an import name (revive)
	url, err := r.setQueryAttributes("/events?" + strings.Join(queryParams, "&"))
	^
Error: client/incus_events.go:180:1: exported: exported method ProtocolIncus.GetEventsByType should have comment or be unexported (revive)
func (r *ProtocolIncus) GetEventsByType(eventTypes []string) (listener *EventListener, err error) {
^
Error: client/incus_events.go:189:1: exported: exported method ProtocolIncus.GetEventsAllProjectsByType should have comment or be unexported (revive)
func (r *ProtocolIncus) GetEventsAllProjectsByType(eventTypes []string) (listener *EventListener, err error) {
^
3 issues:
* revive: 3
```

client/incus_events.go
```diff
@@ -48,20 +50,27 @@ func (r *ProtocolIncus) getEvents(allProjects bool) (*EventListener, error) {
...
+
+	if len(eventTypes) > 0 {
+		for i := range len(eventTypes) {
+			eventTypes[i] = url.QueryEscape(eventTypes[i])
+		}
+
+		queryParams = append(queryParams, "type="+strings.Join(eventTypes, ","))
 	}
 
+	eventsURL, err := r.setQueryAttributes("/events?" + strings.Join(queryParams, "&"))
```
stgraber, line 67: Isn't it a problem passing `/event?` even if no parameters are passed here?

breml: Why do you think this this is a problem? Semantically they are the same.
One place where this could make a difference is, if the request URL is put to a log file without "normalization". In this case, e.g. filtering by URL could return different results. So one could argue, that this is not clean. I can change it, if you wish.

stgraber: I didn't have time to recheck exactly what setQueryAttributes does and wanted to avoid a situation where we end up with multiple `?xyz=`, breaking the URL.

But the handling inside of seqQueryAttributes seems fine and it will normalize the URL anyway, so this should be fine.

PR #2622, opened 2025-11-05
Feat: --format raw added added for getting raw data.

### LINKED ISSUE #2611: Have an option for raw disk usage reported in "incus list -c nD --format=csv"

I want to do a pre-check with a script in bash to make sure the destination location has sufficient space before running `incus export`. 

In bash, the easiest method is to run `incus list -c nD --format=csv` to get the list of instances names and  sizes. However the size as reported by `incus list` converts to "human readable" of "GiB" instead of keeping it as a raw number. 

So I'd have to write scripts to look for MiB, GiB, TiB, etc instead of just comparing bytes as reported by df and incus. 

I can envision these options

1) modify the format flag to allow a 'raw' option as in `incus list -c nD --format=csv,raw` similarly to how there already exists "noheader" .   Maybe TableOptionRaw in `internal/cmd/table.go` ? 

2) modify incus/list.go `diskUsageColumnData`  to just report in raw format. While this is the easiest for me to do, it would break old behavior.  

3) create a new shorthand `-c R`  for "raw disk usage" instead of "human-readable disk usage" or `-c B` for disk usage in bytes?

Mujib-Ahasan:
## Description:
This PR is opened to introduce raw formatting in data.

## How to test it:
`incus list -c D --format=csv,raw` 
This could be used in other formatting( json, markdown, yaml etc )too.

close #2611

### DIFF
--- a/cmd/incus/list.go
+++ b/cmd/incus/list.go
@@ -39,6 +39,7 @@ type cmdList struct {
 	flagFast        bool
 	flagFormat      string
 	flagAllProjects bool
+	rawFormat       bool
 
 	shorthandFilters map[string]func(*api.Instance, *api.InstanceState, string) bool
 }
@@ -409,6 +410,15 @@ func (c *cmdList) Run(cmd *cobra.Command, args []string) error {
 		return errors.New(i18n.G("Can't specify --project with --all-projects"))
 	}
 
+	// Check for raw unit formatting.
+	formatFields := strings.SplitN(c.flagFormat, ",", 2)
+	if len(formatFields) == 2 {
+		formatOptions := strings.Split(formatFields[1], ",")
+		if slices.Contains(formatOptions, "raw") {
+			c.rawFormat = true
+		}
+	}
+
 	// Parse the remote
 	var remote string
 	var name string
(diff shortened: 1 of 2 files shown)


stgraber:
```
Error: cmd/incus/list.go:821:1: File is not properly formatted (gofumpt)

^
Error: cmd/incus/list.go:775:10: indent-error-flow: if block ends with a return statement, so drop this else and outdent its block (revive)
		} else {
			return strconv.FormatInt(cInfo.State.Memory.Usage, 10)
		}
Error: cmd/incus/list.go:816:10: indent-error-flow: if block ends with a return statement, so drop this else and outdent its block (revive)
		} else {
			return strconv.FormatInt(cInfo.State.Disk[rootDisk].Usage, 10)
		}
3 issues:
* gofumpt: 1
* revive: 2
```

Mujib-Ahasan:
hey @stgraber, no idea why this is failing now!

presztak:
Hey @Mujib-Ahasan , you are missing new lines after `if` blocks in these two places:
- cmd/incus/list.go:776
- cmd/incus/list.go:816

stgraber:
Thanks!

PR #2637, opened 2025-11-09
Fix sparse writer performance

stgraber:


cmd/incus-migrate/migrate_ova.go
```diff
@@ -311,11 +311,17 @@ func (m *OVAMigration) unpackOVA(outPath string) error {
...
+		for {
+			_, err = io.CopyN(outFile, tarReader, 4*1024*1024)
+			if err != nil {
+				if errors.Is(err, io.EOF) {
+					break
+				}
+
+				fmt.Println("Error extracting file:", err)
+				outFile.Close()
+				continue
```
hallyn, line 323: This 'continue' will continue the inner, not outer loop.  But you've closed outFile, and would only re-open it if you continued the outer loop.

stgraber: Oops, yeah, we need to continue the outer loop, not the inner one... I'll send a fix.

PR #2645, opened 2025-11-10
incusd: only apply qemu rtc adjustments if it is configured

The127:
This commit makes it so that we do not pass the -rtc option to qemu if it is configured. And also moves it down closer to the actual qemu call to reduce time shift.

We have noticed up to 7 seconds of discrepancy in some cases (without passing an adjustment in the configuration).


The127:
Not 100% sure if the if check makes sense. Incus might just always write a value in this field anyways. Either way moving this piece of code closer to the command execution should fix the issue.

stgraber (approved):

Issue #2636, opened 2025-11-09
Get the entire code base to pass `golangci-lint run`


stgraber:
# Introduction
We've been using `golangci-lint` for a little while now within the Incus project, but it's always been configured to only check the code being modified, purposefully ignoring the pre-existing code so we didn't have to fix hundreds/thousands of issues.

A number of contributors over the past few years have tackled cleaning up failures in various places, but it may be worth making more of a concerted effort in addressing this, leading to us changing our CI to do a full run after we've fixed everything (or added the relevant exceptions).

This would then let us finally tweak our `golangci-lint` config to enable more linters.
Probably not to the same level we've done in IncusOS which has just about every linter enabled, but still increase our coverage a fair bit.

# What can be done
This is an issue that can be easily parallelized with many folks, all you have to do is:

- Comment here to say what package or package(s) you're looking into (e.g. `cmd/**`, `internal/server/network/...`, `cmd/incusd`, ...)
- Run `golangci-lint <package>` to get an idea of what you're dealing with
- Run `golangci-lint --fix <package>` to apply the automatic fixes when possible
- Review those automatic fixes to make sure they make sense, then commit them as their own commit
- Run `golangci-lint <package>` again to get what needs manual attention and look at fixing them or add `#nolint:XYZ` for the few cases where we should ignore the error
- Commit those manual changes as their own commit
- Open a pull request referencing this issue (but not closing it)

RageZBla:
@stgraber please assign it to me 🙇 and let me start with the `cmd/**`

stgraber:
Done!

dector:
I would love to grab `internal/**` package.

I have fixes for it. @stgraber , would you prefer me to publish them as:
- one PR with multiple commits (for sub-packages)
- one PR with single commit
- multiple PRs (for sub-packages)
?

stgraber:
@dector One PR per package with sub-commits in there to separate the automatic fixes (`golangci-lint run --fix`) and then one commit per type of fixes being manually made after that.

That should make it reasonably easy to review for us.

Start with one, we'll make sure we're happy with the way that one is structured, then apply to the rest.

dector:
> One PR per package with sub-commits in there to separate the automatic fixes (`golangci-lint run --fix`) and then one commit per type of fixes being manually made after that.

Great, was just wondering if I should try to avoid ddosing test infra. 😁

RageZBla:
[...]

PR #2655, opened 2025-11-14
incus/util: #2636 fix linter complaints in `internal/util`

dector:
Changes:

  - Use `revive:disable-next-line:var-naming` for package `util`. This name is generic but seems good enough in this context.
  - Renamed variable name to follow Go conventions.
  - Add missing documentation comment for MkdirAllOwner function.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -69,6 +69,8 @@ linters:
           - godot
         text: "Comment should end in a period"
         source: '^// Example:'
+      - path: internal/util/
+        text: "avoid meaningless package names"
     paths:
       - third_party$
       - builtin$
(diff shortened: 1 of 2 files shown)


internal/util/cert.go
```diff
@@ -1,3 +1,4 @@
+//revive:disable-next-line:var-naming
```
stgraber: Assuming this is triggering:
```
avoid meaningless package names
```

Then the way we've handled it elsewhere is with a tweak to `.golangci.yml` with:
```
  exclusions:
    generated: lax
    rules:
      - path: internal/util/
        text: "avoid meaningless package names"
```

This is a bit cleaner as you don't need an override at the top of every file in the package.

stgraber: See https://github.com/lxc/incus-os/blob/main/incus-osd/.golangci.yml

dector: Perfect, will fix it, thanks

dector: Pushed new changes.

PR #2661, opened 2025-11-16
cmd/incus-agent: address golangci-lint issues

RageZBla:
covers `cmd/incus-agent` part of #2636

### DIFF
--- a/cmd/incus-agent/exec.go
+++ b/cmd/incus-agent/exec.go
@@ -202,6 +202,7 @@ func (s *execWs) Do(op *operations.Operation) error {
 	logger.Debug("Waiting for exec websockets to connect")
 	select {
 	case <-s.requiredConnectedCtx.Done():
+		//nolint:revive //whyNoLint: this is intentional, the flow should continue if all websockets are connected
 		break
 	case <-time.After(time.Second * 5):
 		return errors.New("Timed out waiting for websockets to connect")


cmd/incus-agent/dev_incus.go
```diff
@@ -66,6 +66,7 @@ var DevIncusConfigGet = devIncusHandler{"/1.0/config", func(d *Daemon, w http.Re
 		return smartResponse(err)
 	}
 
+	//nolint:revive //whyNoLint: the function is short and the shadowing is fine
```
stgraber: I'd still have made it `conf` or something instead just to avoid the shadowing should the function eventually get copy/pasted elsewhere or evolve to become problematic.

cmd/incus-agent/main_agent.go
```diff
@@ -49,6 +49,7 @@ func (c *cmdAgent) Run(cmd *cobra.Command, args []string) error {
 	// Setup logger.
 	err := logger.InitLogger("", "", c.global.flagLogVerbose, c.global.flagLogDebug, nil)
 	if err != nil {
+		//nolint:revive //whyNoLint: this is equivalent to main
```
stgraber: We should be able to just `return err` and have the caller handle the error.

cmd/incus-agent/main_agent.go
```diff
@@ -186,6 +187,7 @@ func (c *cmdAgent) Run(cmd *cobra.Command, args []string) error {
 	cancelStatusNotifier() // Ensure STOPPED status is written to QEMU status ringbuffer.
 	cancelFunc()
 
+	//nolint:revive //whyNoLint: this is equivalent to main
 	os.Exit(exitStatus)
```
stgraber: Looks like the two options for the exit status here are 0 or 1, so we should just `return nil` on success and `return err` on failure, then let the parent handle the actual exit.

cmd/incus-agent/exec.go
```diff
@@ -158,6 +158,7 @@ func (s *execWs) Connect(op *operations.Operation, r *http.Request, w http.Respo
 			}
 
 			s.connsLock.Lock()
+			//nolint:revive
```
stgraber: Should put a line break before that

cmd/incus-agent/exec.go
```diff
@@ -202,6 +202,7 @@ func (s *execWs) Do(op *operations.Operation) error {
 	logger.Debug("Waiting for exec websockets to connect")
 	select {
 	case <-s.requiredConnectedCtx.Done():
+		//nolint:revive //whyNoLint: this is intentional, the flow should continue if all websockets are connected
```
stgraber, line 205: Line break after this one

stgraber:
Just a few minor tweaks needed.

cmd/incus-agent/dev_incus.go
RageZBla: 8ae0d885c07a4b7d6e1b9054cb11b7a01b6ba9e8

cmd/incus-agent/main_agent.go
RageZBla: 990ffc785281a50ff13c1a933247028cf8405519

cmd/incus-agent/main_agent.go
RageZBla: 990ffc785281a50ff13c1a933247028cf8405519

cmd/incus-agent/exec.go
RageZBla: 92a085656c82d703c86f1f6c42759bf8a6f25dcf

cmd/incus-agent/exec.go
RageZBla: 92a085656c82d703c86f1f6c42759bf8a6f25dcf

RageZBla:
@stgraber thanks for reviewing, I have addressed your feedback :bow:

PR #2671, opened 2025-11-18
Support for the Incus Agent as a Service on Windows

MOBergeron:
### Description

This pull request is to add better support for the Incus agent under Windows as a service with the installation scripts.

### Not so Sure About Those...

I'm not sure about `m.agentCmd.mountHostShares()` at line 48 in `os_windows.go` if it is necessary. There might be things missing but I was not sure it they were necessary such as the notification channels.

### How to Use

Simply add the CD-ROM to the virtual machine with `incus config device add my-windows-vm cdrom disk source=agent:config` and then, run as an administrator the `install.ps1` from the CD-ROM drive within the virtual machine.

### DIFF
--- a/doc/howto/instances_create.md
+++ b/doc/howto/instances_create.md
@@ -206,12 +206,19 @@ The subsequent commands will run the installation script `install.sh` to install
 #### On Windows
 
 For Windows systems, the virtual CD-ROM drive must be used.
-The agent can manually be started by opening a terminal and running (assuming `d:\` is the CD-ROM):
+The agent can be installed as a service by running the `install.ps1` file from the CD-ROM drive (by either executing the file from the explorer or terminal).
+To automatically let the agent update itself, leave the CD-ROM available to the virtual machine
+
+```{note}
+To install: The agent as a service, local administrator privileges are required.
+
+To update: When rebooting the virtual machine, the CD-ROM will be mounted again with the newest version and the service will automatically update itself using the files in the CD-ROM drive.
+```
+
+Otherwise, the agent can manually be started by opening a terminal and running (assuming `d:\` is the CD-ROM):
 
     d:\
     .\incus-agent.exe
-
-To have it persist and run automatically, a system service can be manually defined to start it up.
 
 #### On macOS
 


cmd/incus-agent/main.go
```diff
@@ -9,8 +9,9 @@ import (
 )
 
 type cmdGlobal struct {
-	flagVersion bool
-	flagHelp    bool
+	flagVersion        bool
+	flagHelp           bool
+	flagWindowsService bool
```
stgraber: I'd go with `flagService`

cmd/incus-agent/main.go
```diff
@@ -33,6 +34,7 @@ func main() {
 	app.PersistentFlags().BoolVarP(&globalCmd.flagHelp, "help", "h", false, "Print help")
 	app.PersistentFlags().BoolVarP(&globalCmd.flagLogVerbose, "verbose", "v", false, "Show all information messages")
 	app.PersistentFlags().BoolVarP(&globalCmd.flagLogDebug, "debug", "d", false, "Show all debug messages")
+	app.PersistentFlags().BoolVar(&globalCmd.flagWindowsService, "service", false, "Start as a Windows service")
```
stgraber: Should be flagService and just say `Start as a system service` but put the line under a `if runtime.GOOOS == "windows"` so it's only visible on Windows for now.

cmd/incus-agent/os_darwin.go
```diff
@@ -38,6 +38,10 @@ func parseBytes(b []byte) string {
 	return string(b[:n])
 }
 
+func runService(name string, isDebug bool) error {
+	return errors.New("Can only run a Windows service under Windows.")
```
stgraber: Based on above comment, I'd change the error to just "Not implemented" as the codepath should basically be unreachable given the flag itself won't be present on non-Windows.

cmd/incus-agent/os_linux.go
```diff
@@ -52,6 +52,10 @@ var (
 	osGuestAPISupport      = true
 )
 
+func runService(name string, agentCmd *cmdAgent) error {
+	return errors.New("Can only run a Windows service under Windows.")
```
stgraber: Same as MacOS

cmd/incus-agent/os_windows.go
```diff
@@ -4,25 +4,109 @@ package main
...
+func (m *incusAgentService) Execute(args []string, r <-chan svc.ChangeRequest, changes chan<- svc.Status) (ssec bool, errno uint32) {
+	const cmdsAccepted = svc.AcceptStop | svc.AcceptShutdown
+
+	tick := time.Tick(2 * time.Second)
+
+	changes <- svc.Status{State: svc.StartPending}
+
+	// Mount shares from host.
+	m.agentCmd.mountHostShares()
+
```
stgraber: You can drop that, I'm pretty sure the function is a no-op on Windows anyways.

cmd/incus-agent/os_windows.go
```diff
@@ -4,25 +4,109 @@ package main
...
+		}
+	}
+
+	changes <- svc.Status{State: svc.StopPending}
+
+	return
+}
+
+func runService(name string, agentCmd *cmdAgent) error {
+	var err error
```
stgraber, line 80: Line break here

cmd/incus-agent/os_windows.go
```diff
@@ -4,25 +4,109 @@ package main
...
+	if agentCmd.global.flagLogDebug {
+		elog = debug.New(name)
+	} else {
+		elog, err = eventlog.Open(name)
+		if err != nil {
+			return err
+		}
+	}
+	defer elog.Close()
```
stgraber, line 91: Line break here

cmd/incus-agent/os_windows.go
```diff
@@ -4,25 +4,109 @@ package main
...
+
+	elog.Info(1, fmt.Sprintf("Starting %s service", name))
+	run := svc.Run
+	if agentCmd.global.flagLogDebug {
+		run = debug.Run
+	}
+	err = run(name, &incusAgentService{agentCmd: agentCmd})
```
stgraber, line 99: Line break here

cmd/incus-agent/os_windows.go
```diff
@@ -4,25 +4,109 @@ package main
...
+	if err != nil {
+		elog.Error(1, fmt.Sprintf("%s service failed: %v", name, err))
+		return err
+	}
+	elog.Info(1, fmt.Sprintf("%s service stopped", name))
```
stgraber, line 105: Line break here

internal/server/instance/drivers/driver_qemu.go
```diff
@@ -3272,6 +3272,41 @@ func (d *qemu) generateConfigShare() error {
...
+		// Install script for manual installs.
+		agentFile, err = incusAgentLoader.ReadFile("agent-loader/install-windows.ps1")
+		if err != nil {
+			return err
+		}
+
+		err = os.WriteFile(filepath.Join(configDrivePath, "install.ps1"), agentFile, 0o700)
+		if err != nil {
+			return err
+		}
```
stgraber, line 3309: Having `-windows` in the filenames feels a bit redundant when they're in a chunk of OS-specific code, so can probably shorten their names a bit by removing it.

stgraber (changes requested):
Just a few nits and code style things

stgraber:
Thanks a lot, that will make it a LOT more approachable for Windows users!

Could you add another commit to tweak the documentation?
https://linuxcontainers.org/incus/docs/main/howto/instances_create/#on-windows

That should be in `doc/howto/instances_create.md`

PR #2677, opened 2025-11-19
Workaround Linstor URL parsing issue

stgraber:
Upstream issue is https://github.com/LINBIT/golinstor/issues/19

### DIFF
--- a/test/lint/govulncheck.sh
+++ b/test/lint/govulncheck.sh
@@ -1,9 +1,9 @@
 #!/bin/sh -eu
 
-echo "Checking for vulnerabilities in Go dependencies using govulncheck..."
+echo "Checking for vulnerabilities in Go dependencies using govulncheck... (non-fatal)"
 
 GOVERSION=$(go version | cut -d' ' -f3 | sed "s/go//g")
 cp go.mod go.mod.bak
 sed "s/^go 1.*/go ${GOVERSION}/" -i go.mod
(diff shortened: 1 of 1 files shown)


internal/server/storage/linstor/linstor.go
```diff
@@ -79,8 +81,51 @@ func NewClient(controllerConnection, sslCACert, sslClientCert, sslClientKey stri
 
 	// Setup the Linstor client.
 	httpClient := &http.Client{Transport: httpTransport}
-	controllerUrls := strings.Split(controllerConnection, ",")
-	c, err := linstorClient.NewClient(linstorClient.Controllers(controllerUrls), linstorClient.HTTPClient(httpClient), linstorClient.Log(linstorLogger{}))
+
+	parseConnection := func(connection string) (*url.URL, error) {
```
jarrodu, line 85: Should we add a comment with a bit of context? If the upstream issue is fixed then this block of code might not need to be here.

https://github.com/LINBIT/golinstor/issues/19

stgraber: I'm subscribed to the upstream issue, but their current parsing is very very weird and even if they relax the port restriction to make things work here, I'm not sure I'd want to use the rest of the logic. We're likely better off doing our own parsing here and just telling Linstor exactly what to connect to.

PR #2698, opened 2025-11-25
Add option for disabling PCI firmware

### LINKED ISSUE #2542: Add a way to disable ROM-Bar

Some PCIE device doesn't work with ROM-Bar (or make the boot process very long).
As advertised by the proxmox [wiki](https://pve.proxmox.com/wiki/PCI(e)_Passthrough#_host_device_passthrough) it may be needed to be disabled for it to work.

Add a parameter for rom-bar in the pcie configuration to explicitly enable (which is enabled by default in recent qemu version from what I could glance)  or disable (what I want to do).
Maybe also add a parameter to specify a rom file (like proxmox, libvirt or qemu can do)

Mujib-Ahasan:
This PR is raised to propose some changes by which ROM-bar could be disabled. A new boolean config option is added, which by default value is true. By making this false ROM-bar can be disabled.

Closes #2542


stgraber:
Looking into this one now

Issue #2621, opened 2025-11-05
Introduce recursion=2 for storage volumes and storage buckets


breml:
### What are you currently unable to do

Get storage volume and storage bucket details including snapshot and backup information, similarly to how it is possible for instances with `client.GetInstanceFull`.

### What do you think would need to be added

Introduce `recursion=2` with `...Full` struct for storage volumes and storage buckets.
Add the respective methods to Incus client package.

PR #2705, opened 2025-11-27
Add client.StorageVolumeFull

breml:
@stgraber I looked into #2621 a bit. For the `StorageVolumeFull`, we can take a short cut but only implementing the backwards compatible variant in the client. This causes 4 round trips instead of 1, but I assume, that we will require this implementation anyway down the road, even if we later add a more efficient implementation with `recursion=2`.

Relates to #2621


stgraber:
Yeah, my preference is still to implement the server side logic which really shouldn't be too difficult but if I run out of time, I can go with the client-side only logic, then handle the API extension later.

Issue #2654, opened 2025-11-14
Bandwidth limits on instances attached to OVN network


DarkressX:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

Currently it's not possible to limit the ingress or egress bandwidth of a nic device which is using a ovn network, similar t how it would work on a bridge driver with `limits.ingress`.

### What do you think would need to be added

I think this might be possible to do via OVNs QoS mechanism.

Preliminary tests via ovn-nbctl allowed me to restrict the bandwidth limits of a logical switch port.

By adjusting the ovn_driver and other components such as ovn_nb_actions, the QoS functionality could be implemented.

Since adding ACLs and QoS work slightly similar in OVN, we might should be able to use the ACL implementation as an example.

stgraber:
Ah, would be nice if QoS worked these days. I remember looking at it a few years back and it really didn't provide an safe and efficient way to do this. That is, it was sufficient for traffic prioritization but it didn't have a good way to make traffic slow down without needing to buffer it (which would then consume a bunch of memory).

DarkressX:
OVN, at least nowadays seems to use OpenFlow rules which use meters. These meters will then drop traffic exceeding the meters limit.

stgraber:
Ah, that's good, that should avoid the kind of issues we've seen with some of the tc/qdisc stuff in the past.

DarkressX:
@stgraber Could you please assign me to this?
I'm currently working on it anyways and might be able to have a prototype ready soon :)

stgraber:
Done!

DarkressX:
Unfortunately, while developing I discovered some weird behavior where ovs (on the node where a qos has been applied) looses the connection to its peers after some time.
This seems to be the same behavior as described [here](https://mail.openvswitch.org/pipermail/ovs-discuss/2025-May/053594.html).
I have not found the root cause yet.

Restarting the ovs-vswitchd unit helps, but only temporarily.

stgraber:
Hmm, that sounds like a rather annoying one.

What OVS and kernel are you testing on?

DarkressX:
I'm running Kernel 6.12.57 on Debian 13 in a three node cluster
OVS version: 3.5.0
OVN version: 25.03.0

stgraber:
The OVS guys seem to think it could be a kernel issue of some kind, so I suspect any further debugging on that end should be done on latest stable (6.17.x) and ideally latest OVS too.
[...]

PR #2709, opened 2025-11-27
Feat/ovn bandwidth limit

DarkressX:
This PR resolves https://github.com/lxc/incus/issues/2654
This PR implements the creation of ingress and egress bandwidth limits on an instance level.
These limits are converted to ovn compatible QoS rules.

I decided to make the `limits.ingress` and `limits.egress` config mutually exclusive with `limits.max`, as I think it could be confusing for users, as they would be unsure which limit is actually being set if both configuration options are given.

### DIFF
--- a/internal/server/network/driver_ovn.go
+++ b/internal/server/network/driver_ovn.go
@@ -5135,20 +5135,35 @@ func (n *ovn) InstanceDevicePortStart(opts *OVNInstanceNICSetupOpts, securityACL
 		qosPriority = 100
 	}
 
-	// TODO: implement limits.max config key
-	var rules []networkOVN.OVNQoSRule
-	if opts.DeviceConfig["limits.egress"] != "" {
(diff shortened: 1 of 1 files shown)


stgraber:
Thanks, that looks very good!

It's very late here but I should be able to do some tidying up of the branch and merge that tomorrow before Incus 6.19.

DarkressX:
No worries :)

stgraber:
Looking into this one now

stgraber:
Ah right, you're using an org account so I can't easily do in-place rebase/tweaks.

I'll pull the commits into one of my branches, then do the tweaks there and close this PR once mine is up.

DarkressX:
Ah sorry, I wasn't aware of that limitation. I can rebase it too if thats easier :)

stgraber:
I opened https://github.com/lxc/incus/pull/2711 which has effectively identical code except for the extra API extension commit, rebased content and slightly tweaked commit descriptions.

PR #2720, opened 2025-11-30
Add Default Environment Variables for the Windows Agent

MOBergeron:
Currently, the Windows agent did not have `WINDIR`, `TEMP` and `TMP` environment variables which are used by some cmdlets and more.

That makes the a few cmdlets fail and thus, making it mandatory to add environment variables through the Incus exec command as a parameter such as `incus exec --env WINDIR="C:\\WINDOWS" ...`. However, some of them should be present by default in a Windows environment.

Furthermore, I've updated the current working directory that is usually `C:\WINDOWS\System32\` when running as `nt authority\system`.

### DIFF
--- a/cmd/incus-agent/os_windows.go
+++ b/cmd/incus-agent/os_windows.go
@@ -237,5 +237,14 @@ func osExitStatus(err error) (int, error) {
 }
 
 func osSetEnv(post *api.InstanceExecPost, env map[string]string) {
-	env["PATH"] = "C:\\WINDOWS\\system32;C:\\WINDOWS"
+	// SystemRoot is already set by default
+	env["WINDIR"] = "C:\\WINDOWS"
+	env["TMP"] = "C:\\WINDOWS\\Temp"
+	env["TEMP"] = env["TMP"]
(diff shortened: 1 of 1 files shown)


MOBergeron:
Example before:
```
$ incus exec -T webb -- cmd /c set
COMSPEC=C:\WINDOWS\system32\cmd.exe
PATHEXT=.COM;.EXE;.BAT;.CMD;.VBS;.JS;.WS;.MSC
PROMPT=$P$G
HOME=/root
LANG=C.UTF-8
PATH=C:\WINDOWS\system32;C:\WINDOWS
TERM=tmux-256color
USER=root
SYSTEMROOT=C:\WINDOWS
```

Example after:
```
$ incus exec -T webb -- cmd /c set
COMSPEC=C:\WINDOWS\system32\cmd.exe
PATHEXT=.COM;.EXE;.BAT;.CMD;.VBS;.JS;.WS;.MSC
PROMPT=$P$G
TMP=C:\WINDOWS\Temp
TEMP=C:\WINDOWS\Temp
PATH=C:\WINDOWS\System32;C:\WINDOWS;C:\WINDOWS\System32\WindowsPowerShell\v1.0
TERM=tmux-256color
WINDIR=C:\WINDOWS
SYSTEMROOT=C:\WINDOW
```

PR #2736, opened 2025-12-07
Feature: restrict agent features from the guest

### LINKED ISSUE #2684: Allow restricting agent features from within the guest

We are currently migrating from a legacy virtualization system to Incus and are expecting every VM maintainer to install the Incus Agent.
However some of the people maintaining VMs in our infrastructure are not happy with the fact that you are able to easily get a root shell in the WebUI thanks to the agent.
The current workaround is to remove the agent however that causes CPU, memory and IPs to not be visible which is also not ideal.

We would like to see the possibility to disable the Terminal feature of the agent from inside the VM.

Mujib-Ahasan:
## Description
This PR implements feature-based access control in the Incus agent using a new configuration file (`incus-agent.yml`).
This allows VM maintainers to selectively disable agent capabilities.

close #2684

### DIFF
--- a/doc/explanation/bpf-tokens.md
+++ b/doc/explanation/bpf-tokens.md
@@ -14,24 +14,24 @@ The permissible values for these options depend on the kernel version and can be
 (`include/uapi/linux/bpf.h` in the kernel tree, `/usr/include/linux/bpf.h` on most distributions if you have the kernel
 sources installed):
 
- Key                               | Kernel `enum`      | Remove prefix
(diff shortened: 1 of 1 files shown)


Mujib-Ahasan:
Hello @stgraber, I made few changes, but did not get for files, mount and info. Also let me know about the implementation.

stgraber:
Made a few changes/refactoring, added documentation and added the remaining codepaths.
I ran through a test (on Linux) and everything looks good.

PR #2747, opened 2025-12-11
Improve VM state handling

stgraber:


internal/server/instance/drivers/driver_qemu.go
```diff
@@ -366,6 +366,12 @@ func (d *qemu) qmpConnect() (*qmp.Monitor, error) {
 // Callers should check that the instance is running (and therefore mounted) before calling this function,
 // otherwise the qmp.Connect call will fail to use the monitor socket file.
 func (d *qemu) getAgentClient() (*http.Client, error) {
+	// Check that the VM is in a state where the agent may be reachable.
+	status := d.statusCode()
+	if !d.isRunningStatusCode(status) || status == api.Frozen {
```
candlerb, line 371: Does this test mean that the comment above the function ("Callers should check that the instance is running ... before calling this function") is no longer relevant?

stgraber: Yeah, the callers could get away with not performing the check now.

stgraber: I think we were requiring the check be done externally as a way to save ourselves the likely double lookup of that information, but given we're now doing something less usual (checking for Frozen), we need the raw value which the callers may not even have (if they just called IsRunning).

PR #2753, opened 2025-12-12
shared/tls: implement Happy Eyeballs (RFC 8305) in RFC3493Dialer

johnae:
The RFC3493Dialer function, before this, tried addresses sequentially with a 10-second timeout per address. This causes problems in dual-stack environments where one address family is unreachable:

- If DNS returns IPv4 addresses first but IPv4 is unreachable (e.g., in IPv6-only environments), the dialer would spend 10+ seconds timing out on each IPv4 address before trying IPv6.
- This exceeded typical HTTP client timeouts, causing connection failures even when IPv6 connectivity was available.
  For example, [cluster-api-provider-incus](https://github.com/lxc/cluster-api-provider-incus) would end up with `Context Deadline Exceeded` since it seems to have a 10-second timeout when fetching the list of images.

This change basically implements Happy Eyeballs (RFC 8305):

1. Sort addresses with IPv6 first (RFC 8305 recommends preferring IPv6)
2. Start connection attempts with a 250ms staggered delay (per RFC 8305)
3. Return the first successful connection immediately
4. Finally clean up unused connections

### DIFF
--- a/shared/tls/tls.go
+++ b/shared/tls/tls.go
@@ -18,9 +18,39 @@ import (
 // connectErrorPrefix used as prefix to error returned from RFC3493Dialer.
 const connectErrorPrefix = "Unable to connect to"
 
+// happyEyeballsDelay is the delay between starting connection attempts per RFC 8305.
+const happyEyeballsDelay = 250 * time.Millisecond
+
+// happyEyeballsTimeout is the overall timeout for the Happy Eyeballs algorithm.
+const happyEyeballsTimeout = 30 * time.Second
+
+// sortAddressesByFamily sorts addresses with IPv6 first (per RFC 8305 Happy Eyeballs).
+// Within each family, the original order is preserved.
+func sortAddressesByFamily(addrs []string) []string {
+	var ipv6Addrs, ipv4Addrs []string
+	for _, addr := range addrs {
+		ip := net.ParseIP(addr)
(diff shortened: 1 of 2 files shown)


stgraber:
Looks like your commit is missing the `Signed-off-line` that we need before we can review it.
Also static analysis is apparently failing, so a few more tweaks needed there before we can review this.

In general, I'm fine with happy eyeballs, though I remember us looking into this a while back and then not pursuing it so I'll want to refresh my memory for whether there was a strong technical reason to avoid it.

johnae:
Thanks @stgraber - I've added the signed-off-by line now as requested.

shared/tls/tls.go
```diff
@@ -31,26 +61,105 @@ func RFC3493Dialer(_ context.Context, network string, address string) (net.Conn,
...
+
+	for pendingDials > 0 {
+		select {
+		case result := <-results:
+			pendingDials--
+			if result.err != nil {
+				errs = append(errs, result.err)
+			} else {
+				connections = append(connections, result.conn)
+			}
```
stgraber, line 131: Static analysis seems to want a line break here.

johnae: Should be resolved now.

stgraber:
Thanks!

PR #2761, opened 2025-12-15
generate-database: Allow overwriting the target column for a join

breml:


cmd/generate-database/db/mapping.go
```diff
@@ -448,7 +448,12 @@ func (f *Field) JoinClause(mapping *Mapping, table string) (string, error) {
 		return "", fmt.Errorf("'joinon' tag of field %q of struct %q must be of form '<table>.<column>'", f.Name, mapping.Name)
 	}
 
-	return fmt.Sprintf(joinTemplate, joinTable, joinOn, joinTable), nil
+	joinTo := "id"
+	if f.Config.Get("jointo") != "" {
+		joinTo = f.Config.Get("jointo")
+	}
```
masnax, line 454: To keep this consistent with `joinon`, I think the structure should be `jointo=table_name.column` rather than just `column`.

cmd/generate-database/db/mapping.go
masnax, line 534: If you're using Insert/Update or Delete, this might cause some issues. Currently the way these statements work with joins is that the existing `ID` SQL statement for the entity type associated with the field of the current struct is appended to the new statement. 

You may need to transform this value (the original `joinStmt` returned from `ParseStmt` based on the variable name from `stmtCodeVar`) to have the correct selected column and condition, as well as update both sides of the `new` argument to `strings.ReplaceAll` for cases where the the table in the join clause has a composite foreign key which happens to also be a foreign key on the current table.

cmd/generate-database/db/mapping.go
breml: I am not sure about this one. I can see the argument for consistency, but it also introduces redundancy, which might even cause an invalid combination in my opinion.

My proposal would make the following work:

```go
type Child struct {
  ID int
  ParentForeignKey string
  ParentValue string `leftjoin=parents.value&joinon=clild.parent_foreign_key&jointo=key`
}

type Parent struct {
  ID int
  Key string
  Value string
}
```

which then would generate:

```sql
SELECT child.id, child.parent_foreign_key, parent.key, parent.value
FROM child
LEFT JOIN parent ON child.parent_foreign_key = parent.key
```

If `jointo` now contains the table as well, it would be possible to define `db:"leftjoin=parents.value&joinon=clild.parent_foreign_key&jointo=notparent.key"`, which would generate the following SQL:

```sql
SELECT child.id, child.parent_foreign_key, parent.key, parent.value
FROM child
LEFT JOIN parent ON child.parent_foreign_key = notparent.key
```

My argument is basically, that the target (here parent) table is already defined by the `join` or `leftjoin` directive and therefore I see little value in repeating it again in `jointo`.
What would be possible is to limit the use of the value provided in `join` / `leftjoin` to just the column selection part, if `jointo` is present, which would then produce:

```sql
SELECT child.id, child.parent_foreign_key, parent.key, parent.value
FROM child
LEFT JOIN notparent ON child.parent_foreign_key = notparent.key
```

(but this is still kind of wired)

One Option that I can see is rename `jointo` to `jointocolumn` to make it more explicit, that here only the column name is expected.

cmd/generate-database/db/mapping.go
breml: The fields I care about are purely read time, so this acts more like a view. I think in my case, I likely want to combine the fields with `omit=create,update`. I wonder, if it would be enough to also document this edge case in the readme, basically mention, that `jointo` is not supported for data manipulation queries (create, update, delete).

breml: With the current branch, it looks like the following definition does produce what I need:

```go
package dbtest

//go:generate go run github.com/lxc/incus/v6/cmd/generate-database db mapper generate -b mapper_boilerplate.go -p "github.com/lxc/incus/v6/0/dbtest"
//go:generate gofmt -s -w .
//go:generate go run golang.org/x/tools/cmd/goimports -w .

// Code generation directives.
//
//generate-database:mapper target dbtest.mapper.go
//generate-database:mapper reset
//
//generate-database:mapper stmt -e child objects
//generate-database:mapper stmt -e child objects-by-ID
//generate-database:mapper stmt -e child names
//generate-database:mapper stmt -e child id
//generate-database:mapper stmt -e child create
//generate-database:mapper stmt -e child update
//generate-database:mapper stmt -e child rename
//generate-database:mapper stmt -e child delete-by-ID
//
//generate-database:mapper method -e child ID
//generate-database:mapper method -e child Exists
//generate-database:mapper method -e child GetOne
//generate-database:mapper method -e child GetMany
//generate-database:mapper method -e child GetNames
//generate-database:mapper method -e child Create
//generate-database:mapper method -e child Update
//generate-database:mapper method -e child DeleteOne-by-ID

type Child struct {
	ID               int
	Name             string `db:"primary=yes"`
	ParentForeignKey string
	ParentValue      string `db:"leftjoin=parents.value&joinon=clilds.parent_foreign_key&jointo=key&omit=create,update"`
}

type Parent struct {
	ID    int
	Name  string `db:"primary=yes"`
	Key   string
	Value string
}

type ChildFilter struct {
	ID   *int
	Name *string
}
```

breml: I have updated the description in the README again to emphasize, that `jointo` is intended to be used in conjunction with `joinon` as well as `omit=create,update`.

PR #2763, opened 2025-12-15
incus/io: #2636 fix linter complaints in `internal/io`

dector:
Changes:

  - Added missing docstrings.
  - Renamed `min` function parameter to `minimum`.


stgraber:
Looks good, thanks!

PR #2764, opened 2025-12-15
incus/util: #2636 fix linter complaints in `internal/filter`

dector:
Changes:

  - Suppressed few casts in-line.
  - Used early return.

### DIFF
--- a/internal/filter/match.go
+++ b/internal/filter/match.go
@@ -88,7 +88,7 @@ func DefaultParseBool(c Clause) (bool, error) {
 // DefaultParseRegexp converts the value of the clause to regexp.
 func DefaultParseRegexp(c Clause) (*regexp.Regexp, error) {
 	regexpValue := c.Value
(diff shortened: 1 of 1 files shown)


internal/filter/match.go
```diff
@@ -137,11 +137,10 @@ func (s ClauseSet) match(c Clause, objValue any) (bool, error) {
 	case reflect.Bool:
 		valueBool, err = s.ParseBool(c)
 	case reflect.Slice:
-		if reflect.TypeOf(objValue).Elem().Kind() == reflect.String {
-			valueSlice, err = s.ParseStringSlice(c)
-		} else {
+		if reflect.TypeOf(objValue).Elem().Kind() != reflect.String {
 			return false, fmt.Errorf("Invalid slice type %q for field %q", reflect.TypeOf(objValue).Elem().Kind(), c.Field)
 		}
+		valueSlice, err = s.ParseStringSlice(c)
```
stgraber, line 144: There should be a line break before this line

dector: Done

PR #2773, opened 2025-12-19
OpenFGA: Documentation now explains that the api_token is needed if you want to activate OpenFGA.

tibeer:
Signed-off-by: Tim Beermann <tibeer@berryit.de>

### DIFF
--- a/doc/config_options.txt
+++ b/doc/config_options.txt
@@ -5279,7 +5279,7 @@ This value is required by some providers.
 :scope: "global"
 :shortdesc: "API token of the OpenFGA server"
 :type: "string"
-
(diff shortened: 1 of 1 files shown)


stgraber:
You're modified auto-generated code, so that's not going to work too well ;)

Anyway, I think it's better covered in the actual documentation rather than in the list of config keys as the interaction between config keys can get too complex to be readable in the long run (in the case of OpenFGA, once we introduce OAUTH support).

Instead, I've staged this change in my branch of fixes ahead of 6.20: https://github.com/stgraber/incus/commit/e6b52d69da5874df060d1f49c1c7ca14968b4736

PR #2778, opened 2025-12-20
Add WireGuard network support

xlmnxp:
- Introduced a new network type: WireGuard.
- Updated network type definitions and handling in various components.
- Implemented the `nicWireguard` device type for managing WireGuard interfaces.
- Added validation and configuration logic for WireGuard networks.
- Enhanced network loading and management to include WireGuard functionality.

This update allows users to create and manage WireGuard networks within the system, expanding the networking capabilities significantly.

### DIFF
--- a/internal/server/network/driver_wireguard.go
+++ b/internal/server/network/driver_wireguard.go
@@ -195,7 +195,8 @@ func (n *wireguard) Validate(config map[string]string, clientType request.Client
 					// Validate each CIDR in the comma-separated list
 					ips := util.SplitNTrimSpace(value, ",", -1, true)
 					for _, ipStr := range ips {
-						if err := validate.IsNetworkAddressCIDR(ipStr); err != nil {
+						err := validate.IsNetworkAddressCIDR(ipStr)
+						if err != nil {
 							return fmt.Errorf("Peer %q allowed_ips contains invalid CIDR %q: %w", peerName, ipStr, err)
 						}
 					}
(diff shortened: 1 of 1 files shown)


xlmnxp:
@stgraber Hello, I currently facing issues with `Tests / System` checks because they fail when try to download openfga from github, I don't think it related to my contribution

bensmrs:
@xlmnxp yeah that’s a known problem, unrelated to your contribution

stgraber:
I must say I'm pretty confused about what the goal is behind this network and NIC type.
I think we'd have pretty strongly benefited from having an open issue or forum thread to discuss the actual need before jumping into implementation here :)

Wireguard is a layer 3 VPN technology, Incus networks are layer 2, so the two aren't likely to mix all that well.

The usual approach would be for each server to have its own network with its own subnet, then configure Wireguard on the host (which is the router for the network) to then route those subnets between servers over the tunnel.


Could you describe your needs so we can look into the best way to meet them?

xlmnxp:
> I must say I'm pretty confused about what the goal is behind this network and NIC type. I think we'd have pretty strongly benefited from having an open issue or forum thread to discuss the actual need before jumping into implementation here :)
> 
> Wireguard is a layer 3 VPN technology, Incus networks are layer 2, so the two aren't likely to mix all that well.
> 
> The usual approach would be for each server to have its own network with its own subnet, then configure Wireguard on the host (which is the router for the network) to then route those subnets between servers over the tunnel.
> 
> Could you describe your needs so we can look into the best way to meet them?

Hello, my main issue was how to announce/distrobute the subnet to incus instances because the routed didn't do it for me, also needed a way to manage WireGuard for Incus and connect multiple cluster nodes together (as alternative for OVN)

PR #2779, opened 2025-12-21
doc/openfga: Improve required config keys

tibeer:
Signed-off-by: Tim Beermann <tibeer@berryit.de>


tibeer:
Stumled upon this while trying to use Incus with OpenFGA from a fresh installation of both services. In fact, Incus does not generate the store for you, but only the authorization-model.

PR #2784, opened 2025-12-22
incus-user: Don't reset setup if user has access

### LINKED ISSUE #2524: Limit reset logic in incus-user to support more advanced environments

I'm working on building an app deployment utility on top of Incus.  We make heavy use of user constrained projects where the user represents a "stack" of applications that are getting deployed.  Similar to how a docker compose file manages a group of apps that work together. 

Currently, this results in a lot of projects on the system named like "user-1234" which is a real hit to discoverability/usability as there is a constant need to mentally map the project name to the stack.  

Instead, we'd like to be able to set or change the project name arbitrarily for a user constrained project (like any other project) and request that Incus changes to using config or metadata to find a user's project instead of the project's name. 

When Incus needs to create or find a user constrained project, first look for a config or metadata entry on projects.  

For example, add a config entry on project's named `restricted.constrained-uid` that gets set to the system user's uid when the project is created for that user.

For backwards compatibility, if no project is found using the above method, then fall back to the current method of looking for a matching project name.

FWIW, we pre-create the project for our stack users so we'd set the project name differently from the beginning and set the config value to that user's UID so Incus knows how to link it when the user connects to Incus.

MarcosDaNight:
Fixes #2524

As discussed in the issue, this prevents incus-user from resetting a user's configuration 
if they already have a valid certificate and access to at least one project, 
regardless of the project name.

### DIFF
--- a/cmd/incus-user/server.go
+++ b/cmd/incus-user/server.go
@@ -200,6 +200,17 @@ func serverSetupUser(uid uint32) error {
 		return fmt.Errorf("Unable to connect to the daemon: %w", err)
 	}
 
+	// Don't reset setup if the user already has a valid certificate and project access.
+	checkCert, err := localtls.ReadCert(filepath.Join(userPath, "client.crt"))
+	if err == nil {
+		// Check if the certificate exists in the trust store.
+		checkFingerprint := localtls.CertFingerprint(checkCert)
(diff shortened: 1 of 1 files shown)


stgraber:
Thanks!

Issue #2254, opened 2025-07-04
Snapshot desctiption


LOTRg:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

Hello All , 
First of all ... thank you so much for your hard work on such a quality, well-thought project which is now my default container platform specially that I don't like docker and the likes.

I have a feature request to please add snapshot description field, it would be nice to know what was this snapshot before or after, or the state or other interesting information that can be entered and later be viewed with snapshot list or something, much like other software like VMWare products, KVM or even incus images where description field is available ... this will greatly help with workflow knowing exactly what each snapshot represent.

Once again, thank you for a quality software and the hard work you do .

LOTRg:
Any feedback on this or if it's planned ?

stgraber:
We actually have a description field in the database and API, so should just be a matter of correctly exposing it.

stgraber:
Ah, hmm, not quite, the current description field we have is meant to snapshot the instance's own description and then restore it, so we shouldn't use that.

I think we should do a bit of re-shuffling so that `instances_snapshots` holds data about the snapshot themselves and then use a `instances_snapshots_properties` or something along those line to hold the instance properties, avoiding mixing the two.

At the API level, we'd then probably want to put the snapshot-specific stuff into a separate struct.

So anyway, that will be a fair amount of work, also because we'd want to clean up this for both instances and volumes, but it's certainly worth doing.

LOTRg:
Very much appreciated @stgraber and thank you for the great work ! looking forward to when this get implemented .

heldergomes:
I'd like to take this one!

@stgraber may I ask for help about which components I should digging deep to understand how to make those changes

heldergomes:
I spent some time understanding the snapshot code in Incus and came to the following conclusion regarding this change:

- Need to add new version on update.go to add instances_snapshots_properties, storage_volumes_snapshots_properties and migrate description field to the new table for past snapshots.
- Add description flag on snapshot creation (CLI and API).
- Add new struct "properties" on snapshot retrieve.
- Parse properties model on snapshot creation and restore.

I'm still studying snapshot's feature to undestand if scheduler might have any side effect.

I was thinking to add only description on properties' table for now, is there any other field that should be moved to there?

stgraber:
I think the first step should be to create `instances_snapshots_properties` and `storage_volumes_snapshots_properties` and as part of that, move just about all existing fields from the main table over to that.

[...]

PR #2786, opened 2025-12-23
Add description on snapshot creation

heldergomes:


heldergomes:
because this MR is quite long, I will try to explain a little bit about some change decisions and few concerns!

**storage_volume_snapshot**
Despite the requirement is quite simple, the change is painful because I had to change many piece of code, therefore I decided to change only instance snapshot on this PR, but I could open a new pull request changing volume snapshot afterward.

**shared/api/instance_snapshot.go**
I'd expected to refactor a little bit the InstanceSnapshot struct, snapshot data such as (expiry_at and description) should be on root struct and source instance data like (instance description) should be on nested object like InstanceProperties IMHO, this would design better the model to handle better with more instance properties in the future, however this change would break API, threfore I just add SnapshotDescription field on the model.

**instance_args**
Handle this struct was quite challenge mostly because incus handle snapshots and instance in the same way with instance_args, so although qemu/lxc driver and shared/api/instance_snapshot I ended up handle snapshot fields such as description and expiry on root struct, I took decision to create snapshot_args as nested object of instance_args, in order to make easier add new fields on this highly used struct.

I'd rather use snapshot as a nested struct of instance on every place like driver, api, db, etc, but this change is painful, so it's what it is!

**instance_snapshot_property**
I tried do use the term instance_snapshot_properties, however generate_database script only knew how to create mapper like list if tables name was singular, tables with suffix ies is considered plural on the script.

I think that's all, let me know if have any question or improvements.

heldergomes:
I will check the pipeline errors and fix it!

heldergomes:
This PR implements https://github.com/lxc/incus/issues/2254

PR #2819, opened 2026-01-13
Fix adding disk with a device name longer than 31 bytes

presztak:
Closes: [#2799](https://github.com/lxc/incus/issues/2799)

### DIFF
--- a/internal/server/instance/drivers/util_test.go
+++ b/internal/server/instance/drivers/util_test.go
@@ -64,3 +64,21 @@ func TestExtractTrailingNumber(t *testing.T) {
 		t.Errorf("unexpected error message: got %q, want %q", err.Error(), expectedErr)
 	}
 }
+
+// Test hashName.
+func TestHashName(t *testing.T) {
+	value := hashName("test", 5)
+	assert.Equal(t, "test", value)
+
+	value = hashName("test1", 5)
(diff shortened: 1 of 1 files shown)


internal/server/instance/drivers/util.go
```diff
@@ -354,3 +356,20 @@ func filterAndSortQcow2Blockdevs(names []string, rootDevName string) []string {
...
 	return result
 }
+
+// hashName returns a hash of the name if it exceeds the given length limit.
+// Otherwise, it returns the original name unchanged.
+func hashName(name string, maxLength int) string {
+	if len(name) > maxLength {
+		// If the name is too long, hash it as SHA-256 (32 bytes).
+		// Then encode the SHA-256 binary hash as Base64 Raw URL format and trim down to 25 chars.
+		// Raw URL avoids the use of "+" character and the padding "=" character which QEMU doesn't allow.
```
stgraber: Should be updated to remove the hardcoded 25 here and just mention to "maxLength"

stgraber:
Looks good, just a comment that could do with an update.

internal/server/instance/drivers/util.go
presztak: Fixed.

Issue #2780, opened 2025-12-21
Add both `attached` and `connected` properties to NIC devices


tokred:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

Context: I am migrating to an Incus- and OVN-based cluster where the central router/firewall will be running in a container on said cluster. To connect existing network-wide VLANs, I am leveraging (unmanaged) bridges with tagged interfaces configured on the cluster nodes.

To prepare and test migration steps, I realized it would be beneficial to allow adding and configuring `nic` devices to containers/VMs but keep them in a **_disconnected state_** from the actual network. This would allow the device to appear in the container/VM but not (yet) being connected to the physical network.

Proxmox has a [link_down](https://pve.proxmox.com/pve-docs/pve-admin-guide.html#pct_container_network) parameter for NICs which disables connectivity on hypervisor-level, and I think the same functionality would be great for Incus too. This would also be benefical to temporarily disconnect a NIC without having to delete any container config, avoiding workaround like moving the NIC to a placeholder bridge or similar.

### What do you think would need to be added

Adding a new boolean config parameter like `connected` to `nic` devices, and based on this, connect or disconnect the NIC to the configured parent/network.

bensmrs:
The approach we’ve taken with other device types is an `attached` boolean, which effectively removes the device from the guest. That said, nothing really prevents you to configure a non-existing interface. Wouldn’t that be enough for you?
The `connected` bit could probably be implemented with QMP’s `set_link` so that could be very easy.
I can take this one for the next milestone.

tokred:
@bensmrs, I was not aware of the `attached` parameter (for disks), thanks for the reference! As you mentioned, it does remove the device from the guest though, so a slightly different approach.

What I would love to see for NICs is a "link down" like a hypervisor-controlled "cable plugged" property, because that would allow to a) prepare the entire config on Incus-side, and b) prepare the guest-side network configuration because the NIC device is already available in the guest. But not yet connected to any actual network, so avoiding interference. Further, it would allow to temporarily disconnect a NIC from its network for e.g. troubleshooting, without having to remove the entire device.

Let me experiment with your idea of configuring a non-existing interface, as a workaround! Thanks

tokred:
Regarding non-existing interfaces, at least bridged NICs perform strict checks:

```
 » incus config device add fw1 eth1 nic nictype=bridged
[...]

PR #2847, opened 2026-01-22
Add `attached` and `connected` keys to NIC devices

bensmrs:
Closes: #2780

### DIFF
--- a/test/suites/container_devices_nic_bridged.sh
+++ b/test/suites/container_devices_nic_bridged.sh
@@ -709,6 +709,30 @@ test_container_devices_nic_bridged() {
     ! incus start foo2 || false
     incus delete -f foo foo2
 
+    # Test attached and connected keys.
+    incus launch testimage foo
+    incus config device add foo eth0 nic nictype=bridged name=eth0 parent=${brName}
+    incus exec foo ip link set eth0 up
+    [ "$(incus file pull foo/sys/class/net/eth0/operstate -)" = "up" ]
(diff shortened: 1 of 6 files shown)


bensmrs:
I’ll write a few tests in ~3h tops

stgraber:
Cool, static analysis is also looking a bit unhappy :)

bensmrs:
Haven’t done any Go for a while, my reflexes are a bit rusty!

bensmrs:
I don’t have an OVN testbed and the test suite doesn’t test OVN, so it would be nice if you have a free testing instance with OVN to test that OVN works like the others when setting `acceleration=none`.

Shaderbug:
Hi @bensmrs and @stgraber,

did the SR-IOV test changes introduced with this PR run successfully for you? They seem to be failing on my machine and I want to exclude this being a me-problem 😅

Thank you very much :)

stgraber:
Our CI doesn't have SR-IOV hardware, so INCUS_NIC_SRIOV_PARENT isn't set there.
We do have daily Jenkins SR-IOV tests and my desktop machine also supports SR-IOV but neither of them commonly et INCUS_NIC_SRIOV_PARENT either.

So it may very well be that the test in question is broken.
Please file a new issue for tracking and include the output you're getting.
We made a few changes that may have broken the test. It can also be an environmental issue depending on the card being used, but we'll figure it out.

Shaderbug:
Done, I've opened #2881 :)

PR #2848, opened 2026-01-22
Security fixes ahead of Incus 6.21

stgraber:
This addresses two security issues that were responsibly disclosed to the Incus security team by Rory McNamara of snyk.io.

They cover:
 - GHSA-x6jc-phwx-hp32 (CVE-2026-23953)
 - GHSA-7f67-crqm-jgh7 (CVE-2026-23954)

### DIFF
--- a/internal/server/instance/drivers/driver_lxc.go
+++ b/internal/server/instance/drivers/driver_lxc.go
@@ -7199,6 +7199,32 @@ func (d *lxc) templateApplyNow(trigger instance.TemplateTrigger) error {
 		containerMeta["privileged"] = "false"
 	}
 
+	// Setup security check.
+	rootfsPath, err := os.OpenFile(d.RootfsPath(), unix.O_PATH, 0)
+	if err != nil {
+		return fmt.Errorf("Failed to open instance rootfs path: %w", err)
+	}
+
+	defer func() { _ = rootfsPath.Close() }()
+
+	checkBeneath := func(targetPath string) error {
+		fd, err := unix.Openat2(int(rootfsPath.Fd()), targetPath, &unix.OpenHow{
+			Flags:   unix.O_PATH | unix.O_CLOEXEC,
+			Resolve: unix.RESOLVE_BENEATH | unix.RESOLVE_NO_MAGICLINKS,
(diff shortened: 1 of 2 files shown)


internal/server/instance/drivers/driver_lxc.go
```diff
@@ -7199,6 +7199,33 @@ func (d *lxc) templateApplyNow(trigger instance.TemplateTrigger) error {
...
+		if err != nil {
+			if errors.Is(err, unix.EXDEV) {
+				return errors.New("Template is attempting access to path outside of container")
+			}
+
+			return nil
+		}
+
+		f := os.NewFile(uintptr(fd), targetPath)
+		_ = f.Close()
```
cyphar: nit: I would just use `unix.Close()` here?

internal/server/instance/drivers/driver_lxc.go
```diff
@@ -7211,8 +7238,38 @@ func (d *lxc) templateApplyNow(trigger instance.TemplateTrigger) error {
 				return nil
 			}
 
+			// Perform some security checks.
+			relPath := strings.TrimLeft(tplPath, "/")
+
+			err = checkBeneath(relPath)
```
cyphar, line 7243: I guess you aren't concerned with races here?

cyphar: Ah, I just saw the comment about this in the security advisory. This is not an issue.

internal/server/instance/drivers/driver_lxc.go
stgraber: Indeed, let me actually change that quickly, will save us an operation.

internal/server/instance/drivers/driver_lxc.go
stgraber: Right, this is run prior to anything being spawned in the container so the user shouldn't have any exposure to those files at this point in time, avoiding any risk of TOCTOU.

The logic would have been a lot less easy to read and self-contained if we had been using openat2 for all the actual file operations so this seemed like a good option given the situation.

PR #2879, opened 2026-01-30
feature: Disk only restore

### LINKED ISSUE #2868: Snapshot restore without config and devices

When restoring a snapshot, the instances config and devices will also be restored to the time, when the snapshot was created. This is problematic, as some devices might no longer exist at the time of the snapshot restoration, which causes the snapshot restore to fail, effectively rendering the snapshot useless.

Since the snapshot restore is already a put request against an instance, passing the device and config keys here is already in the api request. For now the snapshot just ignores them and gets the saved config and devices from the snapshot itself.
We would have to implement the ability to overwrite these keys, when they are passed explicitly in the put request.

This should be mutually exclusive with the `stateful` option, as different ram sizes could potentially cause problems.

DarkressX:
This PR implements the option to trigger a diskOnly restore.
As the name of this option suggest, it allows to restore only the instances disk from a given snapshot.
This option is mutually exclusive with the stateful option.

closes #2868

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: LXD\n"
 "Report-Msgid-Bugs-To: lxc-devel@lists.linuxcontainers.org\n"
-"POT-Creation-Date: 2026-02-04 16:37+0100\n"
+"POT-Creation-Date: 2026-02-08 11:49+0100\n"
 "PO-Revision-Date: 2025-12-27 03:16+0000\n"
 "Last-Translator: Dklfajsjfi49wefklsf32 "
 "<nlincus@users.noreply.hosted.weblate.org>\n"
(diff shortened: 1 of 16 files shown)


DarkressX:
@stgraber Please let me know if this is ok as it is and if you want me to squash some commits. I wasn't sure how granular you want the commit history to be :)

DarkressX:
I have to convert it to a draft again. I noticed that now the `volatile.uuid.generation` is no longer persisted.

DarkressX:
Should be fixed now. Sorry about the back and forth

DarkressX:
rebasing

stgraber:
Sorry for the delay, I've been slowly working on the backlog since coming back from FOSDEM, this one should be next in line now :)

DarkressX:
No worries :)
I'll quickly rebase.

PR #2883, opened 2026-02-01
Improvements to incus-migrate

stgraber:
This adds the ability to run the tool unprivileged for operations that don't involve filesystems and it also allows for the use of a URL when passing the path to an image.


stgraber:
Also had to add a couple of websites to the skip list of the link checker as they're running anti-AI measures which block the box.

hallyn (approved):

PR #2885, opened 2026-02-02
MarkdownLint fixes, fix nonstandard {Important}

AJRepo:
If you are viewing the online documentation at https://github.com/lxc/incus/blob/main/doc/backup.md#full-backup the display of the {Important} tag in a browser cuts off the text. This is because GitHub does not support the {important} directive and so renders that text as a code block. That fenced‑directive syntax works for MyST/Quarto/MkDocs, but not viewing docs on Github. 

This PR makes the same formatting using cross-compatible formatting.  Since I was addressing that issue I decided to run mdl (markdown linter) and clean up the rest of the .md file too. 

This PR includes both the Github compatibility fix and the markdown linting fixes.

### DIFF
--- a/doc/backup.md
+++ b/doc/backup.md
@@ -1,14 +1,19 @@
 (backups)=
+
 # How to back up an Incus server
 
 In a production setup, you should always back up the contents of your Incus server.
 
-The Incus server contains a variety of different entities, and when choosing your backup strategy, you must decide which of these entities you want to back up and how frequently you want to save them.
+The Incus server contains a variety of different entities, and when choosing
+your backup strategy, you must decide which of these entities you want to
+ back up and how frequently you want to save them.
 
 ## What to back up
 
-The various contents of your Incus server are located on your file system and, in addition, recorded in the {ref}`Incus database <database>`.
-Therefore, only backing up the database or only backing up the files on disk does not give you a full functional backup.
+The various contents of your Incus server are located on your file system and, in
+addition, recorded in the {ref}`Incus database <database>`.
+Therefore, only backing up the database or only backing up the files on disk
+does not give you a full functional backup.
 
 Your Incus server contains the following entities:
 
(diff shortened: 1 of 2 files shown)


AJRepo:
Odd: The test on github testing markdown gives an incorrect error: 

```
Run make doc-lint
  make doc-lint
  shell: /usr/bin/bash -e {0}
  env:
    GOTOOLCHAIN: local
doc/.sphinx/.markdownlint/doc-lint.sh
Failed!
.tmp/doc/backup.md:3: MD012 Multiple consecutive blank lines
.tmp/doc/backup.md:118: MD012 Multiple consecutive blank lines
make: *** [Makefile:183: doc-lint] Error 1
Error: Process completed with exit code 2.
```

There are no multiple consecutive blank lines in backup.md

AJRepo:
Oh - I get it. The issue is that `sphinx/.markdownlint/doc-lint.sh` has a sed command to add a blank line and mdl is checking the modified line, not the original one. I'll have a fix in a bit

Edit:  The key issue is that the sed command adds the blank line no matter what and saves that and checks as a tmp file. The fix is to modify that sed command to only add a blank line if one does not exist already. This done in the following commit: [#f623f2fa](https://github.com/lxc/incus/pull/2885/commits/f623f2fa397f09a9d98bf72139cc6abf0d8489af)

doc/backup.md
```diff
@@ -19,87 +24,114 @@ Your Incus server contains the following entities:
...
-That means that whether the directory also contains the data of your instances depends on the storage drivers that you use.
+That means that whether the directory also contains the data of your
+instances depends on the storage drivers that you use.
 
-```{important}
-If your Incus server uses any external storage (for example, LVM volume groups, ZFS zpools, or any other resource that isn't directly self-contained to Incus), you must back this up separately.
-```
+> **If your Incus server uses any external storage
+> (for example, LVM volume groups, ZFS zpools, or any other resource that isn't
+> directly self-contained to Incus), you must back this up separately.**
```
stgraber: Does that get rendered the same by Sphinx?

doc/backup.md
```diff
@@ -19,87 +24,114 @@ Your Incus server contains the following entities:
...
 
+```
     incus admin sql local .dump > <output_file>
+```
 
 Use the following command to dump the content of the global database to a file:
 
+```
     incus admin sql global .dump > <output_file>
+```
```
stgraber: This looks good, for code blocks, I'd either expect the three backticks syntax OR the 4 space indent, here we've now got both?

doc/backup.md
AJRepo: No, It does not display the same.  The Sphinx/MyST stylesheet puts an image and a  box around the text block.

Github has  [Callouts](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax) like this: 

> [!IMPORTANT]
> **If your Incus server uses any external storage
> (for example, LVM volume groups, ZFS zpools, or any other resource 
> that isn't directly self-contained to Incus), you must back this up separately.**

Unfortunately the markdown linter (mdl) complained as 
`No link definition for link ID '!important' ` and I'd guess the Sphinx wouldn't read it in. 

I tried different language types to have similar github and  Sphinx/MyST displaying like

````
```md
{important}
**If your Incus server uses any external storage
(for example, LVM volume groups, ZFS zpools, or any other resource that isn't
directly self-contained to Incus), you must back this up separately.**
```
````

but no languages got it similarly bolded.  

All symbols inside a code block
are rendered so the asterisks  ( ** ) are rendered too instead of making text bold.  

Given the incompatibility in styles and boxes, 
I'll go back to the Sphinx format and just break the lines
to keep it from running off the page . 

A new push will be coming shortly.

Comment (stgraber) [reply doc/backup.md]: Thanks!

**Comment by stgraber** [comment]
image

So that PR is still breaking the Sphinx warning. We may need to work out how to put exceptions in the linter for those. We don't really care about how things render on Github so long as the .md is readable/reviewable. What we care about is how it renders on the actual doc site.

**Comment by AJRepo** [comment]
Sorry about that. 

I figured out what was going on with my display of it and have now tested several sphinx renderings vs different browser renderings.  I was able to get both a good rendered Sphinx doc as well as markdown-linter-compliant github markdown where both display well online.

The reason I started this PR is because I did an online search for "incus backup" and it was the github page suggested and that's what I used as a primary source. 

Thank you for your patience and guidance.

A new commit will be inbound after this comment. 

---

(Notes for future readers who  (like I was) might be new to docs here using Sphinx): 

1. If editing `doc/FOO.md`, then the Sphinx rendered document will be in `doc/html/FOO/index.html` 
2. Generate docs with `make doc` .  E.g. To check in bash use: 
```bash
make doc
vi doc/html/FOO/index.html
```
3. If using Linux via a browser installed with `snap,`  the browser will not render the `file:///...doc/html/FOO/index.html`  similarly to the online version if the file is in a directory under a soft link to a different partition than your home directory.  The browser will silently deny some css for security reasons.

PR #2896, opened 2026-02-05
Add support for QCOW2 migration whether the instance is stopped

presztak:


stgraber:
What was causing the issue/corruption when migrating an instance into lvmcluster?

presztak:
@stgraber I compared a "fresh" block with one after migration and found that data was missing at the end of the block. Initially, I suspected that data was not being flushed but it turned out to be related to how zero blocks are handled. `qemu-nbd` has a `detect-zeroes` option that affects this behavior.

Issue #2807, opened 2026-01-08
Allow sending backups with no intermediary file


bensmrs:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

`incus export`, and more generally the backup API, writes the exported tarballs on disk. In scenarios where we do not have enough space on disk or where we would like to process the data stream with low latency, this intermediate step is problematic.

### What do you think would need to be added

There are several ways to see this:
- a `direct` flag in the backup API endpoint allowing the data to be exported on demand through `/export` without dumping to disk, to minimize friction with the existing clients,
- a new `GET` API endpoint that behaves like `/export` but without a backup ID, exporting data on the go.

`incus export` should also use this feature

stgraber:
I think we should keep the API through POST as that makes it easy to send in JSON data for what should be in the backup. It would also in theory allow our S3 logic to also use streaming, making it possible to pass in the S3 bucket info and key through the normal JSON request as we do today, but then have the server treat that as a streaming request and skip the whole writing to disk + delete logic.

bensmrs:
Alright, but then there will be a kind of “virtual” backup created, so that means potential changes to the DB (except if we allow a magic value for `name` that does it), and we still need to delete the backup afterwards. What would you do? Add a `virtual` boolean to the backup tables?

stgraber:
I don't know that we'd need a database entry at all if we're not going to have any file associated with it.

In the S3 case, we'd just ban the `name` key as it's pointless, then process the request and stream it straight into S3, no need for anything on disk or in the database.

For a regular streamed backup, the client would likely need to opt into it with something like `Accept: application/octet-stream`. The `name` property would similarly be banned in that scenario. We may still want to ask for an additional property in the request to opt into it being streamed, but we'll then just have the API call return the raw data directly, no need for DB records or files.

bensmrs:
Ah, so the regular POST, but no `/export` needed as it’s the POST endpoint that returns the data?
Works for me.

stgraber:
Yep, exactly, it won't ever show up in the collection and you'll never use /export, it's just the initial POST that gets you the data directly.

bensmrs:
You can mark this one for 6.22, it’s a bottleneck that I want out pretty soon :)

bensmrs:
[...]

PR #2899, opened 2026-02-05
Implement direct backup exports

bensmrs:
Closes: #2807

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -2987,3 +2987,7 @@ This introduces two new properties for NICs:
 This introduces a new property for SR-IOV NICs:
 
 * `security.trusted` allows setting the `trusted` flag for the virtual function if the parent NIC supports it.
+
+## `direct_backup`
+
+This allows to perform backups of instances, custom storage volumes and storage buckets without disk buffering when the `Accept` header is set to `application/octet-stream` in the respective POST API endpoints. Doing so, said POST endpoints return the data stream instead of creating a backup entry in the database.
--- a/internal/version/api.go
+++ b/internal/version/api.go
@@ -518,6 +518,7 @@ var APIExtensions = []string{
 	"network_zones_dns_contact",
 	"nic_attached_connected",
 	"nic_sriov_security_trusted",
+	"direct_backup",
 }
 
 // APIExtensionsCount returns the number of available API extensions.


bensmrs:
@stgraber, quick notes and questions.
1. You mentioned that we should forbid using the `name` parameter when performing direct backups; there are several places in the code where that could happen, should I put the check in the REST API functions in `cmd/incusd`?
2. I noticed that image backup lifecycle events are done in `cmd/incusd/backup.go`, but that is not the case for the others (storage volumes and buckets); any particular reason or just legacy?
3. I didn’t care about making the operations cancellable; do you insist on that?

stgraber:
> You mentioned that we should forbid using the `name` parameter when performing direct backups; there are several places in the code where that could happen, should I put the check in the REST API functions in `cmd/incusd`?

Yeah, request validation should happen as early as possible, so directly in the Post handler would be good.

stgraber:
> I noticed that image backup lifecycle events are done in `cmd/incusd/backup.go`, but that is not the case for the others (storage volumes and buckets); any particular reason or just legacy?

I don't think there's any particular reason for that. The package setup for backups is a bit unusual so that may have had something to do with it. In practice, so long as it's done in a spot that's reliable, we're fine :)

stgraber:
> I didn’t care about making the operations cancellable; do you insist on that?

Nah, very few operations are cancellable, it's basically limited to operations that download stuff from the Internet and which can therefore be easily cancelled.

In the case of streaming backups out, the background operation wouldn't even really be needed as the API call is synchronous in this case and a cancellation would come in the way of a closed connection from the user having closed their side of the connection. That will make all write operations fail which will in turn cancel everything.

bensmrs:
> I don't think there's any particular reason for that. The package setup for backups is a bit unusual so that may have had something to do with it. In practice, so long as it's done in a spot that's reliable, we're fine :)

Alright, I’m wondering if streamed backups should trigger a lifecycle event. I don’t know if these events implicitly convey some kind of object creation on the server, in which case they should be disabled.

> Nah, very few operations are cancellable, it's basically limited to operations that download stuff from the Internet and which can therefore be easily cancelled.

Well, closing a pipe is also pretty easy, it’s just that I felt lazy.

> In the case of streaming backups out, the background operation wouldn't even really be needed as the API call is synchronous in this case and a cancellation would come in the way of a closed connection from the user having closed their side of the connection. That will make all write operations fail which will in turn cancel everything.

Agreed, but I think it’s pretty nice to have a reporting that this action is currently running. In a way, console access isn’t completely asynchronous either, from a high level (there’s bidirectional data that can only flow if the consumer stays there).
Now, imagine that I’m the cluster admin and I want to free up some space, really quickly. A user is backing up a 1TiB instance and I need that TiB now, so I’m deleting the instance. My understanding is that the data will only be cleared when the user has finished their backup, which can impact my usage of the cluster.

bensmrs:
Maybe I’m whatifing too much :)

stgraber:
> Alright, I’m wondering if streamed backups should trigger a lifecycle event. I don’t know if these events implicitly convey some kind of object creation on the server, in which case they should be disabled.

The event should be emitted. Lifecycle events are meant for auditing activity on an Incus server.

bensmrs:
Eh, there are quite a few more edge cases to fix. Pipes are fun. I also need to prevent creating DB entries.

bensmrs:
Alright, that should be good now.

client/incus_instances.go
```diff
@@ -2998,6 +2999,71 @@ func (r *ProtocolIncus) GetInstanceBackupFile(instanceName string, name string,
 	return &resp, nil
 }
 
+// GetInstanceDirectBackupFile requests that Incus creates and returns new direct backup for the
+// instance.
+func (r *ProtocolIncus) GetInstanceDirectBackupFile(instanceName string, backup api.InstanceBackupsPost, req *BackupFileRequest) error {
```
stgraber: I'd call that `CreateInstanceBackupStream` instead. Having a `GetXYZ` for a `POST` operation feels wrong :)

stgraber: Also function should have the API extension check.

client/incus_storage_volumes.go
```diff
@@ -1095,6 +1096,66 @@ func (r *ProtocolIncus) GetStorageVolumeBackupFile(pool string, volName string,
 	return &resp, nil
 }
 
+// GetStorageVolumeDirectBackupFile requests that Incus creates and returns new direct backup for
+// the storage volume.
+func (r *ProtocolIncus) GetStorageVolumeDirectBackupFile(pool string, volName string, backup api.StorageVolumeBackupsPost, req *BackupFileRequest) error {
```
stgraber: Same as with instances, that should be `CreateStorageVolumeBackupStream` and have an API extension check.

client/incus_storage_buckets.go
```diff
@@ -505,6 +507,66 @@ func (r *ProtocolIncus) GetStoragePoolBucketBackupFile(pool string, bucketName s
 	return &resp, nil
 }
 
+// GetStoragePoolBucketDirectBackupFile requests that Incus creates and returns new direct backup
+// for the storage bucket.
+func (r *ProtocolIncus) GetStoragePoolBucketDirectBackupFile(poolName string, bucketName string, backup api.StorageBucketBackupsPost, req *BackupFileRequest) error {
```
stgraber: And same here with `CreateStoragePoolBucketBackupStream` and API check.

internal/server/response/response.go
```diff
@@ -697,3 +697,31 @@ func (r *sftpResponse) Render(w http.ResponseWriter) error {
 
 	return nil
 }
+
+type pipeResponse struct {
```
stgraber, line 701: Not necessarily opposed to introducing this simpler response type, but what was the issue with using FileResponse instead?

I thought we had support for streaming data out like that by using a single file with the following set:
 - Identifier
 - Filename
 - File
 - FileSize
 - FileModified

stgraber:
This actually ended up looking a lot cleaner than I expected ;)

client/incus_instances.go
bensmrs: I thought that because the only call site had the API extension check, that wasn’t useful, but ok

internal/server/response/response.go
bensmrs: The issue is this: https://cs.opensource.google/go/go/+/refs/tags/go1.25.7:src/net/http/fs.go;l=247

We really don’t want to seek to the end of our stream :)
And there are plenty of other hidden features of ServeContent that are not well-suited for streamed data.

bensmrs:
Ok I’ll do those quick patches later today.

bensmrs:
I got caught up in other matters, I’m looking into this now.
(Rebasing in a separate force-push to simplify auditing the diffs)

internal/server/response/response.go
stgraber: Fair enough :)

PR #2903, opened 2026-02-06
SR-IOV: Implement support for trust property

### LINKED ISSUE #2864: SR-IOV NICs - Ability to set the trusted property for the VF

Setting the trusted property on SR-IOV VF NICs is currently not supported by Incus.

This feature is necessary to configure active-passive failover bonds inside a virtual machine, as the bond driver attempts to change the MAC addresses on its assigned interfaces on failover, or immediately when adding them to the bond. An [OpenStack specification](https://specs.openstack.org/openstack/nova-specs/specs/rocky/implemented/sriov-trusted-vfs.html) also confirms this necessity.

On the [Incus forums](https://discuss.linuxcontainers.org/t/configure-trust-on-sriov-network/25026), there was also another user that had a need for setting this flag for VRRP support.

A new boolean config property `security.trusted` should be introduced for SR-IOV NICs that allows setting the trusted flag. It should be `false` by default.

@stgraber recommended using the same logic as for the other VF properties to implement this on the [forum](https://discuss.linuxcontainers.org/t/configure-trust-on-sriov-network/25026/5).

Shaderbug:
I've tested this on my development machine and seems like it works for me™

For the commit structure and changes, I've tried to orient myself at previous work done on the files, and similar PRs.

Looking forward to your feedback :)

Closes #2864

### DIFF
--- a/doc/config_options.txt
+++ b/doc/config_options.txt
@@ -1554,6 +1554,14 @@ The custom policy routing table ID to add IPv6 static routes to (in addition to 
 
 ```
 
+```{config:option} security.trusted devices-nic_sriov
+:default: "false, if supported by parent device"
+:managed: "no"
+:shortdesc: "Allows the instance to configure the NIC in ways that may negatively impact security."
+:type: "bool"
+
+```
+
 ```{config:option} vendorid devices-nic_sriov
 :required: "no"
 :shortdesc: "The vendor ID of the parent host device"
@@ -2940,6 +2948,12 @@ The parent host device used when allocating a VF into an instance.
 :shortdesc: "SR-IOV virtual function original spoof check setting"
 :type: "string"
 The original spoof check setting used when moving a VF into an instance.
+```
+
+```{config:option} volatile.<name>.last_state.vf.trusted instance-volatile
+:shortdesc: "SR-IOV virtual function original trusted setting"
+:type: "string"
+The original trusted setting used when moving a VF into an instance.
 ```
 
 ```{config:option} volatile.<name>.last_state.vf.vlan instance-volatile
--- a/internal/server/metadata/configuration.json
+++ b/internal/server/metadata/configuration.json
@@ -1761,6 +1761,15 @@
 						}
 					},
 					{
+						"security.trusted": {
+							"default": "false, if supported by parent device",
+							"longdesc": "",
+							"managed": "no",
+							"shortdesc": "Allows the instance to configure the NIC in ways that may negatively impact security.",
+							"type": "bool"
+						}
+					},
+					{
 						"vendorid": {
 							"longdesc": "",
 							"required": "no",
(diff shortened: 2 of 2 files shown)


Shaderbug:
Oops, will fix the doc linter complaints

stgraber:
Looking into this one now

internal/server/ip/link.go
```diff
@@ -312,11 +327,14 @@ func (l *Link) GetVFInfo(vfID int) (VirtFuncInfo, error) {
 
 	for _, vf := range link.Attrs().Vfs {
 		if vf.ID == vfID {
+			// The value is an uint32 to also indicate lack of support (-1)
+			// https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/net/core/rtnetlink.c?h=linux-6.18.y#n1513
```
stgraber: I'm confused by this one. uint32 is unsigned so cannot hold a -1 value whereas a default int type would be able to accommodate the -1.

internal/server/device/device_utils_network.go
```diff
@@ -731,6 +731,11 @@ func networkSRIOVSetupVF(d deviceCommon, vfParent string, vfDevice string, vfID
 	volatile["last_state.vf.vlan"] = fmt.Sprintf("%d", vfInfo.VLAN)
 	volatile["last_state.vf.spoofcheck"] = fmt.Sprintf("%t", vfInfo.SpoofCheck)
 
+	// Only persist trust state if it's supported by the NIC.
+	if vfInfo.Trust != ^uint32(0) {
```
stgraber: Oh, so you're storing it as a uint32 and then check for the overflow to get the -1 value?

internal/server/device/device_utils_network.go
```diff
@@ -764,6 +769,20 @@ func networkSRIOVSetupVF(d deviceCommon, vfParent string, vfDevice string, vfID
 		}
 	}
 
+	// Setup VF trust setting if specified
+	if d.config["security.trust"] != "" {
+		if vfInfo.Trust == ^uint32(0) {
+			return vfPCIDev, 0, fmt.Errorf("Failed setting trust for vf %q: The driver claims no support for it", volatile["last_state.vf.id"])
```
stgraber: Maybe `The driver reported missing support for this feature`?

internal/server/device/nic.go
```diff
@@ -31,6 +31,7 @@ func nicValidationRules(requiredFields []string, optionalFields []string, instCo
 		"limits.max":                           validate.IsAny,
 		"limits.priority":                      validate.Optional(validate.IsUint32),
 		"security.mac_filtering":               validate.IsAny,
+		"security.trust":                       validate.Optional(validate.IsBool),
```
stgraber: I'd go for `security.trusted` instead.

internal/server/device/device_utils_network.go
Shaderbug: This value was a bit of an odd situation for me as well. 

The kernel code assigns `-1` to that `__u32` [field](https://github.com/torvalds/linux/blob/master/include/linux/if_link.h#L30), which C seems to permit, just storing the two's complement of the number (i.e. all bits set).

The netlink package Incus includes then returns a `uint32` for that value, but Go doesn't seem to allow casting `-1` to `uint32` in a constant, so I had to go with not 0.

Maybe I should rephrase my comments a bit to make that clearer.

internal/server/device/nic.go
Shaderbug: Sure, I'll change that. This was also a bit of an unclear naming situation, as the `netlink` package calls the function `LinkSetVfTrust`, `ip link` calls it `trust` and the kernel struct then calls it `trusted` again.

Issue #2909, opened 2026-02-08
Add ability to disable memory ballooning for VM (QEMU) instances


edgjj:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

Currently there's no ability to disable ballooning, as virtio_balloon is added as fixed device on QEMU config generation.

Under some weird circumstances (e.g. VM instance of FreeBSD/opnsense), ballooning performs weirdly and just reserves entire memory limit - meanwhile real memory consumption is way lower.

In Proxmox it's possible to disable balloon device, and it solves the initial [issue](https://forum.opnsense.org/index.php?topic=15708.0), but with Incus i'm unable to try the same fix.

### What do you think would need to be added

Ability to disable ballooning device on QEMU VMs.

stgraber:
I don't think that we want a config key specifically to control ballooning as whether we're using ballooning or another mechanism should be invisible from the user. Adding a config key for it would then limit our options in the future.

For those kind of exceptions/workaround, using `raw.qemu.conf` is usually the easiest.
I just tested it here and it appears to be working as expected for the balloon device though given it's a function within a multi-function PCIe device, deleting it through `raw.qemu.conf` doesn't quite work, instead I had to replace it with a placeholder. This looks like this:

```
  raw.qemu.conf: |-
    [device "qemu_balloon"]
    driver = "virtio-serial-pci"
```

And it behaves as expected without any further changes:
```
stgraber@castiana:~$ incus config set d13 limits.memory=512MiB
Error: Failed updating memory limit: DeviceNotActive: No balloon device has been activated
```

Closing as we don't really want to expose such low level config options as a primary configuration key and there is a viable option through existing configuration keys.

PR #2910, opened 2026-02-08
incusd/instance/drivers: Add limits.memory.balloon config option

fatyzzz:
## What this PR does

Adds a new VM-only configuration option `limits.memory.balloon` (bool, default `true`) that allows disabling the `virtio-balloon` device on QEMU VMs.

Fixes https://github.com/lxc/incus/issues/2909

Currently there's no ability to disable ballooning, as `virtio_balloon` is added as a fixed device on QEMU config generation.

Under some circumstances (e.g. VM instances of FreeBSD/OPNsense), ballooning performs weirdly and reserves the entire memory limit — meanwhile real memory consumption is way lower.

In Proxmox it's possible to disable the balloon device, and it solves the issue, but with Incus there's no way to do the same.

### What was added

A new `limits.memory.balloon` option. When set to `false`:

- The balloon device (`virtio-balloon-pci`/`virtio-balloon-ccw`) is **not** added to the QEMU config
- The PCI bus slot is still consumed to preserve device address ordering (stable NIC naming like `enp5s0`)
- Memory **increase** via hotplug continues to work normally
- Memory **decrease** returns an error since balloon is unavailable

### Usage

```
incus config set <vm> limits.memory.balloon=false
incus restart <vm>
```

### Files changed

- `internal/instance/config.go` — new config key in `InstanceConfigKeysVM`
- `internal/server/instance/drivers/driver_qemu.go` — conditional balloon device, restructured `updateMemoryLimit` to separate hotplug/balloon paths
- `doc/reference/instance_options.md` — documentation note
- `doc/config_options.txt` — auto-regenerated
- `internal/server/metadata/configuration.json` — auto-regenerated

### DIFF
--- a/doc/config_options.txt
+++ b/doc/config_options.txt
@@ -2393,6 +2393,17 @@ Various suffixes are supported.
 See {ref}`instances-limit-units` for details.
 ```
 
+```{config:option} limits.memory.balloon instance-resource-limits
+:condition: "virtual machine"
+:defaultdesc: "`true`"
+:liveupdate: "no"
+:shortdesc: "Whether to enable the memory balloon device"
+:type: "bool"
+Set this option to `false` to disable the memory balloon device.
+This can help with guests (such as FreeBSD) that don't handle ballooning well.
+Note that this also disables live memory reduction for the VM.
+```
+
 ```{config:option} limits.memory.enforce instance-resource-limits
 :condition: "container"
 :defaultdesc: "`hard`"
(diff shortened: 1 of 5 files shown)


stgraber:
Per the associated issue, there is a pretty easy way to get rid of ballooning through `raw.qemu.conf` and doing so avoids exposing internal details through primary configuration keys like `limits.memory`.

PR #2934, opened 2026-02-14
Improve live migration handling

stgraber:


stgraber:
@presztak the first commit here fixes a regression with the recent live-migration changes

PR #2941, opened 2026-02-15
Prevent concurrent QCOW2 snapshot deletions

presztak:
Closes: https://github.com/lxc/incus/issues/2930


stgraber:
@presztak just noticed you had left it in draft. The code looks pretty simple and correct to me. Was there a reason to have it remain draft?

presztak:
@stgraber, so, I was thinking whether we could improve this further. In theory, we could update snapshots that are not directly dependent in parallel. For now, though, let’s go with the current approach and consider further optimizations in the future.

stgraber:
@presztak I think it'd also be fine to move the locking to a more generic code path as even outside of QCOW2, it's probably a good idea to only handle one snapshot deletion at the time.

PR #2958, opened 2026-02-19
Windows agent over vsock

masnax:


antifob:
@masnax  This seems to have broken agents. Is there anything special needed on the VM's side to support this?

image

antifob:
All right, I had a bit of catching up to do.

This requires the viosock driver installed from the latest+stable (as of today, 0.1.285) virtio-win build. It also needs to be "forcibly" installed manually as it's not installed by running the msi.

Since the agent doesn't fall back to the TCP listener, any agent without that driver is effectively not reachable.

I'm planning to do some work on the agent, any interest for a fallback mechanism?

https://github.com/virtio-win/kvm-guest-drivers-windows/issues/1185#issuecomment-3292652571
https://fedorapeople.org/groups/virt/virtio-win/direct-downloads/archive-virtio/virtio-win-0.1.285-1/

antifob:
Last tidbit. It's unclear what's responsible (qemu, incus, or viosock), but it's super unstable. The whole VM crashes when the service restarts unexpectedly on an active session. I would revert this until it's investigated. A simple reproducer that sometimes work.
```
incus exec -T vm -- powershell.exe -c 'stop-process -force -name incus-agent'
```

stgraber:
Yeah, the state of the VirtIO driver and its installer is a bit unfortunate.
It looks like the next signed build should improve things a bit.

I'd rather not implement a fallback mechanism if we can avoid it as there are security implications from offering the HTTPS fallback, especially in multi-tenant environments where the user owns their own networks (OVN) and so can effectively get the host to connect to arbitrary addresses. Not that this would result in anything particularly successful given we perform mutual TLS validation on top of the connection, but it's still an attack surface I'd like to limit.

My current feeling is that very very few people have the Windows agent in place at this stage. While I implemented it a little while back now, it's only become more widely usable over the past couple of releases through the work of @MOBergeron on both the install drive logic and the Ansible support, then made easier to use through your own work.

We know of some issues with the vsock kernel driver on Windows which can cause a crash, the service stop while there's an active connection is likely one of those. We haven't been able to identify any scenario where this can be used by an unprivileged user to cause a VM crash or escalate privileges so while we're hoping to see an updated VirtIO driver in the coming weeks, the security implications of the former logic (HTTPS) are for us more problematic than those of the vsock driver and service.

PR #2971, opened 2026-02-24
incusd/storage/linstor: Propagate error when volume filler fails

luissimas:
The current implementation was incorrectly suppressing errors that
occurred in the filler function when creating volumes via the Linstor
driver.

When any error ocurred in the filler function, the reversion logic would
delete the volume, but the return nil would signal to the caller that
the operation succeeded, causing a "Resource definition not found" error
when performing any subsequent Linstor operations on the volume.

Signed-off-by: Luís Simas <luissimas@protonmail.com>


luissimas:
Most likely the cause to #2964

Issue #2897, opened 2026-02-05
Extend authorization scriptlet (certificate and instance details)


fatyzzz:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

Currently, Incus certificate-based access control works at the project level. When using `incus config trust add --restricted --projects <project>`, the certificate gains access to all instances within that project.

In CI/CD environments (GitLab CI, GitHub Actions, etc.) with Ansible automation, this creates a security concern: if a runner is compromised, the attacker gains exec access to ALL instances in the project, not just the ones that specific CI job needs.

For example, I have a project with 15 instances including sensitive ones (Vault, GitLab, Keycloak) and less sensitive ones (Grafana, YouTrack). I want my `deploy-grafana` CI job to only have exec access to the `grafana` instance, not to `vault` or `gitlab`.

The current workarounds are:
- **Multiple projects** — splits infrastructure awkwardly, complicates networking and management
- **OpenFGA integration** — requires deploying and maintaining a separate authorization service, significant operational overhead for simple use cases
- **SSH keys per instance** — works but adds complexity (key management, rotation, sshd in containers) that Incus exec elegantly avoids

### What do you think would need to be added

Add the ability to restrict certificate access to specific instances, not just projects. Something like:

```bash
# Restrict certificate to specific instances within a project
incus config trust add ansible-grafana \
  --restricted \
  --projects myproject \
  --instances grafana,monitoring

# Or using instance patterns
incus config trust add ansible-runners \
  --restricted \
  --projects myproject \
  --instances "gitlab-runner-*"
```

This would allow the certificate to:
- List/view only specified instances
- Exec/console/files access only to specified instances
- No access to other instances in the project

This fits well with the existing `--restricted` model and would make Incus much more suitable for secure CI/CD automation without requiring external authorization systems.

stgraber:
Did you try the authorization scriptlet?
https://linuxcontainers.org/incus/docs/main/authorization/#scriptlet-authorization

fatyzzz:
> Did you try the authorization scriptlet? https://linuxcontainers.org/incus/docs/main/authorization/#scriptlet-authorization


I looked into the authorization scriptlet approach, but I think per-instance restriction would be much better as a native feature (extending the existing `--restricted` model) rather than a scriptlet workaround. Here's my reasoning:

**Why scriptlets don't fit this use case well:**

[...]

PR #2973, opened 2026-02-25
Expose certificate data in authorization scriptlet

bensmrs:
Closes: #2897

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -3012,3 +3012,9 @@ This allows specifying a comma separate list of image server domains from which 
 ## `storage_lvmcluster_size`
 
 Adds support for specifying the `size` parameter when updating `lvmcluster` storage pool.
+
+## `authorization_scriptlet_cert`
+
+This adds two fields, `Chain` and `Certificate`, to the `details`
+argument of the authorization scriptlet, to give more information about
+the client to authorize.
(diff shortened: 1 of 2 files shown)


bensmrs:
For now, I’m just checking whether this commit breaks anything.

bensmrs:
I just copied some logic without thinking, but why are we recomputing the fingerprint here https://github.com/lxc/incus/blob/51b9ca34bb4cbf909ad7bac6bb23e2bd70edee69/cmd/incusd/certificates.go#L290

when we already have it in the DB, @stgraber? Am I missing something or can I just remove the costly calls?

stgraber:
Can't think of a good reason why we're doing this, not like the two can get out of sync so should be fine to just use the DB value

bensmrs:
Not sure how forwarded requests work. Does `r.TLS` correspond to the forwarder’s chain or to the client’s? In the first case, that would prevent us from getting the chain (and would require a quick patch to not include the forwarder’s chain if the request indeed is forwarded).

bensmrs:
Hmmm, the instrumentation in `permissions_test` is a bit broken now that the certificate parsing is deferred to the cache. Maybe I’ll need to generate a dummy certificate.

bensmrs:
Alright, that should do the trick. Depending on https://github.com/lxc/incus/pull/2973#issuecomment-3959654258, additional work may be needed.
Edit: not sure why the Markdown linter complains…

Issue #2944, opened 2026-02-16
`incus file mount` and `incus storage volume file mount` don’t perform enough checks


bensmrs:
### Is there an existing issue for this?

- [x] There is no existing issue for this bug

### Is this happening on an up to date version of Incus?

- [x] This is happening on a supported version of Incus

### Incus system details

```yaml
HEAD
```

### Current behavior

`incus file mount` and `incus storage volume file mount` don’t check whether instances, storage pools or storage volumes exist.

### Expected behavior

An error if the instance, the storage pool or the storage volume don’t exist.

### Steps to reproduce

```
> incus file mount xxxxxx
SSH SFTP listening on 127.0.0.1:54786
Login with username "QmfQk43Y" and password "e7sXMtza"
^C
> incus storage volume file mount xxxxxx yyyyyy
SSH SFTP listening on 127.0.0.1:54843
Login with username "ZMPIZdxS" and password "XU1rnoTz"
```

J0nasDotDev:
Hi @stgraber 👋

I’d like to take this issue as my first contribution. Even as a user, I’m quite new to the project, but this issue looks very approachable.

As far as I understand, it simply involves adding checks just before calling `sshSFTPServer` in these two places:
- https://github.com/lxc/incus/blob/v6.21.0/cmd/incus/file.go#L1081-L1082
- https://github.com/lxc/incus/blob/v6.21.0/cmd/incus/storage_volume.go#L2782-L2785

Regarding tests, would you like me to add some _negative_ integration tests for this?
If so, I will probably try to add them around some of these code blocks:
- https://github.com/lxc/incus/blob/v6.21.0/test/suites/basic.sh#L439-L456
- https://github.com/lxc/incus/blob/v6.21.0/test/suites/filemanip.sh#L145-L157
- https://github.com/lxc/incus/blob/v6.21.0/test/suites/storage_volume_filemanip.sh#L71-L73

stgraber:
Yep, that sounds right as far as what needs updating.

For testing, the only place where we test `incus file mount` is in `test/suites/filemanip.sh` so it would make sense to include a test there, then add something similar to `test/suites/storage_volume_filemanip.sh` to cover the volume case.

PR #2974, opened 2026-02-25
incus: Add checks before starting sshSFTPserver

J0nasDotDev:
## Summary
Fix `incus file mount` and `incus storage volume file mount` in SSH SFTP listener mode to fail early when the entity (instance, storage pool, or storage volume) does not exist.
Previously, these commands could start the listener process even non-existent entities, resulting in confusing runtime behavior.

Resolves: #2944

## Description
The `sshSFTPServer` doesn't require the entity directly as a parameter, they are used indirectly through the `sftpConn` lambda function.

These changes try to get the entity before calling `sshSFTPServer` and fail early if the entity does not exist.

Note: `sftpConn` is only called after an SSH connection, and `sshSFTPServer` doesn't return an error if `sftpConn` fails.

I also updated the help texts to clarify the listener behavior when the target path is omitted.

## Behavior after changes
```
> incus file mount xxxxxx
Error: Failed to fetch instance "xxxxxx" in project "default": Instance not found
> incus storage volume file mount xxxxxx yyyyyy
Error: Storage pool not found
> incus storage volume file mount default yyyyyy
Error: Storage pool volume not found 
```

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: LXD\n"
 "Report-Msgid-Bugs-To: lxc-devel@lists.linuxcontainers.org\n"
-"POT-Creation-Date: 2026-02-16 16:31+0000\n"
+"POT-Creation-Date: 2026-02-25 20:21-0500\n"
 "PO-Revision-Date: 2025-12-27 03:16+0000\n"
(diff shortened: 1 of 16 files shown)


J0nasDotDev:
@stgraber FYI, I updated the translation templates https://github.com/lxc/incus/pull/2974/changes/64bfdc2b0f9b8bf36fc6095e05f27c6511325fa2, so normally the CI should pass.

Issue #2963, opened 2026-02-22
`incus cluster set` fails when there is only one node


bensmrs:
### Is there an existing issue for this?

- [x] There is no existing issue for this bug

### Is this happening on an up to date version of Incus?

- [x] This is happening on a supported version of Incus

### Incus system details

```yaml
HEAD
```

### Current behavior

https://github.com/lxc/incus/blob/7c1eb23a3c504b6f4a72ffc2c6231f6afa8db504/cmd/incusd/api_cluster.go#L1821
always passes when the node count is 1.

I’m wondering why we are i) skipping the current node in the DB tx and ii) not checking whether `database-client` is asked for the current node.

### Steps to reproduce

```
# incus cluster set foobar user.foo=bar
Error: Assigning the 'database-client' role to all nodes is not allowed
```

bensmrs:
I see many open issues remaining for 6.22; want me to look at this one tomorrow?

stgraber:
@bensmrs that'd be great, I'm running a bit behind on reviewing stuff let alone working on issues :)

PR #2982, opened 2026-02-26
incusd/api_cluster: Fix database-client count logic

bensmrs:
Closes: #2963


bensmrs:
Those are mostly indentation changes, so a diff without whitespaces helps :)

Issue #2948, opened 2026-02-17
`incus copy src dst -c security.secureboot=false` does not update ovmf vars


masnax:
### Is there an existing issue for this?

- [x] There is no existing issue for this bug

### Is this happening on an up to date version of Incus?

- [x] This is happening on a supported version of Incus

### Incus system details

```yaml
server_version: "6.21"
```

### Steps to reproduce

If `src` has `security.secureboot=true`, performing `incus copy src dst -c security.secureboot=false` will not update the `OVMF_VARS` file, making the VM unbootable.

After copy:
```
-rw-r--r-- 1 root  root      540672 Feb 16 21:07 OVMF_VARS.4MB.ms.fd
lrwxrwxrwx 1 root  root          19 Feb 16 21:07 qemu.nvram -> OVMF_VARS.4MB.ms.fd
```
If then performing `incus config set security.secureboot=true` followed by `incus config set security.secureboot=false`, then the files will be updated:

```
-rw-r--r-- 1 root  root      540672 Feb 17 11:52 OVMF_VARS.4MB.fd
lrwxrwxrwx 1 root  root          16 Feb 17 11:52 qemu.nvram -> OVMF_VARS.4MB.fd
```

bensmrs:
I can have a look at this one also

bensmrs:
Ok I’ve started working on it and also reproduced the behavior with local move. While patching it, a question came to my mind: in https://github.com/lxc/incus/blob/324c07b6c34a2ed5c488ced1d594f87ec005cc81/cmd/incusd/instance_post.go#L566-L604

aren’t we skipping the configuration override logic? Is it ok because as the source pool is remote, no config override is allowed?

Anyway, off to patching the logic in `instanceS_post.go` now.

stgraber:
@bensmrs hmm, so that's a bit confusing. Initially it looked like you were right and that we effectively have a code path where you could move an instance away from a dead server AND apply a config/device override with the latter being lost due to that code path not handling it.

But looking at the CLI logic, the presence of any config/device/profile override causes it to skip the server-side migration logic entirely, which in the case of a dead server means that the migration will immediately fail to the source server being offline.

So it looks like with our current implementation, it shouldn't be possible to hit that code path (server-side move of instance on shared storage with an offline server) while also having an override present.

bensmrs:
[...]

PR #2983, opened 2026-02-26
incusd/instances: Properly instruct to reset NVRAM when changing secure boot config

bensmrs:
Closes: #2948


bensmrs:
@masnax, could you confirm that works for you, in case I’m missing something?

masnax:
> @masnax, could you confirm that works for you, in case I’m missing something?

Yep, fixed for me. Thanks :)

bensmrs:
Well now that I’m thinking about it, `volatile.apply_nvram` should be included in `InstanceIncludeWhenCopying`, so that when copying a VM with pending changes, the changes are also applied in the copied instance. I’ll push something in a few minutes.

PR #2997, opened 2026-03-02
firewall: fix linter complaints

J0nasDotDev:
This PR relates to #2636 and address the lint issues in `internal/server/firewall/...`.


stgraber:
```
Checking that functional blocks are followed by newlines...
internal/server/firewall/drivers/drivers_xtables.go:396
internal/server/firewall/drivers/drivers_xtables.go:415
internal/server/firewall/drivers/drivers_xtables.go:710
```

J0nasDotDev:
I added newlines after the functional blocks.

Is it expected that there is no configured linter or formatter that enforces this rule?

stgraber:
> Is it expected that there is no configured linter or formatter that enforces this rule?

Yeah. It's only pretty recently that wsl can handle this particular one and we haven't yet switched the check over to it.

PR #3000, opened 2026-03-03
server util: fix linter complaints

J0nasDotDev:
This PR relates to #2636 and address the lint issues in `internal/server/util/....

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -75,6 +75,8 @@ linters:
         source: '^// Example:'
       - path: internal/util/
         text: "avoid meaningless package names"
+      - path: internal/server/util/
+        text: "avoid meaningless package names"
     paths:
       - third_party$
       - builtin$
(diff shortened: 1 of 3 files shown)


stgraber:
Instead of nolint for the packages named `public`, it's probably better to ignore the specific error as we've done within incus-os.

See: https://github.com/lxc/incus-os/blob/main/incus-osd/.golangci.yml#L89

.golangci.yml
```diff
@@ -75,6 +75,8 @@ linters:
         source: '^// Example:'
       - path: internal/util/
         text: "avoid meaningless package names"
+      - path: internal/server/util/
+        text: "avoid meaningless package names"
```
J0nasDotDev, line 79: @stgraber Indeed, it's way better to adapt `golangci-lint` config file.
It was a rather lazy approach on my part to 'nolint' the name of the package.

PR #2945, opened 2026-02-16
incus: Refactor CLI usage strings

bensmrs:
~~This PR was very fun to make~~

This is an opinionated PR aiming at standardizing the CLI usage strings and fixing quite a few bugs along the way. I’m hoping this new usage description method will also be used in `incus-os`, so that we can completely remove the `Usage` function.

This (manual) work is not done only for the sake of style:
* It eases the work of translators, by having them translate a few standard terms (that are often already translated) instead of a bunch of usage lines,
* It allows (in future PRs) to automatically perform checks, such as the argument count, clever pattern matching, and higher order command generation using only the usage description.

Related-to: #2877

# Fixes
* `admin recover [<remote>]:` -> `admin recover [<remote>:]` (misplaced colon)
* `admin sql [<remote>]:<local|global> <query>` -> `admin sql [<remote>:](local|global) <query>` (misplaced colon)
* `copy [<remote>:]<source>[/<snapshot>] [[<remote>:]<destination>]` -> `copy [<remote>:]<instance>[/<snapshot>] [<remote>:][<new instance name>]` (misplaced bracket)
* `image alias list [<remote>:] [<filters>...]` -> `image alias list [<remote>:] [<filter>...]` (misleading plural)
* `network attach [<remote>:]<network> <instance> [<device name>] [<interface name>]` -> `network attach [<remote>:]<network> <instance> [<new device name> [<new interface name>]]` (misplaced bracket)
* `network attach-profile [<remote>:]<network> <profile> [<device name>] [<interface name>]` -> `network attach-profile [<remote>:]<network> <profile> [<new device name> [<new interface name>]]` (misplaced bracket)
* `network forward port remove [<remote>:]<network> <listen_address> [<protocol>] [<listen_port(s)>]` -> `network forward port remove [<remote>:]<network> <listen address> [<protocol> [<listen port>[,<listen port>...]]]` (misplaced bracket)
* `network load-balancer port remove [<remote>:]<network> <listen_address> [<protocol>] [<listen_port(s)>]` -> `network load-balancer port remove [<remote>:]<network> <listen address> [<protocol> [<listen port>[,<listen port>...]]]` (misplaced bracket)
* `storage volume attach [<remote>:]<pool> <volume> <instance> [<device name>] [<path>]` -> `storage volume attach [<remote>:]<pool> <volume> <instance> [<new device name> [<path>]]` (misplaced bracket)
[...]

Issue #2877, opened 2026-01-30
Cleanup command descriptions


bensmrs:
### Is there an existing issue for this?

- [x] There is no existing issue for this bug

### Is this happening on an up to date version of Incus?

- [x] This is happening on a supported version of Incus

### Incus system details

```yaml
Irrelevant
```

### Current behavior

While translating command descriptions, I noticed that their descriptions are very irregular, and some elements are either inconsistent or just wrong:
- Some descriptions (both short and long) end with a period and some don’t;
- Some long descriptions are the same as the short ones, but with one or two words rephrased (which feels useless), sometimes with a singular/plural mismatch;
- Some commands suggest that long column format specifiers are accepted while it’s not the case from the source code;
- Some commands indicate “[…] that control which XXX attributes to output […]” with “XXX” being the wrong word(s); this happens at other places in some command descriptions;
- Some filter descriptions suggest that matches are performed on substrings while they sometimes are performed on prefixes only;
- Some filter descriptions indicate that filters should be combined with commas, where they can actually be different command-line arguments (I think comma separation is legacy and shouldn’t even be documented anymore);
- The default column layout isn’t always shown at a consistent place;
- Some of these default column layout characters don’t match the actual allowed characters;
- Some `list` commands feature a `== Columns ==` separator (which by the way renders poorly on the web documentation) while some don’t.

### Expected behavior

(Marking this as a bug because it’s not just a matter of style and taste)

The above comments should be fixed. It’s rather painful to do but will make command descriptions much clearer. If the person tackling this happens to speak French, I’m suggesting them to have a look at what’s being done in the French translation.

I think some of these elements (such as column layout info) should be automatically generated from data present in the code. It would also have the nice effect for translators not to die of boredom when they reach the `incus <xxx> list` part of the translation :)

Also, do we still need all these line breaks? Content and disposition should be handled at different places if we really want to have fixed-width lines.

### Steps to reproduce

/

bensmrs:
I think I’ll start the grunt work next week. Doesn’t seem fair to just drop the bomb and do nothing about it :)

bensmrs:
Following #2945, I’ve had a look at automated parsing and my first tests are promising, but this `getConfig` is really bad parser-wise. I will have to implement an override mechanism under the hood to display `<key>=<value>` even though the actual parsing rule is `(<key>=<value>|<key> <value>)`…
I guess dropping the old behavior is out of the question, right?

stgraber:
[...]

PR #3007, opened 2026-03-04
Implement new parser for CLI client

bensmrs:
This mega-PR implements a new parser for our CLI client.

The parser is a simple modular greedy parser featuring parser combinators for all the composite atoms defined in #2945. Due to the absence of any kind of clever backtracking to keep good properties, the parser provides an alternative parsing mode to process its arguments from right to left in the case of CLI usages which would break typical left-to-right parsing.

This PR is divided in cumulatively-compilable commits that first set up the groundwork for the new parser, then implement it in the actual commands. Two command couples (`copy`/`move` and `create`/`launch`) are bundled together, as they are not trivially splittable.

The high-level goal for this PR is to remove all ad-hoc parsing and local juggling with arguments, providing a stable positional indexing for arguments. A side-quest was to remove calls to other commands’ `Run` methods to avoid double-parsing them. It also gives a nice visual debugger with `--explain`, that’s also invoked when users make mistakes.

Related-to: #2877

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: LXD\n"
 "Report-Msgid-Bugs-To: lxc-devel@lists.linuxcontainers.org\n"
-"POT-Creation-Date: 2026-02-25 20:21-0500\n"
+"POT-Creation-Date: 2026-03-05 12:55+0000\n"
 "PO-Revision-Date: 2025-12-27 03:16+0000\n"
 "Last-Translator: Dklfajsjfi49wefklsf32 "
 "<nlincus@users.noreply.hosted.weblate.org>\n"
@@ -19,19 +19,19 @@ msgstr ""
 "Plural-Forms: nplurals=2; plural=n != 1;\n"
 "X-Generator: Weblate 5.15.1\n"
 
-#: cmd/incus/info.go:454
+#: cmd/incus/info.go:438
 msgid "  Chassis:"
 msgstr "  Gehäuse:"
 
-#: cmd/incus/info.go:494
+#: cmd/incus/info.go:478
 msgid "  Firmware:"
 msgstr "  Firmware:"
 
-#: cmd/incus/info.go:474
+#: cmd/incus/info.go:458
 msgid "  Motherboard:"
 msgstr "  Mainboard:"
 
-#: cmd/incus/storage_bucket.go:291 cmd/incus/storage_bucket.go:1291
+#: cmd/incus/storage_bucket.go:262 cmd/incus/storage_bucket.go:1149
 #, fuzzy
 msgid ""
 "### This is a YAML representation of a storage bucket.\n"
@@ -56,7 +56,7 @@ msgstr ""
 "### config:\n"
 "###   size: \"61203283968\""
 
-#: cmd/incus/storage.go:304
+#: cmd/incus/storage.go:293
 #, fuzzy
 msgid ""
 "### This is a YAML representation of a storage pool.\n"
@@ -87,7 +87,7 @@ msgstr ""
 "###   source: /home/chb/mnt/lxd_test/default.img\n"
 "###   zfs.pool_name: default"
 
-#: cmd/incus/storage_volume.go:1024
+#: cmd/incus/storage_volume.go:935
 #, fuzzy
 msgid ""
 "### This is a YAML representation of a storage volume.\n"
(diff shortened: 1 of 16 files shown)


bensmrs:
Sorry.

bensmrs:
Now rebasing onto main.

bensmrs:
Where the tests are currently failing:
- ~~`-` doesn’t refer to STDIN in `incus config set` (TODO: fix)~~
- `""` doesn’t work anymore in `incus profile assign` <- I’m not convinced by the `assign ''` syntax, do we want to keep it as-is?
- `incus storage volume snapshot show incustest-rri container/c1/s1` <- do we still want to support this syntax?
- ~~`incus network address-set delete testAS` (TODO: investigate, probably index error)~~
- ~~`incus storage volume attach incustest-q9Y testvol c1 /mnt`~~ (WONTFIX, cf. https://github.com/lxc/incus/issues/2877#issuecomment-3987194905)
- ~~`incus start foo-path4` in `container devices - disk` succeeds instead of failing (TODO: investigate)~~
- ~~`incus start foo` in `container devices - nic - bridged` succeeds instead of failing (TODO: investigate)~~

bensmrs:
@stgraber, so these two failing tests require to make a potentially breaking decision on the CLI design:
- My changes make it so that passing `""` as the last argument of `incus profile assign` doesn’t work, as it now requires at least a profile name. Do we want to revert that to add support for the magic `""` value? Do we rather want to support using the command without any profile name ? Instead, do we want a special flag, e.g. `--none` to be more explicit?
- One test does `incus storage volume snapshot show incustest-rri container/c1/s1`. This looks like a legacy feature that we have moved to `incus snapshot show`. Do we still want to support this syntax?

Once this is decided, I’ll hope no other test fails.

After that, I’ll change my rune counting code to use https://github.com/mattn/go-runewidth, to properly support CJK characters (which take 2 columns in the terminal), as the parser’s debugger relies on horizontal alignment to help the user, and we have a nice JP community.

Finally, when running `make i18n`, `msgmerge` fails with `msgmerge: po/incus.pot: The file contains non-ASCII characters but the present charset "CHARSET" is not UTF-8.`. It may just be a whim of mine, but I’d really want to have those UTF-8 characters I have introduced, as they make the debug output clearer IMO, as curly quotes (`“` and `”`) don’t require us to escape inner quotes.

stgraber:
> My changes make it so that passing `""` as the last argument of `incus profile assign` doesn’t work, as it now requires at least a profile name. Do we want to revert that to add support for the magic `""` value? Do we rather want to support using the command without any profile name ? Instead, do we want a special flag, e.g. `--none` to be more explicit?

We have a `--no-profiles` in `create` and `launch`, so maybe re-use that here?

stgraber:
> One test does `incus storage volume snapshot show incustest-rri container/c1/s1`. This looks like a legacy feature that we have moved to `incus snapshot show`. Do we still want to support this syntax?

Should be `incus storage volume snapshot show incustest-rri container/c1 s1`

bensmrs:
> > One test does `incus storage volume snapshot show incustest-rri container/c1/s1`. This looks like a legacy feature that we have moved to `incus snapshot show`. Do we still want to support this syntax?
> 
> Should be `incus storage volume snapshot show incustest-rri container/c1 s1`

Ok so we should still allow to query them in `incus storage volume snapshot`, I made the assumption that `incus snapshot` would be the only way for them. Anyway, not a very complicated fix.

bensmrs:
> > One test does `incus storage volume snapshot show incustest-rri container/c1/s1`. This looks like a legacy feature that we have moved to `incus snapshot show`. Do we still want to support this syntax?
> 
> Should be `incus storage volume snapshot show incustest-rri container/c1 s1`

~~Hmmm, that’ll be too disruptive I think. I’m gonna switch back to the previous syntax and we can set the target for this new syntax to 7.0.~~ Nah scratch that, I’m implementing the change

bensmrs:
Ok that one’s pretty sad, but my opinion is to not fix it, @stgraber.
`incus config set foo systemd.credential-binary.foo YmF6=========` fails because given the parsing rules, it cannot deterministically be parsed the way we want (at least without backtracking). Because the `<key> <value>` syntax is deprecated, I’ll just patch this test to do `incus config set foo systemd.credential-binary.foo=YmF6=========` instead and hide that under the rug (that only happens for `incus config set`, when using the old-style K/V, when a value contains `=`, so it’s highly specific)

bensmrs:
Ok, looking great now :)
I’ll now have a look at CJK characters, but even though I’ve switched my system to `ja_JP.UTF-8`, Incus doesn’t switch to Japanese, so I’m surely missing something.

Edit: ah yeah, it was because of the way gettext is now broken…

bensmrs:
Alright, should be ready now, modulo the gettext problem which I don’t know how to solve (the charset is set to `CHARSET` by `xgettext-go` instead of `UTF-8`…). Shall I just patch the Makefile with a `sed`? (Edit: done in my last push).

PR #3021, opened 2026-03-11
Add project resource and limit metrics

Adevixxx:
Hey!

This adds metrics for project resources (containers, virtual-machines, images, profiles, networks, storage-volumes) and project limit/usage when limits are configured.

On a side note, I was looking at the existing warnings metric and wanted to add more details like the severity or status. Would it be more useful to have one metric per warning with labels (severity, status, type, project, location, uuid) or is the aggregated count enough?

### DIFF
--- a/doc/reference/provided_metrics.md
+++ b/doc/reference/provided_metrics.md
@@ -95,6 +95,23 @@ The following instance metrics are provided:
   - Current time from guest in seconds since epoch
 ```
 
+## Project metrics
+
+The following project metrics are provided:
+
+```{list-table}
+   :header-rows: 1
+
+* - Metric
+  - Description
+* - `incus_project_resources_total{project="<project>",resource="<resource>"}`
+  - Current count of resources in a project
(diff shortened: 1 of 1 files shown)


stgraber:
> On a side note, I was looking at the existing warnings metric and wanted to add more details like the severity or status. Would it be more useful to have one metric per warning with labels (severity, status, type, project, location, uuid) or is the aggregated count enough?

Yeah, the current warning metric is pretty useless, especially as it's counting "resolved" warnings too...

Thinking of it from a dashboard point of view, I'd like to easy be able to grab:
 - Total count of warnings
 - Total count of warnings for a given severity
 - Total count of warnings for a given server
 - Total count of warnings for a given project
 - Total count of warnings by status

So I'm not sure that the warning type is something we really need to expose as that's the most dynamic of the options and adding new metrics every time we introduce one would make it hard for dashboards to keep up, but having the ability to show a warning count and split by severity or status for a given project, or server would be useful.

PR #3025, opened 2026-03-12
Add initial support for dependent volumes

presztak:
This PR introduces a new `dependent` configuration key for disk devices and custom volumes.
Volumes marked as dependent are tied to the lifecycle of the instance they are attached to. Snapshot operations performed on the instance are automatically propagated to the dependent volume.

Specifically:
- Creating an instance snapshot automatically creates a corresponding snapshot on the dependent volume.
- Deleting an instance snapshot removes the corresponding snapshot on the dependent volume.
- Direct snapshot creation or deletion on a dependent volume is not allowed.

Additionally, exporting an instance automatically includes its dependent volumes.

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: LXD\n"
 "Report-Msgid-Bugs-To: lxc-devel@lists.linuxcontainers.org\n"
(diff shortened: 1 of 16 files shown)


stgraber:
@presztak can you do a quick rebase?

presztak:
@stgraber Done.

PR #3029, opened 2026-03-13
Add FreeBSD agent

bensmrs:
The aim of this PR is to have feature-parity with the Darwin agent.
Its implementation turned out to be a mix of things from the Linux and from the Darwin agents, with most of the actual work being in the loader.

It should easily be portable to other popular BSDs, which will be looked into in future PRs.

### DIFF
--- a/doc/.wordlist.txt
+++ b/doc/.wordlist.txt
@@ -93,6 +93,7 @@ ETag
 failover
 formatters
 FQDNs
+FreeBSD
 frontend
 Furo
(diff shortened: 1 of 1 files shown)


bensmrs:
Sorry to say it, but we now need to package FreeBSD in the public image server, @stgraber  :)

bensmrs:
Hmmm the shellcheck warnings aren’t really justified, that’s just how `rc.d` works…

PR #3050, opened 2026-03-18
Simplify config subvolume naming by using static "instance" prefix

presztak:
This PR updates the naming scheme for Btrfs subvolumes used as VM config volumes.

Previously, subvolumes were named using the instance name:
  - `<instance-name>`
  - `<instance-name>-snap0`

This PR replaces that with a static naming scheme:
  - `instance`
  - `instance-snap0`

By removing the dependency on the instance name, this simplifies a number of operations.
In particular, renaming an instance no longer requires renaming associated Btrfs subvolumes,
which reduces complexity and avoids unnecessary filesystem operations.

### DIFF
--- a/cmd/incusd/patches.go
+++ b/cmd/incusd/patches.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"time"
 
+	internalInstance "github.com/lxc/incus/v6/internal/instance"
 	"github.com/lxc/incus/v6/internal/server/backup"
 	"github.com/lxc/incus/v6/internal/server/certificate"
 	"github.com/lxc/incus/v6/internal/server/cluster"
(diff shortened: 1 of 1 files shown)


cmd/incusd/patches.go
```diff
@@ -1650,4 +1652,102 @@ func patchDefaultStoragePermissions(_ string, d *Daemon) error {
...
+						return fmt.Errorf("Failed to rename %q to %q: %w", oldPath, newPath, err)
+					}
+				}
+			}
+
+			return nil
+		})
+		if err != nil {
+			return err
+		}
```
stgraber, line 1746: @presztak we said we'd make this non-fatal, just log an error as otherwise if there are any broken volumes which need a config rebuild, it will completely break the update with no good way out of the user.

stgraber:
Looks good overall, just need to make the patch be non-fatal on conversion failure to avoid potentially deadlocking upgrades when a machine has some VMs with corrupted config volumes.

stgraber:
Those can be fixed later using the instance repair API which will re-build the volume using the new layout.

cmd/incusd/patches.go
presztak: Fixed. Also in case of other errors we just skip to the next instance.

Issue #3040, opened 2026-03-16
Broken output when --sub-commands option is given


urbalazs:
### Is there an existing issue for this?

- [x] There is no existing issue for this bug

### Is this happening on an up to date version of Incus?

- [x] This is happening on a supported version of Incus

### Incus system details

```yaml
incus v6.22
```

### Current behavior

Execute any `incus` command with `--help` option shows the global flags, too. For example `incus list --help`:

```
Global Flags:
      --debug          Show all debug messages
      --force-local    Force using the local unix socket
  -h, --help           Print help
      --project        Override the source project
  -q, --quiet          Don't show progress information
      --sub-commands   Use with help or --help to view sub-commands
  -v, --verbose        Show all information messages
      --version        Print version number
```

Addig the `--sub-commands` adds an extra indentation to the first line. For example: `incus list --help --sub-commands`:

```
Global Flags:
        --debug          Show all debug messages
      --force-local    Force using the local unix socket
  -h, --help           Print help
      --project        Override the source project
  -q, --quiet          Don't show progress information
      --sub-commands   Use with help or --help to view sub-commands
  -v, --verbose        Show all information messages
      --version        Print version number
```

Nice to have translations support for this tool, but please note that the section headers (Usage, Aliases, Examples, Available Commands, Flags, Global Flags, Additional help topics) are hard-coded in [cmd/incus/main.go](https://github.com/lxc/incus/blob/main/cmd/incus/main.go#L48-L75)
Please mark these headers as translatable while fixing the bug.

### Expected behavior

All output should be displayed with proper layout and localized, if translations are available.

### Steps to reproduce

1. Execute any `incus` command with `--help` option and see the _Global Flags_ section.
2. Execute the same command with `--help --sub-commands` option and see the _Global Flags_ section. The first line has an extra indentation.

PR #3058, opened 2026-03-20
Add colors to the CLI client

bensmrs:
I believe the changes don’t lead to cognitive overload, while still giving a nice pop to some elements that could benefit from it, helping to navigate through long command descriptions.
This PR also incidentally allows translating command usage sections.

Related-to: #2877
Related-to: #3040


bensmrs:
I’ve added a small golangci exception to treat colored printers the same as `fmt` printers, as they are just wrappers with the same signatures.

bensmrs:
FormatSection also needs some patching, it seems

bensmrs:
Phew, that proved a bit more painful than expected.

breml:
@bensmrs I am a bit late on this one, but it might be worth mentioning in the docs / README, that colorized output can be disabled using `NO_COLOR` env, see https://github.com/fatih/color?tab=readme-ov-file#disableenable-color

CC: @stgraber

bensmrs:
Ok, will do. I’ll also add a key in the client config to disable it.

bensmrs:
I’ve got all the bits working, but I’m wondering a few things, @stgraber.
First, do we have a dedicated place in the docs to mention all the client config parameters? I didn’t find one, so I’m tempted to hide that under the rug for now.
Other thing: we have `default-remote` at the top level, but `list_format`, `console_type` and `console_spice_command` at the second level. Should I use dashes or underscores for config keys? As my key is top-level, I’m tempted to choose `show-colors` for now, but we’re not consistent across the codebase.

stgraber:
> I’ve got all the bits working, but I’m wondering a few things, @stgraber. First, do we have a dedicated place in the docs to mention all the client config parameters? I didn’t find one, so I’m tempted to hide that under the rug for now. Other thing: we have `default-remote` at the top level, but `list_format`, `console_type` and `console_spice_command` at the second level. Should I use dashes or underscores for config keys? As my key is top-level, I’m tempted to choose `show-colors` for now, but we’re not consistent across the codebase.

Ah yeah, the `defaults` section is somewhat new. If we had introduced it from the start, we may have made `remote` be stored under there rather than as its own top-level option.

I think for the color handling, it'd make sense to put that under `defaults` especially if it's something that we may want to allow overriding some other ways (say `NO_COLOR=0` or an eventual `--color` global option).

And yeah, the CLI configuration doesn't really have any documentation, which is a bit of a problem on its own and something we should probably be addressing ;)
It's mentioned in the context of the remote documentation, but as it stores more than that these days, it probably deserves its own page which the remote doc could then link to instead.

bensmrs:
Ok, I’ll patch the PR after my evening context switch

PR #3065, opened 2026-03-23
Few tweaks around CLI colors

bensmrs:
Related-to: #3058

### DIFF
--- a/shared/cliconfig/default.go
+++ b/shared/cliconfig/default.go
@@ -40,6 +40,9 @@ type DefaultSettings struct {
 
 	// Alternative SPICE command (SOCKET will be replaced by the socket path).
 	ConsoleSpiceCommand string `yaml:"console_spice_command"`
+
+	// NoColor disables colors in the client.
+	NoColor bool `yaml:"no_color"`
 }
 
 // DefaultConfig returns the default configuration.
--- a/cmd/incus/color/color.go
+++ b/cmd/incus/color/color.go
@@ -12,14 +12,32 @@ func commandHeader(header string) string {
 
 // A few prefixes used throughout the Incus client.
 var (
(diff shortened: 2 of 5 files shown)


stgraber:
As mentioned in #3058 we'll want to move the flag to the Defaults section.
Otherwise this is looking good!

shared/cliconfig/default.go
```diff
@@ -40,6 +40,9 @@ type DefaultSettings struct {
 
 	// Alternative SPICE command (SOCKET will be replaced by the socket path).
 	ConsoleSpiceCommand string `yaml:"console_spice_command"`
+
+	// ShowColors controls whether colors are shown in the client.
+	ShowColors string `yaml:"show_colors"`
```
stgraber: The way it's worded, I'd expect it to be a boolean, but that would then default to False which isn't really what we want here :)

Alternatives could be `Colors` which would be a string value that can be used to change the color scheme or turn them off entirely. Or `NoColors` which would then be a boolean in line with the behavior of NO_COLORS in the environment.

bensmrs: I’m not particularly in the mood to implement a color scheme manager :)
Regarding the fact that it’s a not a boolean, that was exactly the line of reasoning behind it. Additionally, YAML booleans do not conform 1:1 to what Incus parses as booleans (`0` and `1` don’t seem to get a special downcasting), but we can disregard that.

So, `` NoColors bool `yaml:"no_colors"` `` then?

stgraber: sounds good to me

bensmrs: Well, the env var is `NO_COLOR` and the related movement is https://no-color.org, so let’s go singular.

stgraber (approved):

PR #3072, opened 2026-03-24
incus-migrate: Fix OVA handling within os.Root

stgraber:


stgraber:
@bensmrs that should take care of the issue you noticed

bensmrs:
That does the trick, thanks

PR #3086, opened 2026-03-25
Add cp-like flags to pull commands

### LINKED ISSUE #3017: incus file pull no way to pull physical file of an instance symlink

I have it described here - https://discuss.linuxcontainers.org/t/incus-file-pull-not-pulling-file-behind-a-symlink/26337/2

In lxc you could pull a physical file behind a symlink with the **-r** the -r flag doesn't seem to do that in incus.
I understand not wanting to overload the recurse flag with pulling a physical file, but it would be nice to have a switch to allow that so I'm putting this in as a feature request instead of a bug.

Use case from my discuss post:

I have a script that pulls letsencrypt certs from one instance to another.

Note in the nginx container - this is a symlink - /etc/letsencrypt/live/site-name/fullchain.pem to a file in archive folder

In lxd, this worked and brought over the physical file:

`lxc file pull nginx/etc/letsencrypt/live/site-name/fullchain.pem fullchain.pem.new -r`

In incus, it pulls over the symlink instead of the physical file:

`incus file pull nginx/etc/letsencrypt/live/site-name/fullchain.pem fullchain.pem.new -r`

 I’m running incus 6.22

Addition of an extra switch to lxc file pull bring over physical file.

bensmrs:
Closes: #3017
This also ensures files are properly closed in `pull` methods.


bensmrs:
I’m having trouble running `storage_volume_filemanip` locally: `Error: Failed connecting to instance SFTP: Failed getting storage volume SFTP connection: Failed to run forkfile: fork/exec /root/go/bin/incusd: permission denied:`. Rings any bell, @stgraber?

stgraber:
Yeah, we probably run it with reduced privileges so you need to chmod +x all folders leading to the binary so unprivileged users can run incusd

stgraber:
chmod +x /root /root/go /root/go/bin should do it

bensmrs:
Yep, that did it! thanks

bensmrs:
This work needs some follow-up to implement a similar logic for `push` commands, but I probably won’t write it in time for this release.

PR #3091, opened 2026-03-26
Add `image.requirements.cdrom_cloud_init` key

bensmrs:


bensmrs:
@stgraber I’m not seeing any API extension for `image.requirements.cdrom_agent`, so I didn’t define one here. Should I?

stgraber:
> @stgraber I’m not seeing any API extension for `image.requirements.cdrom_agent`, so I didn’t define one here. Should I?

I think we didn't do one last time because this is only borderline user API.
It's primarily API between us and the image server more than us and the user.
There's also no real use for a client to know whether we will show the error or not.

PR #3092, opened 2026-03-26
Fix some security issues

stgraber:


internal/server/instance/drivers/driver_lxc.go
```diff
@@ -9520,16 +9520,22 @@ func (d *lxc) setupCredentials(update bool) error {
 		}
 	}
 
+	credsRoot, err := os.OpenRoot(credentialsDir)
```
gibmat, line 9523: The commit message is missing the full CVE identifier. :)

stgraber: oops, let me go and fix that one :)

stgraber: Fixed!
I think I fixed it on a different computer yesterday and forgot to push...

PR #3093, opened 2026-03-26
Add a NBD based backup API

stgraber:
This implements some of the first bits for a NBD based backup API. I'll now hand the rest over to @presztak to complete as I need to focus on other things right now.

@presztak what we have in this PR is:
 - POST /1.0/storage-pools/POOL/volumes/TYPE/VOLNAME which upgrades to an NBD socket
 - All the needed plumbing to handle both custom and instance volumes
 - All the expected bits at the storage level to handle the request
 - The needed bits in the instance layer to handle this against running QEMU
 - Client support to connect to this
 - CLI support to forward it onto a local server

With this, I can do:

image

The bits that are missing for a full implementation:

- [x] Better errors when an NBD server is already running (doesn't look like QEMU lets us have multiple ones...).
- [x] Support for this in offline mode (`qemu-nbd` running under `ActivateTask`, we'll need to implement `ActivateTask` for the other storage drivers). This will allow for writable NBD too.
- [x] When in online (running) mode, the NBD exported block should be a snapshot so we can have something consistent for reading.
- [x] Implement a new API at /1.0/storage-pools/POOL/volumes/TYPE/VOLNAME/bitmaps to allow for the creation, listing and deleting of dirty bitmaps. This will always be available with QCOW2 storage and only be available when the instance is running otherwise. Given the bitmaps have names, we'll want a standard API with individual entries under /1.0/storage-pools/POOL/volumes/TYPE/VOLNAME/bitmaps/BITMAP.
- [x] Implement a POST-only API under /1.0/instances/NAME/bitmaps to trigger the creation of a new bitmap concurrently on the instance and all dependent volumes concurrently. We don't need to have the GET/DELETE support on this one as that can be done through the volume API.

### DIFF
--- a/internal/server/instance/drivers/driver_lxc.go
+++ b/internal/server/instance/drivers/driver_lxc.go
@@ -9606,22 +9606,22 @@ func (d *lxc) GuestOS() string {
 
 // CreateQcow2Snapshot creates a qcow2 snapshot for a running instance. Not supported by containers.
 func (d *lxc) CreateQcow2Snapshot(devPath string, devName string, snapName string, backingFilename string, stateful bool) error {
-	return nil
+	return instance.ErrNotImplemented
 }
 
 // DeleteQcow2Snapshot deletes a qcow2 snapshot for a running instance. Not supported by containers.
 func (d *lxc) DeleteQcow2Snapshot(devName string, snapshotIndex int, backingFilename string) error {
-	return nil
+	return instance.ErrNotImplemented
 }
 
 // ExportQcow2Block exports a qcow2 block device. Not supported by containers.
 func (d *lxc) ExportQcow2Block(diskName string, diskIndex int) (func(), string, error) {
-	return nil, "", nil
+	return nil, "", instance.ErrNotImplemented
 }
 
 // ConnectNBD exports a disk over NBD. Not supported by containers.
 func (d *lxc) ConnectNBD(diskName string, volSize int64, writable bool) (net.Conn, func(), error) {
-	return nil, nil, nil
+	return nil, nil, instance.ErrNotImplemented
 }
 
 // setNICLink sets the link status of the given device.
@@ -9653,15 +9653,15 @@ func (d *lxc) setNICLink(devName string, connected bool, assumeUp bool) error {
 
 // CreateBitmap creates a dirty bitmap. Not supported by containers.
 func (d *lxc) CreateBitmap(deviceNames []string, data api.StorageVolumeBitmapsPost) error {
-	return nil
+	return instance.ErrNotImplemented
 }
 
 // DeleteBitmap deletes a dirty bitmap. Not supported by containers.
 func (d *lxc) DeleteBitmap(deviceName string, bitmapName string) error {
-	return nil
+	return instance.ErrNotImplemented
 }
 
 // GetBitmaps fetches dirty bitmaps. Not supported by containers.
 func (d *lxc) GetBitmaps(deviceName string) ([]api.StorageVolumeBitmap, error) {
-	return nil, nil
+	return nil, instance.ErrNotImplemented
 }


internal/server/storage/backend.go
```diff
@@ -9159,12 +9157,26 @@ func (b *backend) GetInstanceNBD(inst instance.Instance, writable bool) (net.Con
...
-		return nil, nil, err
+	if !inst.IsRunning() {
+		b.logger.Debug("NBD connection (offline mode)")
+		return b.connectOfflineNBD(vol)
 	}
 
-	volSize, err := drivers.BlockDiskSizeBytes(volDiskPath)
+	var volSize int64
+	var volDiskPath string
+	err = b.Driver().ActivateTask(vol, func(devPath string, op *operations.Operation) error {
```
stgraber: I'm confused by this one.

ActivateTask is designed to be exclusive, but if the VM is running, then there are rather high changes that the block is already active, so the ActivateTask here would fail due to existing activation.

internal/server/storage/backend.go
```diff
@@ -9235,3 +9249,69 @@ func (b *backend) GetCustomVolumeNBD(projectName string, volName string, writabl
 
 	return inst.ConnectNBD(instanceDeviceName, volSize, writable)
 }
+
+// connectOfflineNBD spawns qemu-nbd for the given volume.
+func (b *backend) connectOfflineNBD(vol drivers.Volume) (net.Conn, func(), error) {
+	socketPath := filepath.Join("/tmp", fmt.Sprintf("%s-nbd.sock", vol.Name()))
```
stgraber: We should use `internalUtil.RunPath()` for that so it's under a protected /run/incus/ path rather than the world readable /tmp

cmd/incusd/instance_bitmap.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	inst, err := instance.LoadByProjectAndName(s, projectName, cname)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	if !inst.IsRunning() {
+		return response.BadRequest(fmt.Errorf("Creating bitmaps requires the instance to be running"))
+	}
```
stgraber, line 83: We should also check that we're dealing with a VM here.

cmd/incusd/instance_bitmap.go
```diff
@@ -0,0 +1,148 @@
...
+	// Load storage volume from database.
+	dbVol, err := storagePools.VolumeDBGet(pool, inst.Project().Name, inst.Name(), volType)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	if dbVol.Config["block.type"] != storageDrivers.BlockVolumeTypeQcow2 {
+		return response.BadRequest(fmt.Errorf("Only QCOW2 volumes are supported"))
+	}
+
```
stgraber: Hmm, that shouldn't be limited to QCOW2, dirty bitmaps should still work on other backends, just not be persistent.

cmd/incusd/storage_volumes_bitmap.go
```diff
@@ -0,0 +1,520 @@
...
+	// Convert the volume type name to our internal integer representation.
+	volumeDBType, err := storagePools.VolumeTypeNameToDBType(volumeTypeName)
+	if err != nil {
+		return response.BadRequest(err)
+	}
+
+	// Check that the storage volume type is valid.
+	if !slices.Contains([]int{db.StoragePoolVolumeTypeVM, db.StoragePoolVolumeTypeCustom}, volumeDBType) {
+		return response.BadRequest(fmt.Errorf("Unsupported storage volume type %q", volumeTypeName))
+	}
```
stgraber, line 109: We should also check that if TypeCustom that the contentType is Block

cmd/incusd/storage_volumes_bitmap.go
```diff
@@ -0,0 +1,520 @@
...
```
stgraber, line 226: Same note about validation as above

cmd/incusd/storage_volumes_bitmap.go
```diff
@@ -0,0 +1,520 @@
...
```
stgraber, line 361: And here too :)

cmd/incusd/storage_volumes_bitmap.go
```diff
@@ -0,0 +1,520 @@
...
```
stgraber, line 482: Same here

cmd/incusd/storage_volumes_bitmap.go
```diff
@@ -0,0 +1,520 @@
...
+
+	err = inst.DeleteBitmap(deviceName, bitmapName)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	return response.EmptySyncResponse
+}
+
+func instanceByVolumeName(s *state.State, poolName string, projectName string, volumeName string, volumeDBType int) (instance.Instance, string, error) {
```
stgraber: You may be able to use that function with the NBD functions too as I believe I was doing something similar in there.

stgraber:
@presztak great work!
I left a few comments and notes but hopefully nothing too complex to sort out!

stgraber:
@bensmrs if you have a bit of spare time somehow, it'd be great to also have you take a quick look at this one, mostly to have an extra neutral pair of eyes on this one given that I'm the author for half of it and Piotr is the author for the other half which makes us reviewing each other's commit in the same PR a bit funny ;)

bensmrs:
Gladly, but only from Monday on, as I’m fully taken until then.

stgraber:
> Gladly, but only from Monday on, as I’m fully taken until then.

Sounds good! We may have it merged before then because we have a backup vendor who's waiting to have those APIs available in the daily builds, but a review after the fact will still allow us to fix anything problematic before the next release :)

internal/server/storage/backend.go
presztak: Ahh, it’s supposed to be `MountTask`. Not sure how I was able to validate it successfully on my side. Let me double check.

cmd/incusd/instance_bitmap.go
presztak: You’re absolutely right. I based my suggestion on your description: " This will always be available with QCOW2 storage and only be available when the instance is running otherwise. ". Let me fix that.

internal/server/storage/backend.go
presztak: Fixed.

internal/server/storage/backend.go
presztak: Fixed.

cmd/incusd/instance_bitmap.go
presztak: Done.

cmd/incusd/instance_bitmap.go
presztak: Check removed.

cmd/incusd/storage_volumes_bitmap.go
presztak: Done.

cmd/incusd/storage_volumes_bitmap.go
presztak: Done.

cmd/incusd/storage_volumes_bitmap.go
presztak: Fixed.

cmd/incusd/storage_volumes_bitmap.go
presztak: Fixed.

cmd/incusd/storage_volumes_bitmap.go
presztak: Done.

bensmrs:
I’ll finally be able to have a look today, I’ll start in a bit.

cmd/incusd/instance_bitmap.go
```diff
@@ -0,0 +1,116 @@
...
+	cname, err := url.PathUnescape(mux.Vars(r)["name"])
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	if internalInstance.IsSnapshot(cname) {
+		return response.BadRequest(errors.New("Invalid instance name"))
+	}
+
+	// Handle requests targeted to a container on a different node
```
bensmrs:
```suggestion
	// Handle requests targeted at a VM on a different node
```

internal/server/instance/drivers/driver_lxc.go
```diff
@@ -9619,6 +9619,11 @@ func (d *lxc) ExportQcow2Block(diskName string, diskIndex int) (func(), string,
 	return nil, "", nil
 }
 
+// ConnectNBD exports a disk over NBD. Not supported by containers.
+func (d *lxc) ConnectNBD(diskName string, volSize int64, writable bool) (net.Conn, func(), error) {
+	return nil, nil, nil
```
bensmrs: We technically could even return an error here.

internal/server/instance/drivers/driver_lxc.go
```diff
@@ -9645,3 +9650,18 @@ func (d *lxc) setNICLink(devName string, connected bool, assumeUp bool) error {
 
 	return nil
 }
+
+// CreateBitmap creates a dirty bitmap. Not supported by containers.
+func (d *lxc) CreateBitmap(deviceNames []string, data api.StorageVolumeBitmapsPost) error {
+	return nil
```
bensmrs: Same as above.

internal/server/instance/drivers/driver_lxc.go
```diff
@@ -9645,3 +9650,18 @@ func (d *lxc) setNICLink(devName string, connected bool, assumeUp bool) error {
...
+}
+
+// DeleteBitmap deletes a dirty bitmap. Not supported by containers.
+func (d *lxc) DeleteBitmap(deviceName string, bitmapName string) error {
+	return nil
```
bensmrs: Same as above.

internal/server/instance/drivers/driver_lxc.go
```diff
@@ -9645,3 +9650,18 @@ func (d *lxc) setNICLink(devName string, connected bool, assumeUp bool) error {
...
+}
+
+// GetBitmaps fetches dirty bitmaps. Not supported by containers.
+func (d *lxc) GetBitmaps(deviceName string) ([]api.StorageVolumeBitmap, error) {
+	return nil, nil
```
bensmrs: Same as above.

internal/server/storage/drivers/driver_linstor_volumes.go
```diff
@@ -643,6 +643,25 @@ func (d *linstor) ListVolumes() ([]Volume, error) {
...
+	unlock, err := vol.MountLock()
+	if err != nil {
+		return err
+	}
+
+	defer unlock()
+
+	volDevPath, err := d.getLinstorDevPath(vol)
+	if err != nil {
+		return fmt.Errorf("Could not mount volume: %w", err)
```
bensmrs: We’re technically not mounting anything.

shared/api/storage_pool_volume_bitmap.go
```diff
@@ -0,0 +1,59 @@
...
+//
+// swagger:model
+//
+// API extension: storage_volume_nbd.
+type StorageVolumeBitmap struct {
+	// Bitmap name
+	// Example: bitmap0
+	Name string `json:"name" yaml:"name"`
+
+	// Number of dirty bytes
```
bensmrs, line 13: Bytes or bits?

internal/server/instance/drivers/driver_lxc.go
stgraber: Yeah, the current pattern for functions that don't make sense for containers is to return nil but I agree we probably should put a following commit that changes them all to returning errors.New("Not implemented") so we don't risk messing with a nil pointer.

stgraber:
> I have a few nits, but it’s really nothing. I’m wondering if we shouldn’t add an NBD upgrade endpoint for instances also, but I understand if we want to keep a single API path.

Yeah, I think I had that in my initial design but then went for pushing as much as possible inside the storage API instead as it was around the time we were landing dependent volumes and having the instance /nbd not provide a way to access the dependent volumes felt a bit weird.

shared/api/storage_pool_volume_bitmap.go
stgraber: Bytes

stgraber: (I went to recheck the matching QMP struct)

PR #3098, opened 2026-03-27
github: format INCUS_VERSION from tag in release workflow

0xk1f0:
This PR adjusts the `workflows/release.yaml` to not use a hardcoded version tag and instead get it directly from the tag the action is triggered from.

Outputs:
- `MAJOR.MINOR` if `PATCH` is zero
- `MAJOR.MINOR.PATCH` if `PATCH` is non-zero


stgraber:
Thanks!

I had been doing testing using 6.22 in my own repository and forgot to replace that when pushing it to main...

Issue #1301, opened 2024-10-11
Incus from source on Rocky Linux issues with spice and qemu-kvm


lebonez:
# Required information

 * Distribution: Rocky Linux
 * Distribution version: 9.4
 * The output of "incus info" or if that fails: 
   * Kernel version: 5.14.0-427.37.1.el9_4.x86_64
   * LXC version: 
   * Incus version: 6.6
   * Storage backend in use: Ceph

# Issue description

High likelihood this could be a duplicate, I've built Incus on Rocky Linux 9.4 from source. I am hitting two major issues. It seems that Incus is expecting qemu-system-x86_64 to exist on RedHat derivatives but this does not exist it exists at /usr/libexec/qemu-kvm. Doing a symlink from `ln -s /usr/libexec/qemu-kvm /usr/bin/qemu-system-x86_64` works around it but that isn't ideal. The next major issue is the drop of Spice support in RedHat 9. Is it possible to build incus without expecting spice? 

# Steps to reproduce

 1. Install Rocky Linux 9.4
 2. Build from source
 3. Attempt to run a VM like `lxc launch images:rockylinux/9/cloud --vm` VM feature won't be supported even though it should be"
 4. `ln -s /usr/libexec/qemu-kvm /usr/bin/qemu-system-x86_64`
 5. `lxc launch images:rockylinux/9/cloud --vm` now it will complain about the VM configs due to the spice settings not being part of qemu anymore in RedHat 9.

# Questions

* Is it possible in the future that incus will support the /usr/libexec/qemu-kvm location?
* Is it also possible that incus will dynamically swap away from spice as needed if it doesn't exist?

We are definitely awaiting a RedHat 9 zabbly incus LTS.

stgraber:
1) Should be possible though that's not particularly great as we normally want to use the architecture specific QEMU binaries. How does RHEL handle the case where you have the QEMU system binaries for multiple architectures installed? Is qemu-kvm some kind of shell script or the actual QEMU binary? 

2) We may be able to detect the lack of SPICE support and degrade things, though that means no VGA access for your VMs which will make some use cases pretty unpleasant (like running Windows or a Linux desktop VM).

lebonez:
Thank you for the quick response.

1) Indeed /usr/libexec/qemu-kvm on redhat is just the qemu binary unsure if it differentiates on architectures. Looking at the qemu packages available on EL9 there are no extra architecture packages like there used to be on EL8 and older.

[...]

PR #3099, opened 2026-03-27
feat: feature gate spice behind security.spice flag

damex:
here we introduce security.spice flag which defauls to true
  the same way it was working before this introduction.
setting security.spice = false makes those spice devices along
  with -spice command line flag absent.

### DIFF
--- a/internal/instance/config.go
+++ b/internal/instance/config.go
@@ -1137,6 +1137,16 @@ var InstanceConfigKeysVM = map[string]func(value string) error{
 	//  shortdesc: Whether to use a firmware that supports UEFI-incompatible operating systems
 	"security.csm": validate.Optional(validate.IsBool),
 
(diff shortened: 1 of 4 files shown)


damex:
i tried to match current convention across the repo and used security.spice since other 'non-security' features are also in 'security namespace'. please let me know if i should change it to something else or this is fine.

stgraber:
If we were going to be fine with a config option, we would have added it immediately when processing #1301 months ago. The reason we haven't is because unlike other platforms like libvirt, we like to be able to abstract this kind of stuff from the user and don't like those kind of config options.

In theory you should be able to use `raw.qemu.scriptlet` as a way to rewrite the QEMU command line. But our preference for this one would be to have the lack of SPICE support be detected by `checkFeatures` and then issue a persistent warning and turn it off for the VM.

PR #3101, opened 2026-03-27
incus-simplestreams: Support using name from manifest as product name

csprl:
This would be very useful to be able to build a variety of images on the same distro and arch with different names, rather than just using variant to separate them.

### DIFF
--- a/cmd/incus-simplestreams/main_add.go
+++ b/cmd/incus-simplestreams/main_add.go
@@ -27,8 +27,9 @@ import (
 type cmdAdd struct {
 	global *cmdGlobal
 
-	flagAliases        []string
-	flagNoDefaultAlias bool
+	flagAliases             []string
+	flagNoDefaultAlias      bool
+	flagUseNameFromMetadata bool
 }
 
 // Command generates the command definition.
@@ -48,6 +49,7 @@ This command parses the metadata tarball to retrieve the following fields from i
  - properties["release"]
  - properties["variant"]
  - properties["architecture"]
+ - properties["name"] (if --use-metadata-name is specified)
 
 It then check computes the hash for the new image, confirm it's not
 already on the image server and finally adds it to the index.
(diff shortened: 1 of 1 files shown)


stgraber:
@csprl given that `name` isn't a normal property created by `generate-metadata` or `incus publish`, I'm not sure why this uses that property rather than just add a `--product-name` flag to override the default product name with your specific value.

csprl:
> @csprl given that `name` isn't a normal property created by `generate-metadata` or `incus publish`, I'm not sure why this uses that property rather than just add a `--product-name` flag to override the default product name with your specific value.

It's used and produced by distrobuilder, hence the usage.

stgraber:
Ah, interesting, I forgot that we had that as we never actually use it in the public image server publishing pipeline.

stgraber:
But then the value generated by distrobuilder is unsuitable for that field as it contains the image version whereas what you need here is a product identifier, so excluding any version number.

An example out of distrobuilder is `alpinelinux-edge-x86_64-default-20260326_13:00` but a suitable product identifier for it would be `alpine:edge:amd64:default`.

stgraber:
So I'm kind of back to, it would make more sense to have an argument to import an image into a specific product and so have a `--product-name` argument which could then be passed the expected `alpine:edge:amd64:default` or whatever the case may be for your particular image server.

Though that particular value is basically what we generate today from the other fields and is in line with general simplestreams product patterns.

csprl:
The point is that `name` is easily overridden to _anything_ you want during the build process, which makes it very flexible. Setting the value with an argument is maybe workable, but it would be preferrable to be able to set it during build for more flexibility in deploying from CI (separate build and deployment steps).

My use case here is building (and hosting) images that will be used in CI to build things, and so I have several images based on the same distro, release and architecture. This leaves only the `variant` field for separating them, which isn't flexible enough. Since it seems like you also can't set aliases on specific versions of images, only on the entire product, I actually _would_ like to include a version in the image name itself.

Hope this clears things up.

stgraber:
Your situation seems unusual, we definitely don't want the name property as generated by distrobuilder to ever be used as the product in simplestreams and the pattern you should be using out of distrobuilder as the product name is the one we already generate.

So we definitely won't be using the "name" property for that since the only tool that publicly makes use of it right now sets it by default to an unsuitable value. 

We could read from another new property line "product_id" or something like that though, but it feels kinda odd having what's basically a simplestreams concept (stream product id) be exposed as part of the image when an image could be part of multiple simplestreams products or could be in different products on different simplestreams servers. 

Definitely sounds to me like this shouldn't be pulled from the image at all but be a command line flag.

csprl:
I take your point, I'm sure I can make a command line argument work for my use-case. Will open a new MR.

PR #3107, opened 2026-03-28
Cleanup cmd packages

stgraber:


cmd/incus/alias.go
bensmrs, line 1: Nit: maybe replace `// It <...>` by `// This <...>` or `// This command <...>` or `// This function <...>`.

cmd/incus/config.go
```diff
@@ -374,9 +371,8 @@ type cmdConfigGet struct {
 
 var cmdConfigGetUsage = u.Usage{u.MakePath(u.Instance, u.Snapshot.Optional()).Optional().Remote(), u.Key}
 
-// Command creates a Cobra command to fetch values for given instance or server configuration keys,
 // with optional flags for expanded configuration and cluster targeting.
```
bensmrs: This should probably go.

cmd/incus/alias.go
stgraber: Just removed the comments from this file instead.

The ones that were purely copy/paste placeholders have all been sed-ed away, but the remaining stuff in this file was just bespoke enough not to get caught while still not really bringing in any value.

It makes sense to keep the comment if it describes something unexpected/unusual from one of those functions, but here everything is pretty standard.

cmd/incus/config.go
stgraber: Yup, done

stgraber:
@bensmrs I think we should be good now.
I had a quick look for any stragglers and didn't find anything off now.

Basically went through:
```
stgraber@castiana:~/Code/lxc/incus (git:stgraber/cli)$ git diff lxc/main.. cmd/incus | grep ^-func -B1 | grep '^ /'
 // It handles actions on instances (single or all) and manages error handling, console flag restrictions, and batch operations.
 // instances with their CPU, memory and disk usage columns.
```

To look for any comment being left untouched immediately preceding one of the changes.
The two that are left make sense.

PR #3110, opened 2026-03-28
incusd/instance/qemu: Enable SPICE audio feature gating

damex:


stgraber:
virtio-sound isn't necessarily spice related, you may need to have a sound card exposed to your guest even if spice isn't available.

damex:
> virtio-sound isn't necessarily spice related, you may need to have a sound card exposed to your guest even if spice isn't available.

but here it specifically does spice variant. should we fall it back to something else instead of omitting? having any spice device without spice support breaks vm

PR #3121, opened 2026-03-29
Rework storage caching logic (performance improvement)

stgraber:


internal/server/storage/drivers/driver_zfs_cache.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+
+		record, ok := zfsCache[fields[0]]
+		if !ok {
+			record = map[string]zfsCacheEntry{}
+		}
+
+		for i, value := range fields[1:] {
+			key := zfsCacheProperties[i]
+			record[key] = zfsCacheEntry{Expiry: time.Now().Add(15 * time.Second), Value: value}
```
bensmrs: The expiry feels a bit magical. Couldn’t it benefit from a dedicated configuration key?

internal/server/storage/drivers/driver_zfs_cache.go
```diff
@@ -0,0 +1,202 @@
...
+	queue := []string{}
+	for _, entry := range zfsCachePrefillQueue {
+		record, ok := zfsCache[entry]
+		if ok {
+			valid := true
+
+			for _, propName := range zfsCacheProperties {
+				prop, ok := record[propName]
+				if !ok || prop.Expiry.Before(time.Now()) {
+					valid = false
```
bensmrs: You could also break out of the loop here

internal/server/storage/drivers/driver_zfs_cache.go
```diff
@@ -0,0 +1,202 @@
...
+
+	"github.com/lxc/incus/v6/shared/logger"
+	"github.com/lxc/incus/v6/shared/subprocess"
+)
+
+// This is a global ZFS cache for the system, used to limit the number
+// of expensive requests to ZFS especially during bulk requests like a full
+// instance list.
+type zfsCacheEntry struct {
+	Expiry time.Time
```
bensmrs, line 17: Shouldn’t the expiry be on records rather than entries?

internal/server/storage/drivers/driver_zfs_cache.go
```diff
@@ -0,0 +1,202 @@
...
+
+	// Check if we already have a valid cache for the dataset.
+	record, ok := zfsCache[dataset]
+	if ok {
+		valid := true
+
+		for _, propName := range zfsCacheProperties {
+			prop, ok := record[propName]
+			if !ok || prop.Expiry.Before(time.Now()) {
+				valid = false
```
bensmrs: You could break out of the loop here

bensmrs (changes requested):

internal/server/storage/drivers/driver_zfs_cache.go
stgraber: Not easily as the caching/refresh mechanism is separate from the concept of a pool at that stage, so we can't go lookup the config of a particular pool.

internal/server/storage/drivers/driver_zfs_cache.go
stgraber: As it is right now, yes. I slightly over-designed this so we can in theory send every get and set property requests through this mechanism if we see that as valuable, just with differing expires per property.

The current default of 15s on ZFS and 60s on TrueNAS is suitable for the two current properties (used/referenced) as usage info doesn't need to be crazy up to date. Some other keys are otherwise known to be hit pretty bad in specific cases (volmode during create/start comes to mind) but you don't want that to be cached for long.

internal/server/storage/drivers/driver_zfs_cache.go
stgraber: Fixed both in ZFS and TrueNAS.

internal/server/storage/drivers/driver_zfs_cache.go
stgraber: Fixed both in ZFS and TrueNAS.

internal/server/storage/drivers/driver_zfs_cache.go
stgraber: That could be a global storage.XYZ configuration, though I'm not sure that I like the sound of adding database queries to this logic. I'd say easiest to keep as things are and we can adjust if it causes problems for anyone.

internal/server/storage/drivers/driver_zfs_cache.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+
+		record, ok := zfsCache[fields[0]]
+		if !ok {
+			record = map[string]zfsCacheEntry{}
+		}
+
+		for i, value := range fields[1:] {
+			key := zfsCacheProperties[i]
+			record[key] = zfsCacheEntry{Expiry: time.Now().Add(15 * time.Second), Value: value}
```
bensmrs: Even though there are only 2 cached properties, that number could grow. As that PR is focused on performance, I think it would make sense to compute `time.Now().Add(15 * time.Second)` only once.

internal/server/storage/drivers/driver_zfs_cache.go
```diff
@@ -0,0 +1,204 @@
...
+
+	queue := []string{}
+	for _, entry := range zfsCachePrefillQueue {
+		record, ok := zfsCache[entry]
+		if ok {
+			valid := true
+
+			for _, propName := range zfsCacheProperties {
+				prop, ok := record[propName]
+				if !ok || prop.Expiry.Before(time.Now()) {
```
bensmrs: Same remark as below, `time.Now()` can be precomputed to avoid wasting a few cycles.

internal/server/storage/drivers/driver_zfs_cache.go
```diff
@@ -0,0 +1,204 @@
...
+			if !ok || prop.Expiry.Before(time.Now()) {
+				valid = false
+				break
+			}
+		}
+
+		if valid {
+			zfsCacheMu.Unlock()
+			return
+		}
```
bensmrs: That block is almost a duplicate of the one below, only differing at the end of the loop. Factoring it would improve legibility.

internal/server/storage/drivers/driver_zfs_cache.go
```diff
@@ -0,0 +1,204 @@
...
+		zfsCachePrefillWait.Wait()
+
+		// Check that we made it.
+		zfsCacheMu.Lock()
+		inQueue := slices.Contains(zfsCachePrefillQueue, dataset)
+		zfsCacheMu.Unlock()
+
+		if inQueue {
+			// We didn't make it, re-trigger.
+			d.prefillCachedProperties(dataset)
```
bensmrs, line 98: Why don’t you return here?

internal/server/storage/drivers/driver_zfs_cache.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	// Run the filler.
+	properties := strings.Join(append([]string{"name"}, zfsCacheProperties...), ",")
+	args := []string{"list", "-H", "-p", "-o", properties, "-r", "-t", "fs,vol,snap"}
+	args = append(args, queue...)
+
+	out, err := subprocess.RunCommand("zfs", args...)
+	if err != nil {
+		d.logger.Warn("Coulnd't cache ZFS properties", logger.Ctx{"err": err})
```
bensmrs:
```suggestion
		d.logger.Warn("Couldn't cache ZFS properties", logger.Ctx{"err": err})
```

stgraber:
@bensmrs addressed the review comments on the ZFS driver and applied them to their TrueNAS equivalent too.

PR #3122, opened 2026-03-29
Implement IncusOS update trigger on cluster version mismatch

stgraber:


stgraber:
@breml That should take care of the problem I ran into.

PR #3141, opened 2026-04-02
Support optimized storage for dependent volumes

presztak:


stgraber:
@presztak would make sense to send a follow-up to add a test for this (probably run as part of the ZFS tests would be sufficient).

PR #3143, opened 2026-04-03
Rework physical NIC device validation

stgraber:
This attempts to clarify what options of the physical NICs are valid when as that device type is getting a bit complex what with supporting:

 - Actual physical NIC passthrough to a VM (PCI VFIO, no configurability)
 - Network interface moved to a container (configurable but needs reverting)
 - VLAN interface on top of a physical NIC passed to a container (configurable)
 - Bridged NIC due to use of a "physical" type managed network which is backed by an unmanaged bridge (default on IncusOS)

This sorts out the conditions for the various properties to cover the first 3 cases and then sends validation through the bridged NIC validator for the 4th case.


internal/server/device/nic_physical.go
```diff
@@ -160,20 +171,29 @@ func (d *nicPhysical) validateConfig(instConf instance.ConfigReader, partialVali
...
+		// Check if the parent is a bridge.
 		isParentBridge := d.config["parent"] != "" && util.PathExists(fmt.Sprintf("/sys/class/net/%s/bridge", d.config["parent"]))
-		if isParentBridge && d.network == nil {
-			return fmt.Errorf("Parent device is a bridge, use nictype=bridged instead")
-		}
+		if isParentBridge {
+			// If parent is a bridge, ensure it's managed.
+			if d.network == nil {
+				return fmt.Errorf("Parent device is a bridge, use nictype=bridged instead")
+			}
```
bensmrs: That doesn’t seem possible given the previous lines.

stgraber: How come?

Previous lines only validate the presence of a `d.config["parent"]` not that the struct is loaded.

bensmrs: We are initializing the struct on line 156, then using its fields.

stgraber: Oh oops, I somehow missed that, so I guess the old logic was wrong, will remove that check :)

stgraber: Done!

bensmrs (approved):

PR #3153, opened 2026-04-06
incusd/instances/qemu: Skip vmcoreinfo on ppc64le

stgraber:
Apparently that QEMU device doesn't exist on ppc64le, so just skip it. Only impact should be limited memory layout info on dump.

### DIFF
--- a/internal/server/instance/drivers/driver_qemu.go
+++ b/internal/server/instance/drivers/driver_qemu.go
@@ -3897,7 +3897,9 @@ func (d *qemu) generateQemuConfig(machineDefinition string, cpuType string, cpuI
 	conf = append(conf, qemuConsole()...)
 
 	// VM core info (memory dump).
-	conf = append(conf, qemuCoreInfo()...)
+	if !slices.Contains([]int{osarch.ARCH_64BIT_POWERPC_LITTLE_ENDIAN, osarch.ARCH_64BIT_S390_BIG_ENDIAN}, d.architecture) {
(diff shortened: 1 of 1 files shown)


bensmrs:
https://github.com/qemu/qemu/blob/b6a7d06213e5d2f7d124d16418bc289c4a8a4b82/hw/misc/meson.build#L3 suggests that the condition should be broader and include all the architectures that don’t have `fw_cfg` support. `fw_cfg` is supported on q35 and virt, but I can’t find anything for IBM architectures, so I think we can exclude all of them.

stgraber:
@bensmrs I guess we could use the usual EFI check so it's effectively only enabled on x86_64 and aarch64? There's a decent change that the support will line up with other architecture in the future that way.

bensmrs:
`vmcoreinfo` seems to exist on `riscv64` and `loongarch64`, so I think a check for everything except `ppc` and `s390x` is futureproof enough. But if the EFI check works for `riscv64` (and `loongarch64`, but that’s for another time), then that’s also ok for me.

bensmrs:
Oh I just saw the actual definition of the EFI check :)
We can worry about that later I guess.
Semantically, it’s not related to whether EFI is supported or not, so the name isn’t great, though. At some point we may want to rename that, so that newcomers don’t wonder why a feature is tied to UEFI while it’s really not.

stgraber:
@bensmrs updated

bensmrs (approved):

PR #3162, opened 2026-04-09
Allow limiting storage pools to project

### LINKED ISSUE #3060: Allow limiting storage pools to projects same as networks with allowlist

I'd like to limit storage pools to projects the same as networks can be-
Currently it requires setting limits.disk.pool.name=0 ie denylist which grows quickly and is prone to "race". There should be default deny all and then an allowlist possibility which should be configurable like possible for networks.

Make storage pools allowlist based or provide support for such configuration.

Allowlist configuration should be a default possibility for all also imho.

pranav767:
Resolves: https://github.com/lxc/incus/issues/3060

This PR involves:

1. Registering a restriction `restricted.storage-pools.access`
2. Write  a new function `StoragePoolAllowed` similar to `NetworkAllowed` which checks if access is allowed to a particular storage pool.
3. Update existing permissions, `HiddenStoragePools` now does a additional check if the pool is present/absent in access list (Also as a case where root is always allowed we still need to check with storage pool allowlist)
4. Update use of `HiddenStoragePools` while loading project limits in `storagePoolGet` & `storagePoolsGet` as these are ones project-restricted users can access.

### DIFF
--- a/cmd/incusd/api_project.go
+++ b/cmd/incusd/api_project.go
@@ -1815,6 +1815,14 @@ func projectValidateConfig(s *state.State, config map[string]string) error {
 		//  defaultdesc: `block`
 		//  shortdesc: Whether to prevent creating instance or volume snapshots
 		"restricted.snapshots": isEitherAllowOrBlock,
+
+		// gendoc:generate(entity=project, group=restricted, key=restricted.storage-pools.access)
+		// Specify a comma-delimited list of storage pool names that are allowed for use in this project.
+		// If this option is not set, all storage pools are accessible.
+		// ---
+		//  type: string
+		//  shortdesc: Which storage pool names are allowed for use in this project
+		"restricted.storage-pools.access": validate.Optional(validate.IsListOf(validate.IsAny)),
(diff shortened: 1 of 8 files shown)


stgraber:
Looking into this one now. I'll probably make a few edits as I'm not sure why we need so much extra logic when we may be able to instead treat anything not listed in the config as equivalent to a 0-byte quota on the pool which is something we already support.

stgraber (approved):

pranav767:
> Looking into this one now. I'll probably make a few edits as I'm not sure why we need so much extra logic when we may be able to instead treat anything not listed in the config as equivalent to a 0-byte quota on the pool which is something we already support.

Hey @stgraber, just trying to understand, I think 0 byte quota still allows you to attach disk yeah?
Did you refactor anything specific / have anything else in mind?

stgraber:
Updated now with the missing validation logic.

This keeps the handling separate from the `disk` restriction and instead just applies the validation to anything with a pool reference in the aggregate validator.

PR #3165, opened 2026-04-10
incusd/instance/qemu: Scale SCSI queues with CPUs

stgraber:
Similar to what we do for networking, setup one queue per core.

Sponsored-by: Magalu Cloud (https://magalu.cloud)

### DIFF
--- a/internal/server/instance/drivers/driver_qemu.go
+++ b/internal/server/instance/drivers/driver_qemu.go
@@ -4048,7 +4048,7 @@ func (d *qemu) generateQemuConfig(machineDefinition string, cpuType string, cpuI
 		multifunction: multi,
 	}
 
-	conf = append(conf, qemuSCSI(&scsiOpts)...)
+	conf = append(conf, qemuSCSI(&scsiOpts, cpuInfo.sockets*cpuInfo.cores)...)
 
 	// Export the config directory and agent as 9p drives when supported.
 	if !isWindows && plan9 {
--- a/internal/server/instance/drivers/driver_qemu_config_test.go
+++ b/internal/server/instance/drivers/driver_qemu_config_test.go
@@ -235,6 +235,7 @@ func TestQemuConfigTemplates(t *testing.T) {
 			addr = "00.0"
 			bus = "qemu_pcie1"
 			driver = "virtio-scsi-pci"
+			num_queues = "4"
 			`,
 		}, {
 			qemuDevOpts{"ccw", "qemu_pcie2", "00.2", true},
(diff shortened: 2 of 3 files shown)


bensmrs (approved):

stgraber:
@bensmrs doh... Going to need a follow up on this one as it breaks live-migration post-upgrade for any VM with more than one CPU. We have two problems we need to deal with:

1) When live-migrating from an older version, we're going from single queue to multi-queue on the target, that crashes QEMU
2) When dealing with a VM that has had CPUs hotplugged since start, we need to get the original CPU count so we set the correct number of queues

bensmrs:
Oh… and that didn’t crash for the network stuff as the number of queues is essentially the number of FDs we are effectively letting QEMU know about (so completely different mechanism)?

stgraber:
Yeah, though I wouldn't want to bet on the network queue stuff being bulletproof either...

I've spent the last hour or so thinking about this mess and how to best solve it.
In the past my thought was to "simply" ship around a copy of the original qemu.conf so any newly spawned QEMU process would then be close enough to the original.

But that's not particularly easy to do and will almost certainly come to bite us with stuff that's not in qemu.conf...

Then there would be a solution specific to just this one issue, basically another volatile key that keeps track of the original value. If it's not set, we assume it predates the config so value is 1 and we're good.

But I really don't want to have to do that over and over again every time we tweak something.

Another idea would be an incremental generation ID type thing but that doesn't play well at all with cherry-picks nor with conditional configurations like what we have based on image.os.

Then there's the idea to capture a whole bunch of state during migration and send that as part of the protobuf migration data. This is pretty nice because it's all internal and hidden from the user. But this all falls flat as soon as you consider stateful stop and stateful snapshots, both of which use the QEMU serialization logic but don't use our internal migration mechanism.

That then led me to my current plan. Add yet another volatile key, but one that will hold a bigger JSON struct for everything related to VM state. It would only get populated if migration.stateful is set and it will take over:
 - volatile.vm.definition
 - volatile.vm.hotplug.memory

I expect to want it to include at minimum:
 - Initial QEMU definition
 - Internal hardware generation counter (may end up being useful)
 - Initial CPU layout
 - Initial memory layout
 - Memory hotplug data
 - Startup time OS definition
 - Possibly some PCIe layout info

The plan is to then have a couple of functions, one to convert the boot time VM state into that volatile struct and another to read the volatile value back into a struct we can easily query during startup.

The latter would be able to parse our existing couple of volatile keys so we don't break migration on those VMs.

To handle the SCSI multi-queue config specifically, we'd have the startup logic basically do:
 - Check if we have CPU layout info in the VM state, if we do, use that
 - If not and we're dealing with state restore, use 1 queue
 - Else, it's a clean boot, use one queue per CPU

I'm hoping this gives us enough flexibility in the future as well as a better way to detect those situations, even if it's just to refuse a live migration and require a reboot in some situations.

bensmrs:
That sure sounds good. It certainly helps with migration to newer versions, but aren’t there also cases where we can migrate to previous versions?
Regarding versions, it can also be a capability-based check, where we mark some API extensions as migration breakers and store them in a volatile list.

PR #3181, opened 2026-04-16
Rework tracking of original VM properties for migration and stateful snapshots/stop

stgraber:


internal/server/instance/drivers/driver_qemu_stateful.go
```diff
@@ -0,0 +1,97 @@
...
+		return bs.CPUTopology.Sockets * bs.CPUTopology.Cores
+	}
+
+	return 1
+}
+
+func (d *qemu) getBootState() (*qemuBootState, error) {
+	// Prepare a new state struct.
+	bs := qemuBootState{
+		Version: -1,
```
bensmrs, line 36: Why `-1` instead of `0` if the first “new” version is `1`?

internal/server/instance/drivers/driver_qemu.go
```diff
@@ -2115,20 +2042,14 @@ func (d *qemu) start(stateful bool, op *operationlock.InstanceOperation) error {
 	}
 
 	// Record the QEMU machine definition.
-	if !stateful {
+	if !stateful && d.CanLiveMigrate() {
```
bensmrs, line 2045: Why the new guard? To avoid computing useless things?

bensmrs (approved):
LGTM. A few points I don’t fully get.

internal/server/instance/drivers/driver_qemu_stateful.go
stgraber: The thought behind -1 and skipping 0 was that -1 would clearly detect any scenario where we didn't unpack an actual state struct from JSON but instead used the fallback logic.

Avoiding 0 was to avoid accidents as an uninitialized struct would have the field default to 0.

internal/server/instance/drivers/driver_qemu.go
stgraber: Yeah, exactly, the machine type is only ever used during live migration or stateful snapshot restore (unlike some of the other bits of the boot state that we access in other functions during startup), so we can save ourselves a QMP roundtrip on systems where migration isn't useful.

PR #3182, opened 2026-04-17
Various bugfixes

stgraber:


client/incus_instances.go
```diff
@@ -1715,7 +1715,7 @@ func (r *ProtocolIncus) rawConn(apiURL *url.URL, protocol string) (net.Conn, err
 		return nil, errors.New("Missing or unexpected Upgrade header in response")
 	}
 
-	return conn, err
+	return conn, nil
```
bensmrs, line 1718: That’s not really a fix; `err` is guaranteed to be `nil` here. Still, good to be explicit.

cmd/incusd/instances.go
```diff
@@ -473,27 +473,63 @@ func instancesShutdown(instances []instance.Instance) {
...
+							}
+						}
 					}
-				} else {
-					err := inst.Shutdown(time.Second * time.Duration(timeoutSeconds))
-					if err != nil {
-						logger.Warn("Failed shutting down instance, forcefully stopping", logger.Ctx{"project": inst.Project().Name, "instance": inst.Name(), "err": err})
-						err = inst.Stop(false)
+
+					// Perform the shutdown action..
```
bensmrs:
```suggestion
					// Perform the shutdown action.
```

bensmrs (approved):

client/incus_instances.go
stgraber: Yeah, indeed, not an actual issue in this scenario.

We did have another case where a similar pattern did turn into an actual bug as err could contain a non-fatal error (file not found) so wouldn't have caused an earlier return but could then cause the final return to fail.

cmd/incusd/instances.go
stgraber: Fixed!

PR #3200, opened 2026-04-20
Improve import/export usage consistency

### LINKED ISSUE #3124: Export/Import Custom Storage Volumes to/from Standard Output/Input Using `-`

I am unable to export custom storage volume backups to standard output using the common `-` alias.

Instances can be exported/imported to/from standard output/input by using `-` in place of the `<target/backup file>` with `incus export/import`.

For `incus storage volume export/import` the docs don't mention support for `-`, but I tried this in the hopes it would work the same. However, `-` is instead interpreted as a file name.

I think Incus should support standard output/input for `incus storage volume export/import` using `-` in the same way it is supported for other commands like `incus export/import`, `incus file pull/push` and `incus storage volume file pull/push`.

bensmrs:
This makes all non-image import/export commands support stdin/stdout-like path specs.
(Image commands need quite some cleanup before this can easily be ported, IMO.)
Additionally, this introduces a flag to force overriding backup files and tries to make the overall logic a bit more uniform.

Closes: #3124


bensmrs:
Not sure whether the failing tests really are my fault; re-running them.
Edit: better on the 2nd run; I’m assuming some contention issues on the runners triggering our timeouts…

stgraber:
@bensmrs I retried them a couple of times but it's the exact same tests failing on all storage backends. I'm not seeing this on other PRs so it's definitely your change causing it :)

bensmrs:
Well look at #3201

bensmrs:
And the other recent ones also, @stgraber

stgraber:
@bensmrs yeah, but there's a difference between a PR where one of the "standalone_storage" test fails and one where every single one of them fails with no other testsuite failures ;)

stgraber:
Piotr's PR, if I retry it a couple of times, I'm 99% sure all those will clear. Yours, I've retried twice and we've not gotten a single green `standalone_storage` run.

bensmrs:
Ceph worked after a retry :)

stgraber:
Hmm, right and the dir ones went through too, so it's not causing a full failure but it's definitely nudging timings in a pretty significant way :)

stgraber:
I'm pretty confused as to how importing a container that's less than 5MB large would take 2 minutes though

bensmrs:
I’m really not sure it’s the PR. 2 PRs that fix typos also trigger the timeouts. Here, it fails at the exact same place, and my only fix in the instance import route is changing an `==` with a `slices.Contains`.

stgraber:
Yeah, I just went through this PR pretty closely and I don't see it either.

There's nothing in the import codepath that would affect this particular part of the test and none of the changes to the testsuite are messing with the input for the failing test either.

bensmrs:
Exactly :D 
So I’m happily blaming Azure, with absolutely no proof.

PR #3216, opened 2026-04-22
Improve and extend SELinux support

### LINKED ISSUE #1037: Set correct SELinux context options on container & VM mounts

mschiff:
# Improve SELinux support

## Summary

This PR  expands SELinux support in Incus, addressing [issue #1037](https://github.com/lxc/incus/issues/1037). So far, getting containers and VMs to run cleanly on SELinux hosts has required manual workarounds and still left gaps that couldn't be closed without changes in Incus itself. This PR closes those gaps by adding full per-instance SELinux integration, analogous to the existing AppArmor path.

### New config keys

| Key | Description |
|---|---|
| `security.selinux.domain` | Overrides the default domain (`container_init_t` / `qemu_t`). |
| `security.selinux.type` | Overrides the default file type for the instance rootfs / disk image. |
| `security.selinux.level` | Sets the MLS/MCS level explicitly. If unset, a free MCS range is allocated automatically. |
| `volatile.selinux.context` | Persists the computed context so MCS ranges stay stable across restarts. |

### Core integration

- `internal/server/sys/selinux.go`: detection of the daemon context plus new helpers `SELinuxInstanceContext`, `SELinuxInstanceFileContext` and `SELinuxLabelTree` for computing and applying process and file contexts, including automatic MCS allocation against other instances on the node.
- `shared/subprocess`: `Process` gains a `SELinux` field and a `SetSELinux` setter; a new `SetSELinuxEnabled` toggle and an internal `withSELinuxExecContext` helper apply exec labels around subprocess spawns.
- `shared/validate`: new `IsSELinuxType` and `IsSELinuxLevel` validators for the new config keys.

### Driver integration

- **LXC containers** are started with `lxc.selinux.context` set, and the rootfs and other instance specific files are labeled with the matching file context.
- **QEMU VMs** are spawned under the computed exec context, with instance specific files labeled accordingly.


internal/server/sys/selinux.go
```diff
@@ -3,19 +3,28 @@
...
 	"github.com/lxc/incus/v6/internal/server/db/cluster"
 	"github.com/lxc/incus/v6/internal/server/db/warningtype"
-	internalUtil "github.com/lxc/incus/v6/internal/util"
+
+	"golang.org/x/sys/unix"
+
+	"github.com/lxc/incus/v6/internal/server/instance/instancetype"
 	"github.com/lxc/incus/v6/shared/logger"
 	"github.com/lxc/incus/v6/shared/util"
+	"github.com/opencontainers/selinux/go-selinux"
```
stgraber: Wrong import order, the external packages go in the middle chunk.

internal/server/sys/selinux.go
```diff
@@ -31,42 +40,218 @@ func (s *OS) initSELinux() []cluster.Warning {
...
+		"containerType": s.SELinuxContainerType,
+		"vmType":        s.SELinuxVMType,
+		"mlsEnabled":    selinux.MLSEnabled(),
+	})
+
+	return dbWarnings
+}
+
+// SELinuxInstanceFileContext derives a file context from an instance process context and type.
+func SELinuxInstanceFileContext(processCtx string, instType instancetype.Type, expandedConfig map[string]string) string {
```
stgraber: That function should be moved to a new `internal/server/selinux` package.

internal/server/sys/selinux.go
```diff
@@ -31,42 +40,218 @@ func (s *OS) initSELinux() []cluster.Warning {
...
+	return fmt.Sprintf("%s:%s:%s:%s", ctx["user"], "object_r", seType, ctx["level"])
+}
 
-	// Handle the various SELinux policy variants here.
-	switch s.SELinuxContextDaemon {
-	case "system_u:system_r:container_runtime_t:s0":
-		logger.Debugf("Detected Fedora-style SELinux setup")
-		s.SELinuxContextInstanceLXC = "system_u:system_r:spc_t:s0"
+// SELinuxLabelTree recursively labels all entries under path with the given SELinux label without crossing filesystem boundaries.
+func SELinuxLabelTree(path string, label string, skipPath string) error {
```
stgraber: That function should be moved to a new `internal/server/selinux` package.

internal/server/sys/selinux.go
```diff
@@ -31,42 +40,218 @@ func (s *OS) initSELinux() []cluster.Warning {
...
+	})
+}
+
+// SELinuxInstanceContext derives the SELinux process context for an instance.
+//
+// allocLevel is invoked only when a fresh MCS level must be generated
+// (i.e. no explicit override via security.selinux.level and no previously
+// persisted volatile.selinux.context). The caller is responsible for
+// ensuring collision-freeness against other instances on this host.
+func (s *OS) SELinuxInstanceContext(instType instancetype.Type, localConfig map[string]string, expandedConfig map[string]string, allocLevel func() (string, error)) (string, bool, error) {
```
stgraber: That function should be moved to a new `internal/server/selinux` package.

internal/server/sys/selinux.go
```diff
@@ -31,42 +40,218 @@ func (s *OS) initSELinux() []cluster.Warning {
...
+
+	// Persist when context differs from cached volatile.
+	needsPersist := localConfig["volatile.selinux.context"] != ctx
+
+	return ctx, needsPersist, nil
+}
+
+// SELinuxSetExecContext sets the SELinux transition context for the next exec on the current thread.
+// Returns a cleanup function that must always be called to clear the context and release the thread lock.
+func SELinuxSetExecContext(ctx string) (func(), error) {
```
stgraber: That function should be moved to a new `internal/server/selinux` package.

shared/subprocess/proc.go
```diff
@@ -9,13 +9,20 @@ import (
...
+	"runtime"
+	"strings"
+	"sync/atomic"
 	"syscall"
 
 	"go.yaml.in/yaml/v4"
 
+	"github.com/lxc/incus/v6/shared/logger"
 	"github.com/lxc/incus/v6/shared/util"
+	"github.com/opencontainers/selinux/go-selinux"
```
stgraber: Wrong import block, external imports go to the middle block.

shared/subprocess/proc.go
```diff
@@ -58,6 +66,37 @@ func (p *Process) hasApparmor() bool {
 	return true
 }
 
+// SetSELinuxEnabled enables or disables SELinux-aware subprocess launching.
+// It should be called once during daemon startup.
+func SetSELinuxEnabled(v bool) {
+	selinuxEnabled.Store(v)
```
stgraber: I don't think we need the selinuxEnabled stuff.
Just like with AppArmor, all of this is a no-op unless the caller used SetSELinux, so the caller can just do the check itself.

internal/server/instance/drivers/driver_common.go
```diff
@@ -43,6 +42,7 @@ import (
 	"github.com/lxc/incus/v6/shared/revert"
 	"github.com/lxc/incus/v6/shared/subprocess"
 	"github.com/lxc/incus/v6/shared/util"
+	"github.com/opencontainers/selinux/go-selinux"
```
stgraber: Wrong import ordering

internal/server/instance/drivers/driver_common.go
```diff
@@ -1712,49 +1712,84 @@ func (d *common) setOOMPriority(pid int) error {
...
-			return "", err
+			return "", fmt.Errorf("Failed to parse allocated SELinux label %q: %w", label, err)
+		}
+
+		level := parsed["level"]
+		if level == "" {
+			return "", fmt.Errorf("Allocated SELinux label %q has empty level", label)
+		}
+
+		if _, clash := used[level]; !clash {
```
stgraber: We don't do define and check syntax in Incus

internal/server/instance/drivers/driver_common.go
```diff
@@ -1712,49 +1712,84 @@ func (d *common) setOOMPriority(pid int) error {
...
+const selinuxAllocMaxAttempts = 16
+
+// selinuxAllocateLevel returns an SELinux MCS level that does not collide
+// with any level currently in use by other instances on this host. It
+// delegates generation to the upstream selinux package and retries on
+// collision. Callers must hold d.state.OS.SELinuxAllocLock() for the
+// duration of the allocation and the subsequent persistence of the
+// resulting context (volatile.selinux.context), so that two instances
+// starting concurrently cannot observe the same "free" level.
+func (d *common) selinuxAllocateLevel() (string, error) {
```
stgraber: I think this and the following function could also be moved to the `internal/server/selinux` package. Rather than require the caller to grab a global lock, calling this function could acquire a lock and return a release function to the caller instead.

internal/server/sys/os.go
```diff
@@ -75,9 +75,15 @@ type OS struct {
...
-	SELinuxContextInstanceLXC string
+	SELinuxEnabled       bool
+	SELinuxContextDaemon string
+	SELinuxContainerType string
+	SELinuxVMType        string
+
+	// selinuxAllocMu serializes SELinux MCS level allocation across
+	// concurrently starting instances, so collision detection against
+	// already-used levels stays race-free.
+	selinuxAllocMu sync.Mutex
```
stgraber: As mentioned, I think this should be moved to the new internal/server/selinux package and have the allocating function over there return a release/unlock function to the caller. The caller can then either call it at the right time or defer it.

stgraber:
Looking pretty good!

There are some basic nits like the import order, but other than that I think the main change would be to move most of the logic into a new `internal/server/selinux` package.

The OS logic will still do the startup time detection and will have the boolean to report it being enabled and the attributes to get the right labels, but all the allocation and labeling functions should instead be in the new package.

The other change would be in how we handle the locking as I think it'd be cleaner to have calls to the allocator return a release function to the caller rather than expect the caller to know that it needs to acquire and release the lock.

stgraber:
That's not going to make it into Incus 7.0 unfortunately but nothing in here prevents backporting so it will get backported and included in 7.0.1.

mschiff:
@stgraber I think I resolved all your comments. I also rebased the PR

mschiff:
@stgraber I need some advice on how to fix the way of setting the SELinux context for qemu VMs, please.

The current implementation in this PR sets the context in `shared/subprocess/proc.go`  via `withSELinuxExecContext()`. It uses a proper way to do this. But:

Result:
```
Error: Unable to start process: /usr/bin/incusd forklimits limit=memlock:unlimited:unlimited fd=3 fd=4 -- /usr/bin/qemu-system-x86_64
```

The SELinux policy allows `incusd`, running as `incusd_t` to start `qemu` as `qemu_t`. The problem here is that the thread will start `incusd forklimits` instead which is denied by the kernel because incusd is not supposed to run incusd as qemu_t.

A possible solution would be to extend the forklimits command so that it can be given a selinux context and then will set the proper context right before starting qemu.
This would basically mean moving parts from `shared/subprocess/proc.go` to `cmd/incusd/main_forklimits.go`

My idea is to extend forklimits from

`cmd.Use = "forklimits [fd=<number>...] [limit=<name>:<softlimit>:<hardlimit>...] -- <command> [<arg>...]"`

to

`cmd.Use = "forklimits [fd=<number>...] [limit=<name>:<softlimit>:<hardlimit>...] [secontext=<selinux context>] -- <command> [<arg>...]"`

Should the locking remain in `shared/subprocess/proc.go` then?

What do you think? Would you support that approach or do you have something better?

stgraber:
Hmm, that's pretty annoying...

I think the clean solution, but a bit more work for you is:
 - Rename `forkstart` to `forklxc`
 - Rename `forklimits` to `forkqemu`
 - Add the SELinux label logic to `forkqemu`
 - Drop the change to shared/subprocess as it's unused

Does that make sense?

mschiff:
> Hmm, that's pretty annoying...
> 
> I think the clean solution, but a bit more work for you is:
> 
>     * Rename `forkstart` to `forklxc`
> 
>     * Rename `forklimits` to `forkqemu`
> 
>     * Add the SELinux label logic to `forkqemu`
> 
>     * Drop the change to shared/subprocess as it's unused
> 
> 
> Does that make sense?

Thanks!

Sounds good to me, I will prepare and add more commits to this PR for review in the next days.

mschiff:
@stgraber there were some no-op checks missing for non-SELinux systems, which is fixed now. Is this ready for merge for you now?

mschiff:
@stgraber Ok, all checks have passed now ;-)

stgraber:
Been working on rebasing and cleaning up this one. I'm almost done, so we should have the change in 7.2.

PR #3218, opened 2026-04-23
incus-agent: Use psutil for process count

stgraber:
The counting logic was widely off for Linux VMs, so use the psutil logic instead.

### DIFF
--- a/cmd/incus-agent/os_linux.go
+++ b/cmd/incus-agent/os_linux.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/mdlayher/vsock"
+	"github.com/shirou/gopsutil/v4/process"
 	"golang.org/x/sys/unix"
 
 	"github.com/lxc/incus/v6/internal/linux"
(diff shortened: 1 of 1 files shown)


bensmrs:
If we start using `gopsutil` for the Linux agent, we may as well go a little deeper and refactor `os_common.go` -> `os_common.go`+`os_nonlinux.go`, remove all the bits of the Linux agent from `os_linux.go` that we are fine giving to `gopsutil` and call it a day. That’d avoid duplicating code.
Best scenario would be an empty `os_nonlinux.go`, but not sure that’ll happen.

stgraber:
@bensmrs yeah, I agree that we may be able to move more logic over to psutil, maybe to do the point of deprecating all of the metrics stuff in favor of it.

Here I just did the proc stuff as the logic we had would only properly account for the first level of processes, the attempt at recursion wasn't working leading to widely off values affecting production environments.

bensmrs:
Imma let that one slide :)
But it would be nice if you could identify later the bits that could use `gopsutil` to factor the logic a bit better.

bensmrs (approved):

PR #3226, opened 2026-04-24
shared/validate: Fix name validation on single character

stgraber:


shared/validate/validate.go
```diff
@@ -586,7 +586,7 @@ func IsAPIName(value string, allowSlashes bool) error {
 	}
 
 	// Check beginning and end.
-	match, err := regexp.MatchString(`^[a-zA-Z0-9]+.*[a-zA-Z0-9]+$`, value)
+	match, err := regexp.MatchString(`^[a-zA-Z0-9]+($|.*[a-zA-Z0-9]+)$`, value)
```
bensmrs: Maybe
```suggestion
	match, err := regexp.MatchString(`^[a-zA-Z0-9]+(.*[a-zA-Z0-9]+)?$`, value)
```
?

bensmrs: Although the cost of a regex may probably justify doing two simple char tests instead

stgraber: Did some testing and it covers all the cases I care about. Updated the PR now.

bensmrs (approved):

PR #3234, opened 2026-04-26
Bunch of typo fixes in sources strings

kazan417:
Foxes many typos of sorce strings.
Also some more descriptive terms for some strings.


cmd/incus/info.go
```diff
@@ -129,7 +129,7 @@ func (c *cmdInfo) renderGPU(gpu api.ResourcesGPUCard, prefix string, initial boo
 		}
 
 		if gpu.DRM.ControlName != "" {
-			fmt.Printf(prefix+"  "+i18n.G("Control: %s (%s)")+"\n", gpu.DRM.ControlName, gpu.DRM.ControlDevice)
+			fmt.Printf(prefix+"  "+i18n.G("Vision control api: %s (%s)")+"\n", gpu.DRM.ControlName, gpu.DRM.ControlDevice)
```
stgraber: Original string is correct

cmd/incus/config_device.go
```diff
@@ -364,7 +364,7 @@ func (c *cmdConfigDeviceOverride) command() *cobra.Command {
 	cmd.Use = cli.U("override", cmdConfigDeviceOverrideUsage...)
 	cmd.Short = i18n.G("Copy profile inherited devices and override configuration keys")
 	cmd.Long = cli.FormatSection(color.DescriptionPrefix, i18n.G(
-		`Copy profile inherited devices and override configuration keys`))
+		`Copy profile inherited devices overriding configuration keys`))
```
stgraber: Short and Long should be kept in sync.

The original string is also technically more correct than the suggested change as the original suggests that the device is copied and then configuration keys overridden (which is accurate). Whereas the new version suggests that configuration keys are being overridden as a consequence of the copy.

cmd/incus/cluster.go
```diff
@@ -1419,7 +1419,7 @@ func (c *cmdClusterUpdateCertificate) run(cmd *cobra.Command, args []string) err
 	if util.PathExists(certf) {
 		err = os.WriteFile(certf, cert, 0o644)
 		if err != nil {
-			return fmt.Errorf(i18n.G("Could not write new remote certificate for remote '%s' with error: %v"), remoteName, err)
+			return fmt.Errorf(i18n.G("Could not write new certificate for remote client '%s' with error: %v"), remoteName, err)
```
stgraber: Original string is correct.

`remote client` doesn't mean anything in Incus' context and what we are writing is the remote certificate for that remote.

cmd/incus/admin_sql.go
```diff
@@ -37,7 +37,7 @@ func (c *cmdAdminSQL) command() *cobra.Command {
   cluster-specific data (such as profiles, containers, etc).
 
   If you are running a non-clustered server, the same applies, as that
-  instance is effectively a single-member cluster.
+  installation is effectively a single-member cluster.
```
stgraber: I'd just replace that with:
```
Non-clustered servers still have both local and global databases.
```

cmd/incus/image.go
```diff
@@ -662,7 +662,7 @@ var cmdImageImportUsage = u.Usage{u.Sequence(u.Either(u.Tarball, u.Directory, u.
 func (c *cmdImageImport) command() *cobra.Command {
 	cmd := &cobra.Command{}
 	cmd.Use = cli.U("import", cmdImageImportUsage...)
-	cmd.Short = i18n.G("Import images into the image store")
+	cmd.Short = i18n.G("Import images into the image storage")
```
stgraber: `image store` is correct terminology

stgraber: We have an image store (/1.0/images), image storage is a different concept entirely in Incus which is what storage volume backs the compressed copy of the image.

cmd/incus/info.go
```diff
@@ -242,7 +242,7 @@ func (c *cmdInfo) renderNIC(nic api.ResourcesNetworkCard, prefix string, initial
 				fmt.Print(prefix + "    " + i18n.G("Infiniband:") + "\n")
 
 				if port.Infiniband.IsSMName != "" {
-					fmt.Printf(prefix+"      "+i18n.G("IsSM: %s (%s)")+"\n", port.Infiniband.IsSMName, port.Infiniband.IsSMDevice)
+					fmt.Printf(prefix+"      "+i18n.G("infiniband IsSM: %s (%s)")+"\n", port.Infiniband.IsSMName, port.Infiniband.IsSMDevice)
```
stgraber: That sounds quite duplicate when the key is already in the Infiniband section. I'd keep it as it is.

cmd/incus/info.go
```diff
@@ -138,7 +138,7 @@ func (c *cmdInfo) renderGPU(gpu api.ResourcesGPUCard, prefix string, initial boo
 	}
 
 	if gpu.Nvidia != nil {
-		fmt.Print(prefix + i18n.G("NVIDIA information:") + "\n")
+		fmt.Print(prefix + i18n.G("GPU information:") + "\n")
```
stgraber: This is the NVIDIA specific section in the GPU section, current title is correct.

cmd/incus/remote_unix.go
```diff
@@ -40,7 +40,7 @@ func (c *cmdRemoteProxy) command() *cobra.Command {
 	cmd.Use = cli.U("proxy", cmdRemoteProxyUsage...)
 	cmd.Short = i18n.G("Run a local API proxy")
 	cmd.Long = cli.FormatSection(color.DescriptionPrefix, i18n.G(
-		`Run a local API proxy for the remote`))
+		`Run a local API proxy for the remote connection`))
```
stgraber: We're referring to an Incus CLI remote here, not a remote connection

cmd/incus/project.go
```diff
@@ -670,7 +670,7 @@ func (c *cmdProjectList) run(cmd *cobra.Command, args []string) error {
 		for _, column := range columns {
 			if column.Name == i18n.G("NAME") {
 				if project.Name == info.Project {
-					project.Name = fmt.Sprintf("%s (%s)", project.Name, i18n.G("current"))
+					project.Name = fmt.Sprintf("%s (%s)", project.Name, i18n.G("currently in use"))
```
stgraber: I prefer the original as it's far shorter. The project list is pretty wide already, so doubling the length of the marker here isn't ideal.

cmd/incus/remote.go
```diff
@@ -1020,7 +1020,7 @@ func (c *cmdRemoteList) remoteNameColumnData(name string, _ config.Remote) strin
 
 	strName := name
 	if name == conf.DefaultRemote {
-		strName = fmt.Sprintf("%s (%s)", name, i18n.G("current"))
+		strName = fmt.Sprintf("%s (%s)", name, i18n.G("currently in use"))
```
stgraber: Same as with project

cmd/incus/remote.go
```diff
@@ -496,7 +496,7 @@ func (c *cmdRemoteAdd) run(cmd *cobra.Command, args []string) error {
 			digest := localtls.CertFingerprint(certificate)
 
 			fmt.Printf(i18n.G("Certificate fingerprint: %s")+"\n", digest)
-			fmt.Print(i18n.G("ok (y/n/[fingerprint])?") + " ")
+			fmt.Print(i18n.G("accept (y/n/[fingerprint])?") + " ")
```
stgraber: If I were to reword this one, I'd go with `correct` instead of `accept`.
But given the high likelihood that this particular prompt has been scripted by users, I don't see this change as improving usability enough to justify breaking a bunch of users over.

cmd/incus/info.go
stgraber: `Vision control api` doesn't mean anything. `Control` is correct here in this context and lines up with what the operating system shows.

stgraber:
I've cherry-picked the changes that seemed correct and beneficial, merged them together, tweaked things a tiny bit and re-generated the translation template.

PR #3240, opened 2026-04-27
Improve flags

bensmrs:


bensmrs:
Adding linting rules so that no new occurrence of `cmd\.Flags\(\)\..*` is allowed would be nice, but that’s waaaaay beyond my limited knowledge of Go’s tooling ecosystem.

PR #3249, opened 2026-04-29
Bump minimum versions and remove version checks

stgraber:


bensmrs:
Not a huge fan of the DRBD check removal. The overwhelming majority of distros don’t ship DRBD9 IIRC, so by default, people end up with DRBD8 if they are not careful.

stgraber:
But then most distros also don't package the Linstor satellite? Doesn't the satellite include a strict dependency on drbd9?

bensmrs:
That’s where things get tricky. No LINSTOR package depends on DRBD (https://github.com/LINBIT/linstor-server/blob/master/debian/control) because LINSTOR can be used without (https://linbit.com/drbd-user-guide/linstor-guide-1_0-en/#s-linstor-without-drbd). `drbd-utils` is suggested, but it doesn’t depend on a specific version of DRBD.

stgraber:
Ah, okay, then yeah, makes sense to keep that last commit.

stgraber:
@bensmrs updated the PR to instead change the comment to explain why we're keeping this particular version check.

bensmrs (approved):
I’m not knowledgeable about some of the other dependencies, but LINSTOR seems to be the only outlier when it comes to it. Feel free to require another approval.

PR #3256, opened 2026-04-29
Allow overriding storage pools for dependent volumes during instance migration

presztak:


internal/migration/migrate.proto
```diff
@@ -80,6 +80,8 @@ message DependentVolume {
 	optional btrfsFeatures		btrfsFeatures	= 7;
 	optional int64			volumeSize	= 8;
 	repeated Snapshot               snapshots	= 9;
+	optional string			newName		= 10;
+	optional string			newPool		= 11;
```
stgraber: I'm not sure that I love the idea of having the migration struct provide both old and new pools and volume names.

My understanding is that you had to do that because just putting the new name and pool info in the instance struct sent through POST causes an issue during migration as we need a way to align the received migration data with the instance configuration.

But wouldn't it be simpler to remove the name and pool fields for the migration struct and instead have the DependentVolume entry use device_name? That would then let us get the pool and volume names based on that device name, getting the new values that way?

presztak: Hmmm... yeah, it should work. Like you said, I used both values to align the source and target and to know what needs to be sent/received on both ends.

presztak: Done.

PR #3266, opened 2026-04-30
Update console.go

kazan417:
term SPICE


stgraber:
Nope, that's wrong. `spicy` is a SPICE client just like `remote-viewer`.

PR #3274, opened 2026-04-30
Implement core.shutdown_action

stgraber:


bensmrs (approved):
Automatic restore would pair well

stgraber:
@bensmrs yeah, we adding an option to do that, possibly after some kind of delay would probably make sense. We'd need to also store a flag to know that the server was automatically evacuated rather than manually evacuated though.

PR #3275, opened 2026-04-30
Add --reuse to incus image copy --copy-aliases and fix cluster join bug

stgraber:


cmd/incus/image.go
```diff
@@ -211,6 +213,10 @@ func (c *cmdImageCopy) run(cmd *cobra.Command, args []string) error {
 		return errors.New(i18n.G("Auto update is only available in pull mode"))
 	}
 
+	if c.flagReuse && !c.flagCopyAliases {
```
bensmrs, line 216: Can’t it just implicitly set it?

cmd/incus/image.go
```diff
@@ -267,6 +273,24 @@ func (c *cmdImageCopy) run(cmd *cobra.Command, args []string) error {
 		Profiles:    c.flagProfile,
 	}
 
+	// If --reuse was passed, delete any conflicting aliases on the target so they can be recreated.
+	if c.flagReuse {
+		conflicting := append([]api.ImageAlias{}, imgInfo.Aliases...)
```
bensmrs, line 278: I don’t get why you’re initializing that way

cmd/incus/image.go
stgraber: You may not want to step over existing aliases, so having `--copy-aliases` imply `--reuse` would be wrong and the other way around is a bit weird as I wouldn't expect most folks to expect an option named `--reuse` to set `--copy-aliases` for them.

cmd/incus/image.go
stgraber: I don't think it's the first time that I use `append` as a way to get a copy of a slice. It's a bit more compact than having to make+copy a slice.

bensmrs: Ok, I find `copy` easier to read, but good enough.

bensmrs (approved):

cmd/incus/image.go
bensmrs: I was implying the second case

PR #3278, opened 2026-04-30
incusd/device/nic: Set next-hop based on configured IP addresses

### LINKED ISSUE #3156: NIC ipv[46].routes: use ipv[46].address as next-hop

When a container starts, I would like to route a subnet to that container's external interface. For example:

* `incusbr0` has address 10.11.12.1/24 on the host
* container has address 10.11.12.58 on its eth0 interface on incusbr0
* I want to route subnet 192.168.0.0/24 to the container

At the command line, this would be done using: `ip route add 192.168.0.0/24 via 10.11.12.58`

Incus has an `ipv4.routes` option for both the NIC and the bridge, but neither allows a next-hop to be specified.  What you get is an ARP broadcast instead, as if the network is directly connected to `incusbr0`.

Discussion [here](https://discuss.linuxcontainers.org/t/device-ipv4-routes-route-on-host-has-no-next-hop/26511).

The proposed option from @stgraber is: if the NIC has an `ipv4.address` configured, then use this as the next-hop when adding the `ipv4.routes` route(s).  (And of course, ditto for ipv6)

This is not strictly backwards-compatible, but if you want the old behaviour, you can set `ipv4.routes` on the bridge instead of on the NIC.

stgraber:
Closes #3156

### DIFF
--- a/internal/server/device/device_utils_network.go
+++ b/internal/server/device/device_utils_network.go
@@ -405,9 +405,26 @@ func networkVethFillFromVolatile(device deviceConfig.Device, volatile map[string
 }
 
 // networkNICRouteAdd applies any static host-side routes configured for an instance NIC.
-func networkNICRouteAdd(routeDev string, routes ...string) error {
+// If viaIPv4 or viaIPv6 are non-empty, they are used as the next-hop gateway for routes of the matching family.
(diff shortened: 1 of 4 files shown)


bensmrs (approved):
Way easier to review after reading the whole forum thread!

PR #3287, opened 2026-05-01
Update create.go

kazan417:
another missed explanation


stgraber:
Will apply through my CLI branch

PR #3288, opened 2026-05-01
Update launch.go

kazan417:
fix third missed explanation


stgraber:
Will pick up in my CLI branch

PR #3292, opened 2026-05-01
Update network_zone.go

kazan417:
add explanation for zone command


stgraber:
Going to apply this through my CLI branch

PR #3299, opened 2026-05-02
Fix another set of explanation typos.

kazan417:
Fix another set of explanation typos.


kazan417:
looks like it is last typos.

kazan417:
Maybe also 'socket file' should be replaced with 'UNIX socket file' (not included in this pull request)

PR #3312, opened 2026-05-04
Fix some network issues

stgraber:


cmd/incusd/main_forknet.go
```diff
@@ -653,6 +665,52 @@ func (c *cmdForknet) dhcpRunV6(errorChannel chan error, iface string, hostname s
...
+		return nil, err
+	}
+
+	var hwAddr net.HardwareAddr
+	for _, ifi := range ifaces {
+		if ifi.Flags&net.FlagLoopback != 0 {
+			continue
+		}
+
+		if len(ifi.HardwareAddr) < 4 {
```
bensmrs, line 688: When does that happen?

bensmrs (approved):

cmd/incusd/main_forknet.go
stgraber: You get > 4 on Infiniband interfaces (not really a problem), you can get < 4 on tun/wireguard interfaces I believe.

PR #3327, opened 2026-05-08
Implement bridge.multicast_snooping

stgraber:


internal/server/network/driver_bridge.go
```diff
@@ -1165,13 +1174,23 @@ func (n *bridge) setup(oldConfig map[string]string) error {
...
+		// Enable VLAN filtering for Linux bridges.
 		err = BridgeVLANFilterSetStatus(n.name, "1")
 		if err != nil {
 			n.logger.Warn(fmt.Sprintf("Failed enabling VLAN filtering: %v", err))
 		}
+
+		// Apply multicast snooping setting.
+		err = BridgeMulticastSnoopingSetStatus(n.name, util.IsTrueOrEmpty(n.config["bridge.multicast_snooping"]))
+		if err != nil {
+			if n.config["bridge.multicast_snooping"] != "" {
```
bensmrs, line 1188: Why just check the empty value?

stgraber: If not explicitly set by the user, it shouldn't be fatal. If explicitly set one way or the other by the user, it should be fatal.

bensmrs (approved):

internal/server/network/driver_bridge.go
bensmrs: Fair

PR #3337, opened 2026-05-11
Adjust instance-agent events

0xk1f0:
I previously contributed #3026, which added the firing of events when an instance's agent started/stopped.

During testing, I discovered that in some cases, it was possible to produce a case in which `instance-agent-stopped` was fired twice, because there is currently no check in the `onStop()` hook that checks if the agent event has been fired before.

In the process of fixing this, I thought it would be a cool idea to also expose the agent state in the instance's config, so I added an additional parameter `volatile.last_state.agent` that now tracks it.

The checks before firing the lifecycle events should prevent two events from being fired f.E. if the agent was killed in the instance and is then being killed forcefully `-f`.

### DIFF
--- a/doc/config_options.txt
+++ b/doc/config_options.txt
@@ -3111,6 +3111,12 @@ The cluster member that the instance lived on before evacuation.
 
 ```
 
+```{config:option} volatile.last_state.agent instance-volatile
+:shortdesc: "Instance agent state as of last host shutdown"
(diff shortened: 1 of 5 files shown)


0xk1f0:
@stgraber Sorry, I was a bit sloppy with the code formatting there, should be good now.

PR #3345, opened 2026-05-16
Update Ansible section with incus-client details

chrisjsimpson:
Added information about using the 'incus-client' package for Ansible connections, as it took a while to find this information and doesnt refernce for the ubuntu installation (though it is [mentioned](https://linuxcontainers.org/incus/docs/main/installing/) for the default `Alpine` distro after looking here)

### DIFF
--- a/doc/third_party.md
+++ b/doc/third_party.md
@@ -14,7 +14,7 @@ The integration with Incus is done through a [dedicated provider](https://github
 [Ansible](https://www.ansible.com) is an infrastructure as code tool with particular focus on software provisioning and configuration management.
 It does most of its work by first connecting to the system that it's deploying software on.
 
-To do that, it can connect over SSH and a variety of other protocols, one of which is [Incus](https://docs.ansible.com/ansible/latest/collections/community/general/incus_connection.html).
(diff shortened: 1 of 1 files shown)


doc/third_party.md
```diff
@@ -14,7 +14,7 @@ The integration with Incus is done through a [dedicated provider](https://github
 [Ansible](https://www.ansible.com) is an infrastructure as code tool with particular focus on software provisioning and configuration management.
 It does most of its work by first connecting to the system that it's deploying software on.
 
-To do that, it can connect over SSH and a variety of other protocols, one of which is [Incus](https://docs.ansible.com/ansible/latest/collections/community/general/incus_connection.html).
+To do that, it can connect over SSH and a variety of other protocols, one of which is [Incus](https://docs.ansible.com/ansible/latest/collections/community/general/incus_connection.html). When using the incus ansible connection type, you may only want to install the [`incus-client` package](https://github.com/zabbly/incus#other-packages) on your ansible control node rather than the full `incus` package.
```
stgraber: Zabbly is just one of many package sources, we shouldn't special case it here.

We should probably say something along the lines of `When using the incus ansible connection type with a remote Incus server, you may benefit from using the incus-client package rather than the full incus package when such a package is available in your distribution`

chrisjsimpson: Your wording is much better, new wording is up. Thanks

PR #3352, opened 2026-05-19
Various bugfixes

stgraber:


internal/server/instance/drivers/driver_lxc.go
bensmrs, line 1: Maybe comment somewhere that LXC does some strange quote checks and that this is safe even if the variable contains quotes (at least, that’s what I understood from the lxc/lxc discussion).

bensmrs (approved):

bensmrs:
LINSTOR looks unhappy, I’ll rerun the checks

bensmrs:
The error has been ignored in a few PRs now. It first appeared in #3338, which suggests it’s an upstream change.

*sigh* https://github.com/LINBIT/linstor-server/blob/master/CHANGELOG.md#1333---2026-05-11

Any idea what’s causing a shrink anyway @stgraber?

stgraber:
@bensmrs The standalone case appears to be an actual container shrink test, so that's now failing given they removed support for it. The cluster case is less obvious but may also be effectively the same thing. The pool's default volume size would be 10GiB. If the container's profile instead sets something smaller, then a shrink would happen on creation which would similarly fail.

So I think the errors are correct from the Incus side, it's the tests that will need updating and probably a note added to the documentation.

internal/server/instance/drivers/driver_lxc.go
stgraber: @bensmrs done

bensmrs:
Ok, I’ll have a look

PR #3354, opened 2026-05-19
test: Disable volume shrinking with LINSTOR

bensmrs:


bensmrs:
Maybe you can help me understand what’s going on, @stgraber.
The default pool seems to be set here: https://github.com/lxc/incus/blob/bf82cd21d3f7be28520ceef073c4e578be75c5d7/test/suites/storage.sh#L255
But that pool is a `dir` pool: https://github.com/lxc/incus/blob/bf82cd21d3f7be28520ceef073c4e578be75c5d7/test/suites/storage.sh#L170
So how is this setting something on LINSTOR? https://github.com/lxc/incus/blob/bf82cd21d3f7be28520ceef073c4e578be75c5d7/test/suites/storage.sh#L787

I’m missing something obvious.

bensmrs:
Oh that’s the `INCUS_DIR` variable at the beginning of the subshell

bensmrs:
Shrinking-wise, it’s now ok. However, the old “multiple resource definitions” is back…

bensmrs:
I have an idea how to solve that, will come up with a PR

PR #3362, opened 2026-05-19
Various bugfixes

stgraber:


cmd/incusd/networks.go
```diff
@@ -1831,14 +1848,26 @@ func networkStartup(s *state.State) error {
...
+		pns := slices.Collect(maps.Keys(initNetworks[priority]))
+		initNetworksMu.Unlock()
 
-				continue
-			}
+		group := new(errgroup.Group)
+		group.SetLimit(numParallel)
+
+		for _, pn := range pns {
+			p := pn
```
bensmrs: Just to be sure, this is done for scoping reasons to have the closure capture the current iteration and not only the last one?

cmd/incusd/networks.go
```diff
@@ -1831,14 +1848,26 @@ func networkStartup(s *state.State) error {
 
 	// Try initializing networks in priority order.
 	for priority := range initNetworks {
-		for pn := range initNetworks[priority] {
-			err := loadAndInitNetwork(pn, priority, true)
-			if err != nil {
-				logger.Error("Failed initializing network", logger.Ctx{"project": pn.ProjectName, "network": pn.NetworkName, "err": err})
+		initNetworksMu.Lock()
```
bensmrs: Not sure why we need to enter the critical section in this loop; this is run sequentially as we’re waiting for the `errgroup`.

cmd/incusd/networks.go
```diff
@@ -1862,34 +1891,51 @@ func networkStartup(s *state.State) error {
...
-					tryInstancesStart := false
+					var tryInstancesStart atomic.Bool
 
 					// Try initializing networks in priority order.
 					for priority := range initNetworks {
-						for pn := range initNetworks[priority] {
-							err := loadAndInitNetwork(pn, priority, false)
-							if err != nil {
-								logger.Error("Failed initializing network", logger.Ctx{"project": pn.ProjectName, "network": pn.NetworkName, "err": err})
+						initNetworksMu.Lock()
```
bensmrs: Same remark here. I’d also love to see a bit of refactoring, as that code sure looks similar to the one above.

internal/server/device/nic_bridged.go
bensmrs, line 1: Wow that’s dark

bensmrs:
The rest looks fine, although I don’t have the Ceph knowledge to review the related commit.

cmd/incusd/networks.go
breml: Since Go 1.22 this is no longer necessary (https://go.dev/doc/go1.22#language).

stgraber:
@bensmrs tweaked the logic a bit to limit duplication

bensmrs (approved):

PR #3363, opened 2026-05-19
Allow volume creation on attach

stgraber:


bensmrs (changes requested):
A `instanceDeviceAdd` failure doesn’t revert the volume creation, I think it should be added in a reverter. Additionally, I’d expect the type to be `block` for VMs when no path is given. If you don’t feel like checking the instance type, could you instead make it obvious in the flag description that the type is `filesystem`?

stgraber:
Hmm, yeah, makes sense that for VM, we'd want block instead.

stgraber:
@bensmrs done and added a test for good measure

bensmrs (approved):

PR #3387, opened 2026-05-24
Update network_forward.go

kazan417:
again typo now  only 1 letter fix


kazan417:
Firrst is 1 letter typo - found while translating into swedish.
Second is example naming unification that is not necessary, but looks reasonable for me.
I belive that is last typo edit.

PR #3394, opened 2026-05-25
Implement ranges in network address sets

### LINKED ISSUE #3389: IP ranges in network address-set

using address-sets in acls make them very easy to use. In ACLs We can use both ranges and subnets. Address sets allow only subnets.

incus network address-set add <NAME> 10.0.0.120-10.0.0.130

stgraber:
Closes #3389

### DIFF
--- a/doc/howto/network_address_sets.md
+++ b/doc/howto/network_address_sets.md
@@ -5,7 +5,7 @@
 Network address sets are working with {ref}`ACLs <network-acls>` and work only with {ref}`network-ovn` or with {ref}`bridged networks <network-bridge-firewall>` using `nftables` only.
 ```
 
-Network address sets are a list of either IPv4, IPv6 addresses with or without CIDR suffix. They can be used in source or destination fields of {ref}`ACLs <network-acls-rules-properties>`.
+Network address sets are a list of either IPv4, IPv6 addresses with or without CIDR suffix. IP ranges (for example `10.0.0.120-10.0.0.130`) are also supported, though a single range may expand to at most 256 addresses. Larger sets of addresses should be expressed using CIDR notation instead. Address sets can be used in source or destination fields of {ref}`ACLs <network-acls-rules-properties>`.
 
 ## Address set properties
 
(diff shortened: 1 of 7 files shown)


doc/howto/network_address_sets.md
```diff
@@ -5,7 +5,7 @@
 Network address sets are working with {ref}`ACLs <network-acls>` and work only with {ref}`network-ovn` or with {ref}`bridged networks <network-bridge-firewall>` using `nftables` only.
 ```
 
-Network address sets are a list of either IPv4, IPv6 addresses with or without CIDR suffix. They can be used in source or destination fields of {ref}`ACLs <network-acls-rules-properties>`.
+Network address sets are a list of either IPv4, IPv6 addresses with or without CIDR suffix. IP ranges (for example `10.0.0.120-10.0.0.130`) are also supported, though a single range may expand to at most 256 addresses, larger sets of addresses should be expressed using CIDR notation instead. They can be used in source or destination fields of {ref}`ACLs <network-acls-rules-properties>`.
```
hallyn: I think splitting the last part of the new text into a new sentence would read better, e.g.

```
IP ranges (for example `10.0.0.120-10.0.0.130`) are also supported, though a single range may expand to at most 256 addresses. Larger sets of addresses should be expressed using CIDR notation instead
```

doc/howto/network_address_sets.md
```diff
@@ -46,7 +46,7 @@ Adding addresses is pretty straightforward:
 incus network address-set add <name> <address1> <address2>
 ```
 
-There is no restriction about the kind of address you are appending in your set, a mix of IPv4, IPv6 and CIDR can be used without disruption.
+There is no restriction about the kind of address you are appending in your set, a mix of IPv4, IPv6, CIDR subnets and IP ranges can be used without disruption.
```
hallyn: Again, I'd change this, make the first comma either a semicolon or a period (or even a colon).

hallyn (approved):
I thought I'd have some real comments, but then saw the validate.IsNetworkRange() which takes care of all of them :)

So just two comments on the docs, take them or leave them.

Thanks.

PR #3395, opened 2026-05-25
More bugfixes

stgraber:


internal/server/storage/drivers/driver_zfs_volumes.go
```diff
@@ -959,70 +959,92 @@ func (d *zfs) CreateVolumeFromMigration(vol Volume, conn io.ReadWriteCloser, vol
...
+			if commonSrcIdx >= 0 {
+				// Send only snapshots strictly newer than the common base.
+				for _, dataset := range migrationHeader.SnapshotDatasets[commonSrcIdx+1:] {
+					syncSnapshots = append(syncSnapshots, &migration.Snapshot{Name: &dataset.Name})
+				}
 
-			// Let the source know that we need all snapshots.
-			syncSnapshots = []*migration.Snapshot{}
+				// Keep target snapshots whose GUID matches a source snapshot
+				// up through the common base; delete the rest.
```
bensmrs:
```suggestion
				// up to the common base; delete the rest.
```

internal/server/storage/drivers/driver_zfs_volumes.go
```diff
@@ -959,70 +959,92 @@ func (d *zfs) CreateVolumeFromMigration(vol Volume, conn io.ReadWriteCloser, vol
...
 
-				for _, migrationSnap := range migrationHeader.SnapshotDatasets {
-					if snapName == migrationSnap.Name {
-						targetOnlySnapshot = false
-						break
+					err = d.DeleteVolume(snapVol, op)
+					if err != nil {
+						return err
 					}
 				}
```
bensmrs, line 1029: Just to be sure, what’s the iteration order? Does ZFS support deleting random snapshots at low cost even if it’s not the last one?
It looks like that level of indirection is useless if we already have the greater common index, am I wrong?

bensmrs (changes requested):

internal/server/storage/drivers/driver_zfs_volumes.go
stgraber: Yeah, ZFS lets you delete snapshots extremely cheaply so long as they're not used as the source of a clone. It's very common for ZFS tooling to create snapshots of all datasets every few minutes or hours and constantly rotate/clear older ones.

bensmrs (approved):

internal/server/storage/drivers/driver_zfs_volumes.go
stgraber: `snapshots` are the local (target) snapshots, `respSnapshots` is the equivalent but with the ZFS data (GUID) added. Then `migrationHeader.SnapshotDatasets` is the equivalent of `respSnapshots` but on the source side.

They are both sorted by their creation order, so we want to pick on the latest shared GUID between source and target.

The extra fun of course is that the target may have created its own set of snapshots, mixed into the snapshot list.

bensmrs: Ah, fun indeed…

PR #3404, opened 2026-05-26
Implement storage volume rebuild

### LINKED ISSUE #3356: `incus rebuild`: Add option to wipe attached custom volumes

Wipe custom volumes attached to an instance rebuilt with `incus rebuild`.

Use case: reset the custom volumes attached to an instance when doing tests with e.g. OCI containers.

Currently one has to detach, delete, create and attach them again or mount them locally and delete their content, which is quite tedious.

It would be great to have a `--wipe-volumes` or similar option to be used in conjunction with `incus rebuild`, which would do the heavy lifting of destroying and recreating the custom volumes.

Because it could lead to data loss, it could ask interactively whether the user is ABSOLUTELY SURE the volume(s) attached to the instance can be destroyed, and the existing `--force` would override the need to reply interactively.

stgraber:
Closes #3356


stgraber:
@presztak could you do a quick review of this one?

presztak (approved):
LGTM

PR #3405, opened 2026-05-26
Fix remaining golangci-lint issues

### LINKED ISSUE #2636: Get the entire code base to pass `golangci-lint run`

stgraber:
This is the result of auto-fixing and splitting the resulting changes per package and per linter.

Closes #2636


stgraber:
Opening as draft as I need to actually go commit by commit myself once :)

stgraber:
Went through all the commits and everything looks good to me.

The least obvious ones of them are the fixes for "defer inside of a loop" which result in code that's probably harder to follow than the original, but not sufficiently worse to start adding a bunch of nolints over.

bensmrs (approved):

PR #3406, opened 2026-05-26
Update remote.go

kazan417:
Better usability. 
Accept not only localized y/n, but also native english y/n, so if user answering on wizard questins in english he can also answer in plain english on this question without switching to their native language. Native language answers still supported.
This is standart localization measure used in "apt" software for example.

### DIFF
--- a/cmd/incus/remote.go
+++ b/cmd/incus/remote.go
@@ -500,9 +500,9 @@ func (c *cmdRemoteAdd) run(cmd *cobra.Command, args []string) error {
 			}
 
 			if string(line) != digest {
-				if len(line) < 1 || strings.ToLower(string(line[0])) == i18n.G("n") {
+				if len(line) < 1 || strings.ToLower(string(line[0])) == i18n.G("n") || strings.ToLower(string(line[0])) == "n" {
 					return errors.New(i18n.G("Server certificate NACKed by user"))
-				} else if strings.ToLower(string(line[0])) != i18n.G("y") {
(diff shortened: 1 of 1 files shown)


kazan417:
While improving usability i found serious typo in mkdir code. Incorrect parametr in api call for granting permission.

bensmrs:
Some languages have their “yes” translation starting with “n” (notably Greek, although it’s technically not the same Unicode code point, which is not a reasonable argument), and some have their “no” translation starting with “y” (notably Uzbek, which the Uzbek government is pushing to write in the Latin alphabet).
I don’t think it’s reasonable to check the letters as you’re doing. A way out would be to check the localized “y”, “n“, and only then check for the English letters; this way you don’t end up messing with other languages, but it feels VERY fragile to me.

bensmrs:
We could also let translators provide a comma-separated list of safe localizations (+ English if they believe it’s safe), and parse that list here. That’s the cleanest approach IMO.

stgraber:
The 0o750 is also not a typo, it's the correctly way to pass octal values to Go.

It was a long time ago (back in LXD) but I do remember wondering what to do with the translation of the y/n type questions and did reach the same conclusion as @bensmrs that we cannot assume that languages won't have them flipped in their translations and that it's far better to have a script fail because it's not running under LC_ALL=C.UTF-8 (as scripts really ought to) than have situations where user confirmation prompts do the exact opposite of what the user intended.

PR #3409, opened 2026-05-27
Add support for static round-robin remotes

### LINKED ISSUE #3023: Add support for round-robin cluster remotes

When adding a cluster member as a remote, it only adds that particular server to the remote list, so if it becomes offline, we need to add a new remote pointing to another cluster member.

Support for so-called cluster remotes would be interesting. What I think should be added is:
- A way to indicate to the client that it’s the desired behavior, either (exclusive) by
  - Adding a `--cluster` flag, indicating that we are pointing to all cluster remotes
  - Making cluster detection automatic so that this behavior becomes the default
- A way to store this information, so essentially making the client configuration file accept a list of remotes, maybe also storing the last valid server so that the round-robin isn’t too dumb (this last idea also has the nice property to allow us to reuse the `addr` field and have another field handling the list)
- A way to update the cluster remote list, either (non-exclusive)
  - Explicitly by adding an `update` subcommand to `incus remote`
  - Implicitly by having `incus cluster list` update said list
- (maybe not desirable) A way to prevent some members to be included in the list, either (non-exclusive) by:
  - Allowing to specify an inclusion/exclusion list to `incus remote add`
  - Allowing to include/exclude specific groups in `incus remote add`

bensmrs:
~~Related-to: #3023~~
Closes: #3023


bensmrs:
(will only push the i18n templates after the other PRs are merged)

bensmrs:
After digging for a while, I think DNS-based retries should be implemented at very low level within Incus, i.e. in `DoHTTP`, but it may require to use a custom dialer, as Go only seems to handle protocolar errors up to TLS. In any case, reading Go’s documentation, the dialer already tries several addresses if necessary, so I think the second part of https://github.com/lxc/incus/issues/3023#issuecomment-4049218809 is already there.
So this should close #3023 if you squint well enough.

stgraber:
@bensmrs needs a `make i18n`

bensmrs:
Yeah I was waiting for the other PRs to land to avoid rebasing several times. Pushing now.

stgraber:
@bensmrs sorry, my branch broke the translations, can you do a quick rebase?

bensmrs:
sure

bensmrs:
Oh, not only the translations, as you added a command :)

PR #3418, opened 2026-05-28
Push operation follow-up

bensmrs:
Related-to: #3329

I also got bored with the massive storage volume file.


bensmrs:
I propose here to make `--uid` and `--gid` overrides recursive, because I cannot see a reasonable reason not to.

stgraber:
@bensmrs sorry, the translations broke things again, can you rebase?

bensmrs:
np

PR #3420, opened 2026-05-28
Allow for specific VM vCPU configuration

stgraber:


shared/validate/validate.go
```diff
@@ -953,6 +953,54 @@ func IsValidCPUSet(value string) error {
...
+// Any of the three fields may be omitted, in which case it defaults to 1.
+// It returns the number of sockets, cores (per socket) and threads (per core).
+func ParseCPUTopology(value string) (int, int, int, error) {
+	sockets := 1
+	cores := 1
+	threads := 1
+
+	seen := map[string]bool{}
+	for _, chunk := range strings.Split(value, ",") {
+		fields := strings.SplitN(chunk, "=", 2)
```
bensmrs: Maybe use `util.SplitNTrimSpace`

stgraber: oh, oops, I keep forgetting about the magic helper :)

bensmrs: erm, wrong place?

bensmrs: Doing this on the `=` actually does the cleaning for the `,` at the same time (for the first and last element(s))
Doing this on the `,` doesn’t clean everything you want

stgraber: Doing it on both now for good measure. Doing it just on the latter should work, but it feels like a cleaner pattern to have the first split take care of spaces before/after the comma separate values and then the second one cleans up anything odd around the =.

bensmrs: I like hunting for a few CPU cycles :)

bensmrs (approved):

shared/validate/validate.go
stgraber: That's what compilers are for ;)

PR #3421, opened 2026-05-28
incusd/devices: Cleanup leftover forkproxy on startup

### LINKED ISSUE #3415: forkproxy helpers leak (orphan to PID 1, never reaped) after incusd is killed out-of-cgroup while a proxy-device container later stops

### Incus system details

Captured from a fresh Hetzner cx23 Ubuntu 24.04 host running incus
7.0.0 via Zabbly (also confirmed on incus 6.0.6 LTS on the same OS;
orphans from a 6.0.6 run survived an in-place upgrade to 7.0.0).
`api_extensions` and the storage driver list are trimmed for brevity;
everything else is verbatim.

```yaml
config: {}
api_extensions:
  - ... (~480 entries, trimmed)
api_status: stable
api_version: "1.0"
auth: trusted
public: false
auth_user_name: root
auth_user_method: unix
environment:
  addresses: []
  architectures:
    - x86_64
    - i686
  certificate: |
    -----BEGIN CERTIFICATE-----
    (trimmed)
    -----END CERTIFICATE-----
  certificate_fingerprint: 9724c289a9fe4b2b1497038dc9e704221efe42d2651f264727e24c1031a8979f
  driver: lxc
  driver_version: 7.0.0
  firewall: nftables
  kernel: Linux
  kernel_architecture: x86_64
  kernel_features: {}
  kernel_version: 6.8.0-117-generic
  lxc_features: {}
  os_name: Ubuntu
  os_version: "24.04"
  project: default
  server: incus
  server_clustered: false
  server_event_mode: full-mesh
  server_name: bubble-cloud
  server_pid: 3832
  server_version: 7.0.0
  storage: dir
  storage_version: "1"
  storage_supported_drivers:
    - name: dir
      version: "1"
      remote: false
    - ... (btrfs, lvm, truenas trimmed)
```

### Current behavior

If incusd is killed without its cgroup (e.g. the kernel OOM killer
killing the single incusd PID rather than the whole unit), and a
container with a proxy device subsequently exits while incusd is
down, the `incusd forkproxy` helper for that container is orphaned
to PID 1 and is never reaped when incusd restarts. Helpers
accumulate one per (OOM-kill + container-exit) cycle, holding their
FDs / RSS forever.

Observed on a shared multi-user dev host: ~270 leaked helpers per
day, ~43 GiB resident over 8.5 days, until the host hit OOM and was
unusable.

### Expected behavior

After incusd restarts, it should detect any forkproxy helpers whose
target instance no longer exists and reap them, the same way that
graceful `Stop()` does via `killProxyProc()`.

### Steps to reproduce

Self-contained script. Run as root on a host with incus and a
default bridge. Confirmed reproducing on 6.0.6 and 7.0.0.

```bash
#!/usr/bin/env bash
# Reproducer: incusd does not reap orphaned `forkproxy` helpers after an
# ungraceful daemon restart, leaking one per (OOM-kill + container-exit)
# cycle. Confirmed on incus 6.0.6 and 7.0.0.
set -euo pipefail
ITERS="${1:-4}"
MARK="fpleak-$$"

count() {
    local n=0 p
    for p in $(pgrep -f "incusd forkproxy -- .*127.0.0.1:39[0-9][0-9][0-9]" 2>/dev/null); do
        [ "$(cat /proc/$p/comm 2>/dev/null)" = incusd ] && n=$((n+1))
    done
    echo "$n"
}

echo "incus: $(incus version | tr '\n' ' ')"
echo "kernel: $(uname -r)"
echo "start: forkproxy helpers in our port range = $(count)"
echo

for i in $(seq 1 "$ITERS"); do
    n="${MARK}-${i}"; port=$((39000 + i))
    incus launch images:ubuntu/24.04 "$n" >/dev/null 2>&1
    for _ in $(seq 1 30); do incus exec "$n" -- true 2>/dev/null && break; sleep 1; done
    incus config device add "$n" p proxy \
        listen="tcp:127.0.0.1:${port}" connect="tcp:127.0.0.1:${port}" bind=container >/dev/null
    sleep 1
    cpid=$(incus info "$n" | awk '/^PID:/{print $2}')

    # (1) OOM-style: SIGKILL only the main incusd process (not the cgroup,
    #     so the forkproxy child survives and reparents to PID 1).
    kill -9 "$(systemctl show -p MainPID --value incus)" 2>/dev/null || true
    sleep 1
    # (2) container exits while incusd is down (incusd never sees the stop).
    kill -9 "$cpid" 2>/dev/null || true
    # (3) bring incusd back.
    systemctl start incus 2>/dev/null || true
    for _ in $(seq 1 30); do incus info >/dev/null 2>&1 && break; sleep 1; done
    sleep 2
    echo "iter $i: container exited while incusd was down -> leaked forkproxy helpers = $(count)"
done

echo
echo "Leaked orphans (PPID=1, referencing now-dead container init PIDs):"
pgrep -f "bin/incusd forkproxy -- .*127.0.0.1:39[0-9][0-9][0-9]" \
  | while read -r p; do ps -o pid,ppid,stat,etimes,args --no-headers -p "$p"; done
echo
echo "Cleanup: kill leaked orphans (incusd won't) + delete any leftover containers."
pgrep -f "bin/incusd forkproxy -- .*127.0.0.1:39[0-9][0-9][0-9]" | xargs -r kill -9 2>/dev/null || true
incus list -c n --format csv 2>/dev/null | grep "^${MARK}-" | xargs -r -n1 incus delete -f 2>/dev/null || true
```

Captured output from a 4-iteration run on the host described above
(incus 7.0.0, kernel 6.8.0-117-generic, Ubuntu 24.04):

```
incus: Client version: 7.0.0 Server version: 7.0.0
kernel: 6.8.0-117-generic
start: forkproxy helpers in our port range = 0

iter 1: container exited while incusd was down -> leaked forkproxy helpers = 1
iter 2: container exited while incusd was down -> leaked forkproxy helpers = 2
iter 3: container exited while incusd was down -> leaked forkproxy helpers = 3
iter 4: container exited while incusd was down -> leaked forkproxy helpers = 4

Leaked orphans (PPID=1, referencing now-dead container init PIDs):
   4546       1 Ssl      179 /opt/incus/bin/incusd forkproxy -- 4402 3 tcp:127.0.0.1:39001 3832 4 tcp:127.0.0.1:39001   0644
   5262       1 Ssl      132 /opt/incus/bin/incusd forkproxy -- 5113 3 tcp:127.0.0.1:39002 4692 4 tcp:127.0.0.1:39002   0644
   5969       1 Ssl       85 /opt/incus/bin/incusd forkproxy -- 5822 3 tcp:127.0.0.1:39003 5398 4 tcp:127.0.0.1:39003   0644
   6640       1 Ssl       37 /opt/incus/bin/incusd forkproxy -- 6541 3 tcp:127.0.0.1:39004 6108 4 tcp:127.0.0.1:39004   0644
```

Each orphan has PPID=1, comm `incusd`, and the third argv field is
the PID of the now-dead container init (4402, 5113, 5822, 6541 —
none of which exist any more). The orphans are never reaped, even
though incusd is restarted between iterations.

stgraber:
If a container shuts down or gets killed while incusd is down, the forkproxy processes get reparented to PID1 and remain there forever. Add logic to detect that and clean them up.

Closes #3415

### DIFF
--- a/cmd/incusd/devices.go
+++ b/cmd/incusd/devices.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"bytes"
+	"errors"
 	"fmt"
 	"os"
 	"path"
@@ -21,6 +23,7 @@ import (
 	"github.com/lxc/incus/v7/internal/server/state"
 	"github.com/lxc/incus/v7/shared/logger"
 	"github.com/lxc/incus/v7/shared/resources"
+	"github.com/lxc/incus/v7/shared/subprocess"
 	"github.com/lxc/incus/v7/shared/util"
 )
 
@@ -675,6 +678,72 @@ func devicesRegister(instances []instance.Instance) {
 		}
 
 		inst.RegisterDevices()
+	}
+}
+
+// cleanupOrphanedProxyHelpers reaps forkproxy helpers left running for stopped containers.
+func cleanupOrphanedProxyHelpers(instances []instance.Instance) {
+	for _, inst := range instances {
+		// Only containers spawn forkproxy helpers; proxy devices on VMs only support NAT mode.
+		if inst.Type() != instancetype.Container {
+			continue
+		}
+
+		// Running instances still need their helpers; only consider stopped instances.
+		if inst.IsRunning() {
+			continue
+		}
+
+		devicesPath := inst.DevicesPath()
+		if !util.PathExists(devicesPath) {
+			continue
+		}
+
+		for devName, devConfig := range inst.ExpandedDevices() {
+			if devConfig["type"] != "proxy" {
+				continue
+			}
+
+			pidPath := filepath.Join(devicesPath, fmt.Sprintf("proxy.%s", devName))
+			if !util.PathExists(pidPath) {
+				continue
+			}
+
+			p, err := subprocess.ImportProcess(pidPath)
+			if err != nil {
+				logger.Warn("Failed to import orphaned forkproxy helper, removing stale pid file", logger.Ctx{
+					"project":  inst.Project().Name,
+					"instance": inst.Name(),
+					"device":   devName,
+					"err":      err,
+				})
+
+				_ = os.Remove(pidPath)
+				continue
+			}
+
+			// Verify the PID still belongs to a forkproxy helper before
+			// signalling it, otherwise PID reuse could result in killing
+			// an unrelated process.
+			cmdline, err := os.ReadFile(fmt.Sprintf("/proc/%d/cmdline", p.PID))
+			if err != nil || !bytes.Contains(cmdline, []byte("forkproxy")) {
+				_ = os.Remove(pidPath)
+				continue
+			}
+
+			err = p.Stop()
+			if err != nil && !errors.Is(err, subprocess.ErrNotRunning) {
+				logger.Warn("Failed to stop orphaned forkproxy helper", logger.Ctx{
+					"project":  inst.Project().Name,
+					"instance": inst.Name(),
+					"device":   devName,
+					"pid":      p.PID,
+					"err":      err,
+				})
+			}
+
+			_ = os.Remove(pidPath)
+		}
 	}
 }
 
--- a/cmd/incusd/daemon.go
+++ b/cmd/incusd/daemon.go
@@ -1497,6 +1497,9 @@ func (d *Daemon) init() error {
 		// This should come after the event handler go routines have been started.
 		devicesRegister(instances)
 
+		// Reap any forkproxy helpers left behind by an out-of-cgroup kill of the previous daemon.
+		cleanupOrphanedProxyHelpers(instances)
+
 		// Setup seccomp handler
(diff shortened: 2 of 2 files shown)


cmd/incusd/devices.go
```diff
@@ -678,6 +681,72 @@ func devicesRegister(instances []instance.Instance) {
...
+
+				_ = os.Remove(pidPath)
+				continue
+			}
+
+			// Verify the PID still belongs to a forkproxy helper before
+			// signalling it, otherwise PID reuse could result in killing
+			// an unrelated process.
+			cmdline, err := os.ReadFile(fmt.Sprintf("/proc/%d/cmdline", p.PID))
+			if err != nil || !bytes.Contains(cmdline, []byte("forkproxy")) {
```
bensmrs, line 729: Maybe a finer match with `incusd\0forkproxy\0` or something like a split on `\0` then additional verification? (dunno if the daemon is guaranteed to be called `incusd`).
This is extremely unlikely, so feel free to discard.

bensmrs (approved):

PR #3427, opened 2026-05-28
Fix a couple of issues with instance memory changes

stgraber:


bensmrs (approved):

bensmrs:
Random thought, can’t QEMU still be given 0 bytes if we pass `n*blockSize-2`, then override with `n*blockSize-1`, hence plugging 0 blocks?

PR #3441, opened 2026-06-01
Align node specific network config

breml:


internal/server/db/storage_pools.go
```diff
@@ -884,6 +884,7 @@ func (c *ClusterTx) RemoveStoragePool(ctx context.Context, poolName string) (*ap
 // NodeSpecificStorageConfig lists all storage pool config keys which are node-specific.
 func NodeSpecificStorageConfig(driverName string) []string {
 	configKeys := []string{
+		"size",
```
presztak: @breml I'm not sure about this change as here: https://github.com/lxc/incus/blob/6255b3956027aaa6c2092ed42869f58e8d1942f5/internal/server/db/storage_pools.go#L896-L898 , we add `size` for all drivers besides `lvmcluster` but for that one `size` is cluster wide not node specific.

breml: Good catch, I will remove `size` again. I simply looked at the data generated in `internal/server/metadata/configuration.json`, where `scope` is set to `local`. For the logic in I need to maintain in Operations Center, it is a bit problematic, if a setting with the same name can be `global` or `local` depending on the entity type. I will have to look into this.

breml: @presztak I checked again the description in https://github.com/lxc/incus/blob/main/internal/server/storage/drivers/driver_lvm.go#L684 and there it says, that `size` has `scope == "local"`. I wonder, if this is a bug then, which should be fixed.
cc: @stgraber

breml: But this is for `lvm` and not for `lvmcluster`. 🤔

presztak: Maybe we should have two different definition. One for `lvm`, second one for `lvmcluster` if this make sense. But I guess that would look weird in doc.

breml: I stumbled over the exact same question. I try to implement https://github.com/FuturFusion/operations-center/issues/246, but the configuration metadata does not contain any information about `storage_lvmcluster`, so it would not behave correctly. In order to rely on the information from `/1.0/metadata/configuration`, it would need to include the definitions for `storage_lvmcluster` as well.

stgraber:
@breml @presztak `lvmcluster` should have its own set of keys. Ideally we should have a syntax to handle all the identical ones so we don't need to duplicate the comments, but for automation like through Operations Center and the web UI, it's critical that each driver has its own metadata.

PR #3458, opened 2026-06-05
incusd: Replace gorilla/mux with http.ServeMux

stgraber:
Apply what we did to incus-agent but to incusd.

Closes #2451


bensmrs (approved):
Maybe put the `go.{mod,sum}` changes in a dedicated commit

PR #3461, opened 2026-06-06
Various bugfixes

stgraber:


internal/server/auth/oidc/oidc.go
```diff
@@ -194,8 +194,25 @@ func (o *Verifier) Callback(w http.ResponseWriter, r *http.Request) {
...
 
+// cookieExpiry returns the cookie expiration, preferring the provider's refresh token lifetime over a 30 day default.
+func cookieExpiry(tokens *oidc.Tokens[*oidc.IDTokenClaims]) time.Time {
+	// Fall back to 30 days when the provider doesn't report a refresh token lifetime.
+	lifetime := 30 * 24 * time.Hour
+
+	if tokens != nil && tokens.Token != nil {
+		expiresIn, ok := tokens.Extra("refresh_expires_in").(float64)
+		if ok && expiresIn > 0 {
+			lifetime = time.Duration(expiresIn) * time.Second
```
bensmrs, line 210: That’s just a nit, but that time can be strictly greater than the time remaining (because getting to this check takes time), so removing a second could be a bit safer.
Also, can’t `expiresIn == 0` be possible, in which case we’d be wrongly reporting 30 days? (I really don’t know).

hallyn (approved):

internal/server/auth/oidc/oidc.go
stgraber: Hmm, I'll add a bit more safety around that. I don't think there's an actual issue if the logic goes wrong compared to the very wrong current handling, but that's no reason not to make it even better ;)

stgraber: From looking around expiresIn == 0 would normally be used to indicate an infinite expiry, so having it go through our fallback to 30 days is fine.

I'll add logic to deduct 1s if the expiry is > 1s so we have the other case correctly handled.

stgraber: And done.

PR #3463, opened 2026-06-06
Implement server.crt update through internal API

stgraber:


cmd/incusd/api_internal.go
```diff
@@ -333,6 +342,53 @@ func internalShutdown(d *Daemon, r *http.Request) response.Response {
...
+
+	cert, err := localtls.KeyPairFromRaw(certBytes, keyBytes)
+	if err != nil {
+		return response.BadRequest(fmt.Errorf("Invalid certificate or key: %w", err))
+	}
+
+	err = internalUtil.WriteCert(s.OS.VarDir, "server", certBytes, keyBytes, nil)
+	if err != nil {
+		return response.SmartError(fmt.Errorf("Failed to write new server certificate: %w", err))
+	}
```
bensmrs, line 392: This can be racy (but not the line below)

cmd/incus/admin_update_server_certificate.go
```diff
@@ -0,0 +1,110 @@
...
+
+type cmdAdminUpdateServerCertificate struct {
+	global *cmdGlobal
+}
+
+var cmdAdminUpdateServerCertificateUsage = u.Usage{u.Placeholder(i18n.G("cert.crt")), u.Placeholder(i18n.G("cert.key"))}
+
+func (c *cmdAdminUpdateServerCertificate) command() *cobra.Command {
+	cmd := &cobra.Command{}
+	cmd.Use = cli.U("update-server-certificate", cmdAdminUpdateServerCertificateUsage...)
```
bensmrs: I feel like if we have `incus cluster update-certificate`, this should be `incus admin update-certificate`; any particular reason why you want `-server` explicitly?

stgraber: It was to try and make it clear that this isn't what you want in clusters, but with the extra mention in the help and the error redirecting you to the other command, I think we can do away with that :)

cmd/incusd/api_internal.go
stgraber: What race are you worried about here?

I can see two things we may or may not care about:

1) One of the file fails to write for "reasons" and we end up with a weird mix of .crt and .key which won't work. We could guard against that by making WriteCert do Create+Rename rather than direct edits, allowing us to do a .new, .bak dance where we have a better chance at reverting.

2) Having the API be called twice concurrently. Given how lightweight it is, that may be a bit of a challenge and it would require you run it twice with different input for it to be a problem.

bensmrs: I’m worried about 2 concurrent cert/key writes, where we could end up with cert A and key B

stgraber: Okay, easy enough to grab a lock so we can do that. I feel like the actual window for a race there would be extremely hard to hit, but definitely not impossible :)

stgraber:
@bensmrs that should do it

bensmrs:
Ack, will review in ~2h

bensmrs (approved):

PR #3464, opened 2026-06-06
Add logging helper for deferred functions and use in all packages

### LINKED ISSUE #2049: Log unhandled error when defering a function that returns an error

If running static code analysis over the codebase i would like if only relevant places are shown where errors are ignored. 
Currently there are a lot of places that defer a function like `Close()` or similar that may return an error.
This leads to a lot of false positives when analysing the code base and hides actual relevant issues where errors are currently ignored.
I have already found 2 relevant places where the result of such errors being ignored #2040 by manually sifting through all of them, but there might be more.

So im suggesting something like a `utils.WarnOnError(func() error)` to log such errors as a warning (which i think is not bad for trouble shooting errors in general).

Also i would be fine with implementing this. I just wanted to ask if this is desired, since it seems like a bunch of effort and i'd rather not do it if its unlikely to get merged.

stgraber:
Closes #2049


stgraber:
This one will almost certainly lead to a bunch of extra noise in the logs which we'll have to silence. I just made sure that `git grep 'defer func()'` returned no hits by sending anything that matched through the new helper.

cmd/incus/file.go
```diff
@@ -573,7 +573,7 @@ func (c *cmdFilePull) pull(parsedFiles []*u.Parsed, target string) error {
 					return err
 				}
 
-				defer func() { _ = f.Close() }() // nolint:revive
+				defer logger.WarnOnError(f.Close, "Failed to close file") // nolint:revive
```
bensmrs, line 576: Is the `nolint` directive still relevant?

bensmrs (approved):

cmd/incus/file.go
stgraber: Yeah, it's not related to the lack of error checking but to the use of a `defer` from within a loop.

PR #3469, opened 2026-06-09
Update debug.go

kazan417:
fix term in the description


stgraber:
Rebased, fixed and added the translation commit.

Note that the initial fix was wrong as it was generalizing the NBD command to refer to "instances" when it's a command which works only on virtual machines. So the needed fix was actually to remove mentions of "instance" and consistently use "virtual machine".

PR #3475, opened 2026-06-09
Various bugfixes

stgraber:
Closes #3476


cmd/incus/cluster.go
```diff
@@ -1460,7 +1460,9 @@ func (c *cmdClusterEvacuate) command() *cobra.Command {
 	cmd.Aliases = []string{"evac"}
 	cmd.Use = cli.U("evacuate", cmdClusterEvacuateRestoreUsage...)
 	cmd.Short = i18n.G("Evacuate cluster member")
-	cmd.Long = cli.FormatSection(color.DescriptionPrefix, i18n.G(`Evacuate cluster member`))
+	cmd.Long = cli.FormatSection(color.DescriptionPrefix, i18n.G(`Evacuate cluster member
+
+The action flag allows overriding the default server-side action ("cluster.evacuate" instance configuration option)`))
```
bensmrs, line 1465: That still requires me to go through the docs :)

internal/server/instance/drivers/qmp/monitor.go
```diff
@@ -12,10 +12,41 @@ import (
...
+	"golang.org/x/sys/unix"
+
 	"github.com/lxc/incus/v7/shared/logger"
 	"github.com/lxc/incus/v7/shared/util"
 )
 
+// dialUnix connects to a unix socket, reaching long paths through a short
+// /proc/self/fd reference to the parent directory.
+func dialUnix(path string) (*net.UnixConn, error) {
+	if len(path) <= 107 {
```
bensmrs: Is there any point in keeping the old logic?
Same for the other function.

stgraber: Hmm, no particularly good reason.

We have similar logic on device names where we only trigger the overflow logic when we're passed it, but that's also more user visible. 

In this case, I agree that using that logic unconditionally will ensure we catch problems earlier.

cmd/incus/cluster.go
stgraber: Sorry :)

Whenever possible I try to avoid the CLI directly mentioning the possible values of config keys given that those can vary between server versions or even configurations.

In the case of restore, there is no server side config for it, so nowhere to point people to, but in the case of evacuation, it's a long-ish list and the logic for the default value has change a few times, so I feel better sending folks to the actual source of truth.

Maybe we can do the fancy escape sequences to make those kind of references actual links that the user can click? I believe systemd uses that pretty extensively.

bensmrs: Never used those directly, but I feel like it should be part of a bigger revamping, maybe with an XDG scheme handler for `incus:`, and reusing the new fragment Incus extension. Those are ideas I’m throwing at you, but a discussion in a feature request could be useful. Resolving this for now.

internal/server/instance/drivers/qmp/monitor.go
stgraber: Changed the logic so we always use it now.

cmd/incus/cluster.go
stgraber: Yeah. I've never used them myself either but they seem pretty handy. Not sure about using our own scheme vs just linking to https://linuxcontainers.org/docs/XYZ, linking directly to the correct section or key over there.

stgraber:
I've pulled the change from @DarkressX in there and rebased both of the changes a bit to use a shared connection helper. Makes the change a bit more compact.

stgraber:
```
stgraber@orilla-d13:~$ incus project create this-is-a-very-long-project-name-to-test-unix-socket-bind-logic
Project this-is-a-very-long-project-name-to-test-unix-socket-bind-logic created
stgraber@orilla-d13:~$ incus project switch this-is-a-very-long-project-name-to-test-unix-socket-bind-logic
stgraber@orilla-d13:~$ incus profile show default --project default | incus profile edit default
stgraber@orilla-d13:~$ incus launch images:debian/13 this-is-a-very-long-instance-name-to-test-unix-socket-bind-logi --vm
Launching this-is-a-very-long-instance-name-to-test-unix-socket-bind-logi
stgraber@orilla-d13:~$ incus list 
+-----------------------------------------------------------------+---------+------+------------------------------------------------+-----------------+-----------+
|                              NAME                               |  STATE  | IPV4 |                      IPV6                      |      TYPE       | SNAPSHOTS |
+-----------------------------------------------------------------+---------+------+------------------------------------------------+-----------------+-----------+
| this-is-a-very-long-instance-name-to-test-unix-socket-bind-logi | RUNNING |      | fd42:8b9f:58e4:b6ac:1266:6aff:fe0d:a4d9 (eth0) | VIRTUAL-MACHINE | 0         |
+-----------------------------------------------------------------+---------+------+------------------------------------------------+-----------------+-----------+
stgraber@orilla-d13:~$ incus project delete -f this-is-a-very-long-project-name-to-test-unix-socket-bind-logic
Remove this-is-a-very-long-project-name-to-test-unix-socket-bind-logic and everything it contains (instances, images, volumes, networks, ...) (yes/no): yes
Project this-is-a-very-long-project-name-to-test-unix-socket-bind-logic deleted
stgraber@orilla-d13:~$ incus project switch default
```

bensmrs (approved):

PR #3483, opened 2026-06-12
doc/incus-cli: Add CLI configuration file reference

### LINKED ISSUE #3424: Add a subcommand to configure the client defaults

Edit the client defaults without editing manually the client configuration file.

1. The client configuration fields are not documented, so that definitely needs fixing
2. A subcommand should be added to edit the defaults in a way that’s easier to understand

augustodsgv:
Partially resolves #3424

Adds a reference documentation about the `config.yml` file

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -245,6 +245,11 @@ doc-serve:
 doc-spellcheck: doc
 	. $(SPHINXENV) ; python3 -m pyspelling -c doc/.sphinx/spellingcheck.yaml
 
+.PHONY: doc-spellcheck-incremental
+# doc: Check spelling errors on the documentation, building the documentation only
+doc-spellcheck-incremental: doc-incremental
+	. $(SPHINXENV) ; python3 -m pyspelling -c doc/.sphinx/spellingcheck.yaml
+
 .PHONY: doc-linkcheck
 # doc: Check broken links on the documentation
 doc-linkcheck: doc-setup
--- a/Makefile
+++ b/Makefile
@@ -404,4 +404,4 @@ dist: doc
 # doc: Show this help
 help:
 	@echo The following targets are supported:
-	@sed -En 's/^#\s*section\(([^)]*)\):\s*(.*)$$/\n\x1b[1m\1:\x1b[0m \2/p;/^\.PHONY:/{N;N;s/^\.PHONY:\s*([^[:space:]]+)\n#\s*doc(:\s*(.*)\n\1:\s*$$|!\s*(.*)\n\1:[^\n]*)/  \1!\3\4/p;s/^\.PHONY:\s*([^[:space:]]+)\s*\n#\s*doc:\s*(.*)\n\1:\s*(.+)$$/  \1!\2 (runs \3)/p}' Makefile | awk -F! '{printf "%-20s%s\n", $$1, $$2}'
+	@sed -En 's/^#\s*section\(([^)]*)\):\s*(.*)$$/\n\x1b[1m\1:\x1b[0m \2/p;/^\.PHONY:/{N;N;s/^\.PHONY:\s*([^[:space:]]+)\n#\s*doc(:\s*(.*)\n\1:\s*$$|!\s*(.*)\n\1:[^\n]*)/  \1!\3\4/p;s/^\.PHONY:\s*([^[:space:]]+)\s*\n#\s*doc:\s*(.*)\n\1:\s*(.+)$$/  \1!\2 (runs \3)/p}' Makefile | awk -F! '{if(NF<2)print$$1;else{s=$$1;if(length(s)%2)s=s" ";while(length(s)<28)s=s" .";print s" "$$2}}'
--- a/doc/client-config.md
+++ b/doc/client-config.md
@@ -1,0 +1,87 @@
+(client-config)=
+# CLI configuration file
+The Incus CLI client uses a configuration file to store environment data and modify its behavior. This documentation describes the file structure and its options.
+
+```{note}
+Changing this configuration file manually may leave the CLI in an inconsistent state, so you should only edit it via CLI commands such as [`incus remote`](incus_remote.md) and [`incus alias`](incus_alias.md).
+```
+
+## The configuration file
+By default, the Incus CLI will use the `$HOME/.config/incus/config.yml`, but it can also use an arbitrary file provided by the `INCUS_CONF` environment variable.
+
+If the file or its path don't exist, it will use the default:
+
+```yaml
+default-remote: local
+remotes:
+  images:
+    protocol: simplestreams
+    public: true
+    addr: https://images.linuxcontainers.org
+  local:
+    protocol: incus
+    public: false
+    addr: unix://
+aliases: {}
+defaults:
+  list_format: ""
+  console_type: ""
+  console_spice_command: ""
+  no_color: false
+```
+
+## `remotes`
+Incus remote information managed by [`incus remote`](incus_remote.md) commands is stored in the `remotes` section. Remotes may have the following configuration keys.
+
+### `addr`
+Address of the remote server. Accepts IPs, FQDNs and URLs. Multiple addresses can be provided, separated with commas `addr1,addr2,...`.
+
+### `last_working_address`
+Used by the CLI to register the last working address of a remote, if the remote has multiple addresses.
+
+### `auth_type`
+Method that the client will use to authenticate to the server API. Accepts `tls` and `oidc`.
+
+### `keepalive`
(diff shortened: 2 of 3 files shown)


bensmrs:
The spell checker isn’t too happy about that as you can see in the test logs (Tests / System is unrelated). Some of them should be solved by wrapping things in backticks.

augustodsgv:
Fixed those issues. Tbh it's been a pain to run the spellchecker locally because they take so long, so I've added a `-incremental` spellchecker in the makefile to speedup things a little.

bensmrs:
It should be a `.PHONY` target

doc/cli_configs.md
```diff
@@ -0,0 +1,87 @@
+(client-configs)=
```
bensmrs: I’d put a singular (same with the file name)

doc/cli_configs.md
```diff
@@ -0,0 +1,87 @@
...
+# CLI configuration file
+The Incus CLI client uses a configuration file to store environment data and modify it's behavior. This documentation describes the file structure and it's options.
+
+## The configuration file
+By default, the Incus CLI will use the `HOME/.config/incus/config.yml`, but it can also use an arbitrary file provided by the `INCUS_CONF` environment variable.
```
bensmrs: Maybe `$HOME` and not `HOME`?

doc/cli_configs.md
```diff
@@ -0,0 +1,87 @@
...
+
+If the file or it's path don't exist, it will use the default:
```
bensmrs:
```suggestion
If the file or its path don't exist, it will use the default:
```

doc/cli_configs.md
```diff
@@ -0,0 +1,87 @@
...
+aliases: {}
+defaults:
+  list_format: ""
+  console_type: ""
+  console_spice_command: ""
+  no_color: false
+```
+
+## `remotes`
+Incus remote information managed by [`incus remote`](incus_remote.md) commands is stored in the remote section. Remotes may have the following configurations.
```
bensmrs:
```suggestion
Incus remote information managed by [`incus remote`](incus_remote.md) commands is stored in the `remotes` section. Remotes may have the following configuration keys.
```

doc/cli_configs.md
```diff
@@ -0,0 +1,87 @@
...
+
+### `addr`
+Address of the remote server. Accepts IPs, FQDN and URLs. Multiple addresses can be provided, separated by commas `addr1,addr2,...`.
```
bensmrs:
```suggestion
Address of the remote server. Accepts IPs, FQDNs and URLs. Multiple addresses can be provided, separated with commas `addr1,addr2,...`.
```

doc/cli_configs.md
```diff
@@ -0,0 +1,87 @@
...
+
+### `last_working_address`
+Defines the first address the CLI will try using when a remote has multiple addresses. It's used by the CLI to give preference to an address that has more probability of success.
```
bensmrs: It’s not completely about probabilities, it’s rather the last address the CLI tried that worked. I feel like you’re describing an observable effect of what it does, rather than what it actually does.

doc/cli_configs.md
```diff
@@ -0,0 +1,87 @@
...
+
+### `auth_type`
+Method that the client will use to authenticate to the server API. Accepts `tls` and `oidc`.
+
+### `keepalive`
+Timeout in seconds that the `keepalive` feature will maintain the connection opened.
+
+### `project`
+Default project used by the CLI in that remote. It's defined by the [`incus project switch`](incus_project_switch.md) command. It's overwritten by the `--project` flag and the `INCUS_PROJECT` environment variable.
```
bensmrs:
```suggestion
Default project used by the CLI when using that remote. It's defined by the [`incus project switch`](incus_project_switch.md) command. It's overwritten by the `--project` flag and the `INCUS_PROJECT` environment variable.
```

doc/cli_configs.md
```diff
@@ -0,0 +1,87 @@
...
+
+### `protocol`
+The protocol the CLI will use to communicate with the remote. This defines if the remote will be used as an Incus or image server. Accepts the following options:
+
+- `incus`: regular Incus server accessible over the network to provide remote access or images
```
bensmrs:
```suggestion
- `incus`: regular Incus server accessible over the network to provide access to instances or images
```

doc/cli_configs.md
```diff
@@ -0,0 +1,87 @@
...
+
+- `oci`: [Open Container Initiative (OCI)](https://opencontainers.org/) server that provides application container images
+
+- `public`: Incus server accessible over the network to provide images only
+
+- `simplestream`: [Simple Stream](https://git.launchpad.net/simplestreams/tree/) server that provide images
+
+### `credentials_helper`
+Defines an helper command that will handle OCI service authentication.
```
bensmrs:
```suggestion
Defines a helper command that will handle OCI service authentication.
```

doc/cli_configs.md
```diff
@@ -0,0 +1,87 @@
...
+
+### `public`
+Defines if the remote is an image only server (public).
```
bensmrs:
```suggestion
Defines if the remote is an image-only server (public).
```

doc/cli_configs.md
```diff
@@ -0,0 +1,87 @@
...
+
+## `default_remotes`
+Defines the default Incus remote used by the CLI to execute commands. It's defined by the [`incus remote switch`](incus_remote_switch.md) command. This configuration is overwritten by the environment variable `INCUS_REMOTE` or by the `<remote>:` prefix.
+
+## `aliases`
+Stores the alias managed by the [`incus alias`](incus_alias.md) command. Each alias is a key value pair that consists of the alias name and the alias command (enclosed by quotes).
```
bensmrs:
```suggestion
Stores the alias managed by the [`incus alias`](incus_alias.md) command. Each alias is a key/value pair that consists of the alias name and the alias command (enclosed by quotes).
```

doc/cli_configs.md
```diff
@@ -0,0 +1,87 @@
...
+
+## `defaults`
+Determine the default CLI behavior for a certain CLI commands.
```
bensmrs:
```suggestion
Determines the default CLI behavior for certain CLI commands.
```

Your other descriptions used the 3rd person

bensmrs (commented):
A few nits, the rest is nice, thanks!
Any reason why you separated list items with blank lines?

doc/cli_configs.md
augustodsgv: In this protocol type the incus remote can manage other resources than instances, like volumes, cluster members etc., is there a way of describing all of those actions? Like "access to the Incus server", or "access to the Incus cluster".

bensmrs: I’d say we don’t want to get too boring. “regular private Incus server” could also work; in the end, we just want something users can quickly understand.

bensmrs:
Maybe we’d also want a quick sentence saying that there is no guardrail when editing the file directly, and that it should rather be modified through the CLI commands.

doc/cli_configs.md
augustodsgv: I see, I was trying to describe the purpose of this configuration by describing one of it's advantages. What do you think about "It's used by the CLI to give preference to an address that has already worked", or remove this phrase at all?

augustodsgv:
Thanks for the review!

> A few nits, the rest is nice, thanks! Any reason why you separated list items with blank lines?

I miss understood the lint message that asked to separate list _blocks_ with blank spaces, just fixed it here.

doc/cli_configs.md
bensmrs: The purpose is just that “it worked before, so we should try to use it now”. Honestly, the nature of this key is in its title, so I wouldn’t spend too much time on finding the perfect phrasing :)

bensmrs:
I’m doing a quick patch as your new make target breaks `make help` a little.

bensmrs:
@stgraber if you feel like reviewing a dirty awk one-liner…
Else I think this is good to merge (apart from the CRIU website doing its usual thing).

augustodsgv:
Gosh, that's an obscure sed command lol. How did you identify this issue?

bensmrs:
> Gosh, that's an obscure sed command lol. How did you identify this issue?

It’s just that the target name was too long so I had to increase the padding. But then it looked ugly so I made that a little more pleasant.

bensmrs:
But yeah, for your sanity, better not spend too much time on that one-liner ;)

bensmrs (approved):

PR #3486, opened 2026-06-12
incus/server/storage/driver/ceph: Shrink images to minimal size after unpacking

DarkressX:
Currently all images will have a default block size of `10GiB` if nothing else is configured in the `volume.size` pool option.
This causes VMs with disk smaller than `10GiB` to not be able the get cloned from that image. Instead it will be copied and unpacked again, which means higher storage usage and slower VM creation.

This PR implements a shrinking of the downloaded image, to the smallest possible size ceph allows, after the image has been filled.

### DIFF
--- a/internal/server/storage/drivers/driver_ceph_volumes.go
+++ b/internal/server/storage/drivers/driver_ceph_volumes.go
@@ -93,10 +93,7 @@ func (d *ceph) CreateVolume(vol Volume, filler *VolumeFiller, op *operations.Ope
 				return err
 			}
 
-			// If the cached volume size is different than the pool volume size, then we can't use the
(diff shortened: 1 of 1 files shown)


bensmrs:
LINSTOR would benefit a lot from this also.

DarkressX:
I just saw that the `volume.size` allows overriding the DefaultBlockSize of 10GiB. That would technically already allow for having smaller base images. But I sill think this addition would be beneficial as it makes storage usage more efficient and as of now I can't find any downsides (other than having to do the shrinking of course).

@bensmrs unfortunately I have not worked with LINSTOR before. I would probably need some time to set it up and test around with it. Would it be acceptable to do that in a separate PR?

bensmrs:
Of course :)

PR #3491, opened 2026-06-13
Implement per-instance BGP

### LINKED ISSUE #2707: Announce individual instance IP addresses over BGP

In a small home-cloud with mikrotik as bgp peer all incus servers promote same /24 subnet via bgp. Possible to announce /32 (every vm ip)?

to announce /32 (every vm ip)

stgraber:
Closes #2707

### DIFF
--- a/internal/server/device/device_utils_network.go
+++ b/internal/server/device/device_utils_network.go
@@ -721,6 +721,12 @@ func bgpAddPrefix(d *deviceCommon, n network.Network, config map[string]string) 
 		}
 	}
 
+	// Advertise the instance's own addresses if enabled on the network.
+	err := bgpAddInstancePrefixes(d, n, config, map[uint]net.IP{4: nexthopV4, 6: nexthopV6}, bgpOwner)
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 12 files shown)


internal/server/device/device_utils_network.go
```diff
@@ -730,15 +736,185 @@ func bgpRemovePrefix(d *deviceCommon, config map[string]string) error {
...
+		}
+
+		prefixLen := 128
+		if ipVersion == 4 {
+			prefixLen = 32
+		}
+
+		// If a static address is configured on the NIC, advertise it directly.
+		staticAddr := config[fmt.Sprintf("ipv%d.address", ipVersion)]
+		if staticAddr != "" && staticAddr != "none" {
```
bensmrs: Use IsNoneOrEmpty

internal/server/device/device_utils_network.go
```diff
@@ -730,15 +736,185 @@ func bgpRemovePrefix(d *deviceCommon, config map[string]string) error {
...
+
+	if len(scanVersions) == 0 {
+		return nil
+	}
+
+	// Determine the MAC address to track in the neighbor table.
+	hwaddrStr := config["hwaddr"]
+	if hwaddrStr == "" {
+		hwaddrStr = d.volatileGet()["hwaddr"]
+	}
```
bensmrs: As that pattern seems pretty common, we may want to refactor it in a helper function, maybe not in this PR, though

internal/server/device/device_utils_network.go
```diff
@@ -730,15 +736,185 @@ func bgpRemovePrefix(d *deviceCommon, config map[string]string) error {
...
+				}
+
+				bgpInstanceScansMu.Unlock()
+			}
+		}
+
+		// Check immediately, then on every tick until the timeout expires.
+		check()
+
+		timeout := time.NewTimer(10 * time.Second)
```
bensmrs: Does that mean that the instance has to boot in only 10 seconds?
Seems short for BSDs

bensmrs (changes requested):

internal/server/device/device_utils_network.go
bensmrs: Additional checks every 10 seconds up to 30s or a minute total would be nice, if the check is too costly to keep 1s ticks.

stgraber: Yeah, for instances that don't have a static entry (ipv4.address/ipv6.address), we wait up to 10s for ARP/NDP. In my quick tests here, that's generally been possible even for VMs (trivial for containers) but maybe we can be a bit more generous and do 10s for containers and 30s for VMs?

Would 30s be sufficient for the BSD images?

bensmrs: Should be more than fine even on my macOS images I think.
Windows boots fast these days, right?

stgraber: Yeah, 30s will be long enough for Windows too.

stgraber:
@bensmrs done

Changes:
 - Made use of IsNoneOrEmpty across the network logic (I found some more cases of it)
 - Introduce and used a new helper to handle the fallback to volatile keys
 - Bumped learning time to 30s for VMs

bensmrs (approved):

PR #3493, opened 2026-06-13
doc: Ignore criu.org link checking

bensmrs:
The CRIU website has proven quite unreliable in terms of uptime.


stgraber:
@mihalicyn FYI. It's been going dead a few times a week for a while now, not sure if there's anything you can do to make it more reliable.

PR #3497, opened 2026-06-14
incus: Close web UI probe response body

fwerkor:
The web UI command only checks the status code returned by the UI endpoint probe before starting the local proxy. Close the response body once the status code has been checked so the probe does not leave the HTTP response open while the command keeps running.

Testing:
- go test ./cmd/incus


stgraber:
Thanks!

PR #3499, opened 2026-06-15
incus: Add default command

### LINKED ISSUE #3424: Add a subcommand to configure the client defaults

augustodsgv:
Resolves #3424

Adds new `incus default` command for managing CLI defaults with `set`, `show` and `unset` subcommands


bensmrs:
Sorry for the delay, I’m having a look at your PR now

bensmrs:
Nice work! May I ask you to conform a little more to the existing set/unset commands, @augustodsgv?
We support `set key1=value1 key2=value2...` and `unset key1 key2...` syntaxes, support setting things to the empty string to actually unset keys, and allow piping from stdin. Luckily for you, all of that already is implemented. Look at the other commands, like in `project.go`, but use `u.KV` instead of `u.LegacyKV`.
After that, I’ll have a more thorough review.

augustodsgv:
Hum, indeed it's not following the other incus "set" commands. I'm going to adjust it and submit the fix. Thanks for the response.

augustodsgv:
Hello! I was busy these days, but I'm working on this feature again.

I'm re-implementing it based on the `incus cluster set` and `incus profile set` commands, and both commands include the `--property` flag, but I didn't completely understood how it works. For what I tested, it's used for defining object specific configuration, such as profile and cluster descriptions. For other configurations such as profile `limits.cpu` it doesn't care if the flag is used or not. So my question is: what is a considered a parameter in the context of the CLI? And should this `default set` command implement it?

Finally, two question about go best practices (code below).

First, trying to make this command cleaner I've used some helper commands that validate the input and modify the `cmd.global.conf` object. Is there any concern about it, like modifying a copy of the object, etc.?

Second, I've found it cleaner to use an `err` variable that's checked only after the `switch-case` statement. Is it ok, or should I `if err != nil` every `err` inside the `cases`?

```go
func (c *cmdDefaultSet) set(cmd *cobra.Command, parsed[]*u.Parsed) error{
	conf := c.global.conf
	// Get the default keys
	keys, err := kvToMap(parsed[0])
	if err != nil {
		return err
	}
	for k := range keys {
		// Set default value, if default name is valid
		v := keys[k]
		var err error
		switch k {
		case "list_format":
			err = c.setListFormat(v)
		case "console_type":
			err = c.setConsoleType(v)
		case "console_spice_command":
			err = c.setConsoleSpiceCommand(v)
		case "no_color":
			err = c.setNoColor(v)
		default:
			err = fmt.Errorf("Invalid default %q", k)
		}
		if err != nil {
			return err
		}
	}
	return conf.SaveConfig(c.global.confPath)
}

func (c *cmdDefaultSet) run(cmd *cobra.Command, args []string) error {
	parsed, err := c.global.Parse(cmdDefaultSetUsage, cmd, args)
	if err != nil {
		return err
	}

	return c.set(cmd, parsed)
}

func (c *cmdDefaultSet) setListFormat(listFormat string) error {
	conf := c.global.conf

	// Validate list format
	switch listFormat {
	case cli.TableFormatCSV, cli.TableFormatJSON, cli.TableFormatTable, cli.TableFormatYAML, cli.TableFormatCompact, cli.TableFormatMarkdown, "":
	default:
		return fmt.Errorf("Invalid value %q for list_format", listFormat)
	}

	conf.Defaults.ListFormat = listFormat
	return nil
}

func (c *cmdDefaultSet) setConsoleType(consoleType string) error {
	conf := c.global.conf

	// Validate console type
	switch consoleType {
	case "console", "vga", "":
	default:
		return fmt.Errorf("Invalid value %q for console_type", consoleType)
	}

	conf.Defaults.ConsoleType = consoleType
	return nil
}

func (c *cmdDefaultSet) setConsoleSpiceCommand(consoleSpiceCommand string) error {
	conf := c.global.conf

	conf.Defaults.ConsoleSpiceCommand = consoleSpiceCommand
	return nil
}

func (c *cmdDefaultSet) setNoColor(noColorStr string) error {
	conf := c.global.conf
	
	// Replace blank value to allow no_color=""
	if noColorStr == "" {
		noColorStr = "false"
	}

	// Validate no color input
	noColor, err := strconv.ParseBool(noColorStr)
	if err != nil {
		return errors.New("no_color must be boolean")
	}

	conf.Defaults.NoColor = noColor

	return nil
}
```

bensmrs:
> I'm re-implementing it based on the `incus cluster set` and `incus profile set` commands, and both commands include the `--property` flag, but I didn't completely understood how it works. For what I tested, it's used for defining object specific configuration, such as profile and cluster descriptions. For other configurations such as profile `limits.cpu` it doesn't care if the flag is used or not. So my question is: what is a considered a parameter in the context of the CLI? And should this `default set` command implement it?

`--property` is here to set configuration parameters outside of the `config` map. For example, if you look at your profile configuration, you will see that `description` lives outside of `config`; you would need to use `--property` to set it.
This is unrelated to this PR, so just skip that flag.

> Finally, two question about go best practices (code below).
> 
> First, trying to make this command cleaner I've used some helper commands that validate the input and modify the `cmd.global.conf` object. Is there any concern about it, like modifying a copy of the object, etc.?

Well, that depends on the code, the question is a bit too abstract. Although, the client structs are short-lived, as the CLI is invoked from scratch at each command, so the risk is really only local to the commands you are implementing.

> Second, I've found it cleaner to use an `err` variable that's checked only after the `switch-case` statement. Is it ok, or should I `if err != nil` every `err` inside the `cases`?
> 
> ```go
> 		switch k {
> 		case "list_format":
> 			err = c.setListFormat(v)
> 		case "console_type":
> 			err = c.setConsoleType(v)
> 		case "console_spice_command":
> 			err = c.setConsoleSpiceCommand(v)
> 		case "no_color":
> 			err = c.setNoColor(v)
> 		default:
> 			err = fmt.Errorf("Invalid default %q", k)
> 		}
> 		if err != nil {
> 			return err
> 		}
> ```

Looks fine by me. Don’t forget to use `gofumpt`, as I think you’re missing a blank line.

stgraber:
Taking a quick look into this one given we're about to tag 7.2 and it seems most of the way there.

stgraber (approved):

augustodsgv:
I was about to submit the new changes, but I understand it was taking longer than it should. Sorry for the delay.

PR #3501, opened 2026-06-15
storage/zfs: avoid raw send for snapshot copies

### LINKED ISSUE #3046: incus copy fails with "permission denied" on encrypted ZFS when source dataset is not a clone

### Incus system details

```yaml
Incus 6.0.6
zfs 2.4.0
Linux 6.12
```

### Instance details

_No response_

### Instance log

_No response_

### Current behavior

When copying an instance that has snapshots on an encrypted ZFS storage pool, `incus copy` fails with `permission denied` if the source dataset is not a ZFS clone (i.e. has no `origin`).

```
Error: Create instance from copy: Failed to mount "pool/incus/virtual-machines/test-vm" on
"/var/lib/incus/storage-pools/zfs-test/virtual-machines/test-vm" using "zfs": permission denied
```

The underlying `mount(2)` syscall returns `EACCES` because the received dataset has `keystatus=unavailable`.

### Root cause

When the source has snapshots, `CreateVolumeFromCopy` in `driver_zfs_volumes.go` (line 674) uses `zfs send -R -w | zfs receive` instead of `zfs clone`.

The `-w` (raw) flag preserves encrypted blocks as-is. When the source dataset is **not a clone** (no `origin` property), ZFS sends a full stream with embedded encryption key metadata. On receive, ZFS creates a **new encryption root** with `keystatus=unavailable`, rather than inheriting encryption from the parent dataset.

Incus then tries to mount the received dataset without loading its encryption key first, and `mount(2)` returns `EACCES`.

When the source **is** a clone (has an `origin`), `zfs send -R -w` sends an incremental stream from the origin. In that case, the received dataset inherits the parent's encryption root and the key is already loaded. This is why VMs created directly on ZFS (which are clones of the image) can be copied successfully.

When does a dataset have no origin?

- Instance moved from a different storage backend (e.g. btrfs → ZFS)
- Maybe: instance received via migration ?
- Maybe: Clone that was promoted (`zfs promote`) ?

### Expected behavior

No error.

Potential fixes:

1. Don't use `-w` (raw) for local copies within the same pool. Since source and destination share the same encryption root, a non-raw send/receive would let ZFS decrypt and re-encrypt with the parent's key automatically. The `-w` flag is only needed for cross-pool or remote transfers.

2. Use individual incremental sends instead of `-R -w`. Send each snapshot incrementally without the replication flag, avoiding the full-stream-with-own-encryption-root problem.

### Steps to reproduce

```bash
# 1. Create an encrypted ZFS dataset and add it as an Incus storage pool
zpool create testpool /dev/sdX
echo "passphrase" | zfs create -o encryption=aes-256-gcm -o keyformat=passphrase \
  -o keylocation=prompt -o mountpoint=legacy testpool/incus
incus storage create zfs-test zfs source=testpool/incus

# 2. Create a btrfs storage pool as default
incus storage create default btrfs size=8GiB
incus profile device add default root disk path=/ pool=default

# 3. Create a VM on btrfs, then move it to ZFS
incus launch images:debian/13 my-vm --vm --storage default
incus stop my-vm
incus move my-vm my-vm --storage zfs-test

# 4. Verify no origin (non-clone)
zfs get origin testpool/incus/virtual-machines/my-vm
# VALUE = -

# 5. Create a snapshot
incus snapshot create my-vm snap0

# 6. Try to copy → fails
incus copy my-vm test-vm
# Error: Create instance from copy: Failed to mount ... permission denied
```

Without step 5 (no snapshot), the copy succeeds because Incus uses `zfs clone` instead of `zfs send`.

fwerkor:
## Summary
- avoid raw ZFS replication sends for same-pool snapshot copies unless recursive send is needed
- add a ZFS encryption regression test for copying a non-clone volume with snapshots

Fixes #3046

## Tests
- `go test ./internal/server/storage/drivers`
- `shellcheck -s bash test/suites/storage_driver_zfs.sh`
- `bash -n test/suites/storage_driver_zfs.sh`

### DIFF
--- a/internal/server/storage/drivers/driver_zfs_volumes.go
+++ b/internal/server/storage/drivers/driver_zfs_volumes.go
@@ -640,6 +640,86 @@ func (d *zfs) receiveFromZfsSender(vol Volume, sender *exec.Cmd) error {
 	return nil
 }
 
+func (d *zfs) useSnapshotSendChain(srcVol Volume) (bool, error) {
+	srcDataset := d.dataset(srcVol, false)
+
+	if d.needsRecursion(srcDataset) {
+		return false, nil
+	}
+
+	encryptionRoot, err := d.getDatasetProperty(srcDataset, "encryptionroot")
+	if err != nil {
+		// Older ZFS versions may not support encryption properties. Keep the existing
+		// replication stream path in that case.
+		return false, nil
+	}
+
+	if encryptionRoot == "" || encryptionRoot == "-" {
+		return false, nil
+	}
+
+	origin, err := d.getDatasetProperty(srcDataset, "origin")
+	if err != nil {
+		return false, err
+	}
+
+	return origin == "" || origin == "-", nil
+}
+
+func (d *zfs) copyVolumeSnapshotsUsingSend(vol Volume, srcVol Volume, srcSnapshot string, snapshots []string) error {
+	srcDataset := d.dataset(srcVol, false)
+	requested := map[string]struct{}{}
+	for _, snapshot := range snapshots {
+		requested[snapshot] = struct{}{}
+	}
+
+	entries, err := d.getDatasets(srcDataset, "snapshot")
+	if err != nil {
+		return err
+	}
+
+	sourceSnapshots := make([]string, 0, len(snapshots)+1)
+	for _, entry := range entries {
+		snapName, ok := strings.CutPrefix(entry, "@snapshot-")
+		if !ok {
+			continue
+		}
+
+		_, ok = requested[snapName]
+		if !ok {
+			continue
+		}
+
+		sourceSnapshots = append(sourceSnapshots, fmt.Sprintf("%s%s", srcDataset, entry))
+	}
+
+	if len(sourceSnapshots) != len(snapshots) {
+		return fmt.Errorf("Failed locating all ZFS snapshots for copy")
+	}
+
+	sourceSnapshots = append(sourceSnapshots, srcSnapshot)
+
+	var parentSnapshot string
+	for _, sourceSnapshot := range sourceSnapshots {
+		args := []string{"send"}
+		if parentSnapshot != "" {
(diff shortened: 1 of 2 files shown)


stgraber:
Tests are failing with:
```
Error: Create instance from copy: Failed ZFS receive: exit status 1 (cannot receive: failed to read from stream)
```

fwerkor:
Thanks, I found the issue. Dropping `-w` while keeping `-R` doesn't work for encrypted datasets, so the sender fails and the receiver only reports a broken stream.

I noticed newer ZFS has `-U`, but it doesn't seem available on the ZFS versions used by CI. Would you be OK with reworking this to avoid `-R` here and send the snapshots individually instead?

stgraber:
Where did you see a `-U` flag?

fwerkor:
Only in the OpenZFS master docs. I don't think it's relevant here since it doesn't seem to be in the stable versions Incus supports.

stgraber:
Ah, I see it now. Yeah, -U will be very convenient for us once it's actually available.

But we'll have to wait for it to be in an actual release and then add feature detection for it as we need to support several years of ZFS releases that won't have that flag.

stgraber:
So the logic of making `-w` optional here still causes a behavior regression for non-encrypted datasets. Per the ZFS manpage, `-w` on unencrypted datasets is equivalent to `-Lec`, but you are just unconditionally unsetting `-w` here, making those flags also disappear.

stgraber:
Anyway, sending the snapshots individually should be fine, again so long as we're not regressing things in the process :)

stgraber:
Can you separate the refactoring from the new logic?

So effectively one commit which just moves the existing logic to the new receive function and calls it from the original spot. Then a second commit adding the snapshot transfer logic?

That will make it easier to review and cherry-pick.

fwerkor:
Done, split into two commits now.

stgraber:
Ah and we usually split the tests into a separate commit too.

PR #3502, opened 2026-06-15
ci: authenticate OpenFGA release lookups

fwerkor:
The system test workflow currently queries the latest OpenFGA server and CLI releases through the GitHub API without authentication. In large matrix runs this can hit the unauthenticated API rate limit and fail during dependency setup before any Incus tests run.

This updates those release metadata lookups to use the workflow `GITHUB_TOKEN`, covering both the OpenFGA server and CLI downloads.

Thanks for maintaining Incus. This project is very useful to me.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -271,6 +271,8 @@ jobs:
         if: matrix.go == 'tip'
 
       - name: Install dependencies
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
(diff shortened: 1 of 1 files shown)


bensmrs:
Using the `2026-03-10` API could be good in the long run

fwerkor:
Done, thanks. Updated the OpenFGA release API calls to pass X-GitHub-Api-Version: 2026-03-10.

Issue #3507, opened 2026-06-16
`ProtocolOCI.GetImageAlias()` requires skopeo


jochumdev:
### Is there an existing issue for this?

- [x] There is no existing issue for this bug

### Is this happening on an up to date version of Incus?

- [x] This is happening on a supported version of Incus

### Incus system details

```
$ apt show skopeo
Version: 1.18.0+ds1-1+b5
```

```
$ apt show incus
Version: 1:7.0-debian13-202605301824
```

```
$ incus version
Client version: 7.1
Server version: 7.1
```

### Current behavior

```
$ incus launch docker.io:library/hello:latest test
Launching test
Error: Failed instance creation: Failed getting remote image info: Failed to run: skopeo --insecure-policy inspect docker://docker-registry.example.com/library/hello:latest --no-tags: exit status 2 (time="2026-06-16T20:51:08Z" level=fatal msg="Error parsing image name \"docker://docker-registry.example.com/library/hello:latest\": reading manifest latest in docker-registry.example.com/library/hello: manifest unknown")
```

### Expected behavior

No error, maybe no skopeo dependency at all for `launch`.

### Steps to reproduce

```
$ sudo apt remove skopeo # on the client only
$ incus launch docker.io:library/hello:test test
```

jochumdev:
Solving this issue would allow the windows client to work with OCI as well.

stgraber:
Hmm, actually, this whole issue looks wrong to me. The error you're getting here is a server-side error, not a client side one. The client is correctly pushing everything to the server here.

The issue is that the image definition you're passing isn't valid.

Example here from a Mac with no Skopeo available but against a remote server running IncusOS:
```
orilla:~ stgraber$ incus remote add docker.io https://docker.io --protocol=oci
orilla:~ stgraber$ incus launch docker.io:library/hello:latest test
Launching test
Error: Failed instance creation: Failed to run: skopeo --insecure-policy inspect docker://docker.io/library/hello:latest --no-tags: exit status 1 (time="2026-06-17T01:18:49Z" level=fatal msg="Error parsing image name \"docker://docker.io/library/hello:latest\": reading manifest latest in docker.io/library/hello: requested access to the resource is denied")

orilla:~ stgraber$ incus launch docker.io:hello-world:latest test --ephemeral --console
Launching test


Hello from Docker!
This message shows that your installation appears to be working correctly.

To generate this message, Docker took the following steps:
 1. The Docker client contacted the Docker daemon.
[...]

PR #3508, opened 2026-06-16
client: use a custom PoC oci registry client for GetImageAlias(), refs #3507

jochumdev:
This adds a PoC registry client to resolve images to fingerprints without the need of `skopeo` 

It is a PoC cause it is mostly written by Claude and not battle tested, will build a custom `incus-compose` client using this soon and test it.

### DIFF
--- a/client/oci.go
+++ b/client/oci.go
@@ -3,6 +3,8 @@ package incus
 import (
 	"errors"
 	"net/http"
+
+	"github.com/lxc/incus/v7/internal/oci"
 )
 
 // ProtocolOCI implements an OCI registry API client.
@@ -18,7 +20,19 @@ type ProtocolOCI struct {
 	// Error tracking for images.
 	errors map[string]error
 
+	// Lazily-initialized registry client (caches auth tokens across calls).
+	registry *oci.Registry
+
 	tempPath string
+}
+
+// getRegistry returns the registry client, creating it on first use.
+func (r *ProtocolOCI) getRegistry() *oci.Registry {
+	if r.registry == nil {
+		r.registry = oci.NewRegistry(r.httpHost, r.http, r.httpUserAgent)
+	}
+
+	return r.registry
 }
 
 // Disconnect is a no-op for OCI.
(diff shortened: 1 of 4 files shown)


stgraber:
Per CONTRIBUTING.md we do not currently allow LLM contributions to this repository.

We're currently looking at relaxing this (see https://github.com/lxc/incus/pull/3506) but even under the new relaxed policy this doesn't seem to fit the requirement of this having been human designed and having you be able to reason about the generated code.

Anyway, the implementation here is pretty clean and readable but if we're going down that road, I'd like skopeo to be completely gone. Doing it only halfway may cover your use case here but for us it just made testing and validation twice as complex :)

Validation for a replacement should at the very least include testing with:
 - DockerHub (using all the URLs/aliases supported by Skopeo)
 - GHCR
 - Quay
 - GAR
 - ECR
 - ACR
 - Artifactory

We'll also want to directly test a few of those using the authenticator logic as well as testing through an HTTPS proxy to make sure that everything goes through the proxy as expected.

@cyphar are you aware of any Go package that may already fit the bill here?
I remember us looking at just pulling out the registry logic from Skopeo but it really wasn't designed for use as a Go module.

jochumdev:
I found:

- [github.com/regclient/regclient](https://github.com/regclient/regclient) - pure go and light.
  Claude writes: explicitly library-first. Does manifest/blob get+put, image copy, and OCI-layout export (so it covers both the `inspect` and the `copy`/pull side), multi-platform, and reads `docker config`/credential helpers. Lightest dependency footprint of both.
  
- [github.com/google/go-containerregistry](https://github.com/google/go-containerregistry)
  Claude writes: the most battle-tested (crane/cosign/ko). `pkg/v1/remote` + `pkg/crane` cover pull/copy, `pkg/authn` does the credential-helper chain, and it has native cloud keychains: `pkg/v1/google` (GAR/GCR via ADC) and the ECR helper (`github.com/awslabs/amazon-ecr-credential-helper`) plugs straight into `authn`.

jochumdev:
Sorry, I should have read CONTRIBUTING.md before opening this. I missed that LLM contributions aren't currently accepted and that's on me; I didn't mean to put extra review burden on you.

PR #225, opened 2023-11-07
Detect btrfs compression

stgraber:

PR #3511, opened 2026-06-18
incusd/storage/drivers: Add "btrfs.compression" volume option

### LINKED ISSUE #3510: btrfs: allow forcing nodatacow on VM/block volumes when pool compression is enabled

I can't get nodatacow (`+C`) on my VM disk images because the btrfs pool sits on a compressed filesystem.

Incus normally sets `+C` on block volumes to avoid CoW fragmentation, but since #225 it skips that when it sees `compress` in the pool's mount options. The check in `CreateVolume`:

```go
if !slices.Contains(mountOptions, "datacow") && !strings.Contains(mountinfo[len(mountinfo)-1], "compress") {
    subprocess.RunCommand("chattr", "+C", volPath)
}
```

The catch is that compress is a filesystem-wide mount option on btrfs, not per-subvolume. My pool is a subvolume of my root fs, which is mounted `compress=zstd`. I want compression on the OS and data but nodatacow on the VM images, and there's no way to say that right now. Remounting just the pool subvolume without compress flips the whole filesystem.

`+C` itself works fine on a compressed fs, the file just isn't compressed, so this is really Incus being cautious rather than a btrfs limit. The workarounds are dropping compress and redoing it per-subvolume with btrfs properties, or putting the pool on its own filesystem, both of which are overkill for "compression everywhere, nodatacow on the VMs".

Some way to force it back on. #225 already added a `datacow` token to `btrfs.mount_options` for the opposite case, so this is just the missing direction.

I opened a PR (#3511) that adds a `btrfs.nocow` pool option. When it's set, `+C` gets applied to block volumes even on a compressed pool. Those volumes end up uncompressed, everything else stays compressed, off by default.

I went with a dedicated option rather than a `nodatacow` token in `btrfs.mount_options` because for subvolume pools the mount options get applied with a bind-remount, so a nodatacow token there would flip the entire host filesystem.

cmspam:
On btrfs Incus sets nodatacow (+C) on block volumes to avoid CoW fragmentation, but since #225 it skips that when the pool is compressed, since nodatacow and compression are mutually exclusive. There's no way to opt back in when the pool is a subvolume of a globally-compressed file system.

This adds a btrfs.compression volume option that maps to the btrfs compression property and takes the same values (zstd, lzo, zlib, none). Setting it to none turns off compression for the volume, which also lets nodatacow be applied even on a compressed pool. So you can keep compression on the rest of the file system and still get nodatacow on the VM disk images.

It's a volume option, so volume.btrfs.compression=none works pool-wide.

Closes #3510

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -3173,3 +3173,9 @@ again when the instance stops.
 This adds a new `core.https_allowed_websocket_origin` server
 configuration key. It can be set to a comma-separate list of allowed
 origins or to the `*` wildcard.
+
+## `storage_btrfs_compression`
+
+This adds a new `btrfs.compression` storage volume configuration key for
+the `btrfs` driver. It maps to the Btrfs `compression` property and takes
+the same values (for example `zstd`, `lzo`, `zlib` or `none`).
--- a/doc/config_options.txt
+++ b/doc/config_options.txt
@@ -6191,6 +6191,14 @@ Note that the claim must be contained in the access token.
 
 <!-- config group storage_truenas-common end -->
 <!-- config group storage_volume_btrfs-common start -->
+```{config:option} btrfs.compression storage_volume_btrfs-common
+:condition: "appropriate driver"
+:default: "same as `volume.btrfs.compression`"
+:shortdesc: "Compression algorithm to set on the volume, mapping to the Btrfs `compression` property (for example `zstd`, `lzo`, `zlib` or `none`)"
+:type: "string"
+
+```
+
 ```{config:option} initial.gid storage_volume_btrfs-common
 :condition: "custom volume with content type `filesystem`"
 :default: "same as `volume.initial.gid` or `0`"
--- a/doc/reference/storage_btrfs.md
+++ b/doc/reference/storage_btrfs.md
@@ -48,8 +48,9 @@ Therefore, you should never use VMs with Btrfs storage pools.
 
 If you really need to use VMs with Btrfs storage pools, set the instance root disk's [`size.state`](devices-disk) property to twice the size of the root disk's size.
 This configuration allows all blocks in the disk image file to be rewritten without reaching the qgroup quota.
-The [`btrfs.mount_options=compress-force`](storage-btrfs-pool-config) storage pool option can also avoid this scenario, because a side effect of enabling compression is to reduce the maximum extent size such that block rewrites don't cause as much storage to be double-tracked.
-However, this is a storage pool option, and it therefore affects all volumes on the pool.
+The per-volume `btrfs.compression` option can also help with this quota issue, as enabling compression reduces the maximum extent size so that block rewrites don't cause as much storage to be double-tracked.
+It does however prevent Incus from disabling copy-on-write on the volume, since the two are mutually exclusive.
+Set `btrfs.compression=none` to keep copy-on-write disabled.
 ```
 
 ## Configuration options
--- a/doc/rest-api.yaml
+++ b/doc/rest-api.yaml
@@ -64,12 +64,18 @@ definitions:
         description: Certificate represents a certificate
         properties:
             certificate:
-                description: The certificate itself, as PEM encoded X509 (or as base64 encoded X509 on POST)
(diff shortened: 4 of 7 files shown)


stgraber:
Hmm, I think calling that `nocow` when it's already the default behavior is a bit confusing/misleading. The actual impact here seems to be about compression handling.

Wouldn't it make more sense to have a `btrfs.compression` volume option which would then affect the `compression` inode option (`btrfs property set`)? Having it set to `none` would then allow for the `nodatacow` to be forced onto the volume.

That would basically be a `btrfs.compression` volume option which directly maps to `btrfs set property compression` and supports the same range of values. The default empty value would keep things as they are today, the `none` value would disable compression and so allow forcing the no copy on write behavior.

Volume options can be set globally on a pool through `volume.btrfs.compression=none` so it can be done pool-wide.

cmspam:
Yeah, I agree, that's much cleaner. I reworked it into a `btrfs.compression` volume option that maps straight to the compression property and takes the same values (zstd/lzo/zlib/none). Empty keeps things as they are, `none` disables compression on the volume and lets nodatacow be applied, and `volume.btrfs.compression=none` does it pool-wide. Dropped the `btrfs.nocow` option.

stgraber:
Can you update this for our usual commit structure?

This adds a configuration key, so you need a new API extension string and matching commit, then one commit to add the new config key and logic and then one to capture the result of `make update-metadata`.

cmspam:
Sorry about that. I've restructured it to match: one commit for the API extension, one for the config key and logic, and one for the `make update-metadata` output. Let me know if you'd like any other changes made.

stgraber:
The btrfs doc change needs some work. The note section as seen in https://linuxcontainers.org/incus/docs/main/reference/storage_btrfs/ is already getting a bit long, so let's try not to make it longer.

I think we should split the compression stuff into its own paragraph. Now that we do have a per-volume option, we should instead have a paragraph mention that setting `btrfs.compression` can help with the quota issue but also prevents disabling copy-on-write and that `btrfs.compression=none` can be used to keep copy on write disabled.

cmspam:
Sure. I made an attempt at a single paragraph. Please let me know what you think. Happy to make any other necessary adjustments.

stgraber:
Text looks better, thanks! I think that should still be within the note section though, just not as its own paragraph within it.

Can you also split out that storage_btrfs doc change to be its own commit?

cmspam:
Done. Let me know if there's anything else.

cmspam:
@stgraber I have to apologize for not testing this approach thoroughly enough until now, and not being knowledgeable enough about btrfs limitations to bring it up earlier.

For the `none` case, `property set ... compression none` sets the inode's no-compression flag (`m` in lsattr), and the kernel then refuses `chattr +C` on an inode that has `c` or `m`. So the property set and the `+C` fight each other and create aborts:

```bash
btrfs subvolume create /mnt/v
btrfs property set /mnt/v compression none   # lsattr -> ...m
chattr +C /mnt/v                             # chattr: Invalid argument
```

Setting an actual algorithm (zstd/lzo/zlib) is fine, it's only `none` that breaks. Reproduced with incusd off main.

The catch seems to be that enabling nodatacow specifically through the compression property doesn't work, since `none` lands on `m` rather than a clean inode. The mapping is still nice for zstd/lzo/zlib. If you'd rather keep it, the `none` path could special-case (skip the property, `chattr -c` then `+C`), or it might be cleaner to go back to a direct nocow toggle like the original `btrfs.nocow` for that one case since it never touches the property. Happy to rework however you see fit or drop it entirely, and sorry again for any wasted time.

stgraber:
Would likely be more intuitive to folks to wire `none` to do the chattr behind the scenes.

cmspam:
Sounds good, I'll work on it and prepare a follow-up PR.

cmspam:
Opened #3540 for this. It wires `none` to do the chattr like you suggested, and also fixes `volume.btrfs.compression`, which turned out not to work pool-wide for any value (the pool validator was passing `nil` instead of the volume rules).

PR #3524, opened 2026-06-24
doc/cloud-init: Change YAML spec domain

bensmrs:
yaml.org now has a Google tracker, and at the time of this PR, the spec links are broken. This deals with both issues.


stgraber:
Excellent, I was going to look into that one too as it's annoyed me on my OCI branch :)

PR #3529, opened 2026-06-25
Add version specific QEMU args according to `image.os` and `image.release`

### LINKED ISSUE #3195: Add support for OS-specific VM quirks
We currently have some limited quirks for when running Windows, MacOS and FreeBSD but we could benefit from extending that logic a bit with additional tricks to support older operating systems.

That's something that we have FuturFusion have spent quite a few weeks/months working through our work on Migration Manager where we need to be able to migrate existing VMs going as far back as RHEL 4, Debian 5 or Windows 2003 Server. All of those can be made to work mostly fine with Incus through a few `raw.qemu` tweaks.

Rather than keep all that logic in Migration Manager, it would make more sense to add it to Incus and more formally define our expectations around the `image.os` and `image.release` properties to trigger those quirks.

masnax:
Closes #3195

### DIFF
--- a/internal/server/instance/drivers/driver_qemu_machine.go
+++ b/internal/server/instance/drivers/driver_qemu_machine.go
@@ -16,6 +16,7 @@ import (
 	"github.com/lxc/incus/v7/internal/server/instance/drivers/qemudefault"
 	"github.com/lxc/incus/v7/internal/server/instance/instancetype"
 	"github.com/lxc/incus/v7/shared/osarch"
+	"github.com/lxc/incus/v7/shared/osinfo"
 	"github.com/lxc/incus/v7/shared/resources"
 	"github.com/lxc/incus/v7/shared/units"
 	"github.com/lxc/incus/v7/shared/util"
@@ -324,7 +325,7 @@ func (d *qemu) memoryTopology(bs *qemuBootState) (*qemuMemoryTopology, error) {
 	limitsMemoryHotplug := d.expandedConfig["limits.memory.hotplug"]
 	memoryHotplugEnabled := !util.IsFalse(limitsMemoryHotplug)
 
-	if d.GuestOS() == "freebsd" {
+	if d.GuestOS() == osinfo.FreeBSD {
 		memoryHotplugEnabled = false
 
 		// We handle the empty value a bit differently here, as FreeBSD doesn’t have memory hotplug.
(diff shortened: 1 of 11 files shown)


shared/osinfo/windows.go
```diff
@@ -0,0 +1,76 @@
+package osinfo
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Supported windows versions, with matching virtio-win abbreviations.
+var windowsVersions = map[string]string{
+	"XP":             "xp",
```
gibmat, line 11: Also include Windows 7, since I know it's still popular?

masnax: Yeah, might as well complete the set with 7, 8, and 8.1. To be honest most of these aren't used by Incus directly except for XP and Server 2003, which are the only ones that seem to need the additional global qemu args.

internal/server/instance/drivers/driver_qemu_machine.go
```diff
@@ -400,3 +400,34 @@ func (d *qemu) memoryTopology(bs *qemuBootState) (*qemuMemoryTopology, error) {
 
 	return memInfo, nil
 }
+
+func (d *qemu) osVersionSpecificOptions() []string {
```
stgraber: @masnax I'd expect all of those to be supported as qemu.conf configs and not need to be handled as command line arguments. In general we try to limit what we pass as command line arguments to options that QEMU doesn't expose through qemu.conf or QMP.

stgraber: Looked into it a bit and it should be possible to make the function return `[]cfg.Section` which get appended to the full config.

CPU handling (qemu64) should likely move to cpuType() to avoid passing the option twice.

The memory cap should go into memoryTopology() and be effectively a new maximum, so limits.memory.hotplug=false should still reduce it down to the limits.memory value.

masnax: Actually, I narrowed down the kernel panics related to the CPU type to just the `smap` flag, so we don′t even need to force `qemu64`, we can just set `-smap` on whichever CPU type is chosen according to migration state and cluster group config.

I′ll have to test on a few VMs (just looked at Oracle 6 so far) to be sure, but that would be a cleaner solution.

stgraber: Ah, that's nice!

stgraber (approved):

PR #3531, opened 2026-06-25
incusd/endpoints: Fix infinite loop in network error log writer

### LINKED ISSUE #3494: incusd --group incus-admin --logfile /var/log/incus/incusd.log uses a full core

### Incus system details

```yaml
Latest
```

### Instance details

_No response_

### Instance log

_No response_

### Current behavior

I noticed in htop there was

incusd --group incus-admin --logfile /var/log/incus/incusd.log that was burning 100% continuously.

I tried some debugging with an llm, but didn't get to to be bottom of it. 

I restarted the incus daemon and it went away.

Dunno if it matters but last night I did do some network scans on the server, and there was a note or two about
evel=warning msg="Rejecting request from untrusted client"

My LLM also said:
There is also a second issue in the goroutine dump: many goroutines are stuck for hundreds of minutes on Go’s log.(*Logger).output mutex via net/http.(*Server).logf. That suggests Incus’ HTTP server logging path is wedged or heavily contended, likely from broken/rejected connections or websocket/API handling.

### Expected behavior

Incus shouldn't be burning a core continuously.

### Steps to reproduce

Don't have

stgraber:
A log line starting with 'h' but not 'http:', for example 'http2:' would cause an infinite loop.

Closes #3494

### DIFF
--- a/internal/server/endpoints/network_util.go
+++ b/internal/server/endpoints/network_util.go
@@ -27,10 +27,9 @@ func (d networkServerErrorLogWriter) Write(p []byte) (int, error) {
 
 func (d networkServerErrorLogWriter) stripLog(p []byte) string {
 	// Strip the beginning of the log until we reach "http:".
-	for len(p) > 5 && string(p[0:5]) != "http:" {
-		p = bytes.TrimLeftFunc(p, func(r rune) bool {
-			return r != 'h'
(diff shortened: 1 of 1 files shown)


bensmrs (approved):
What the actual fuck

PR #3534, opened 2026-06-25
Use OS-specific CLI paths

stgraber:


shared/cliconfig/file.go
```diff
@@ -13,28 +13,72 @@ import (
...
 	if os.Getenv("INCUS_CONF") != "" {
-		configDir = os.Getenv("INCUS_CONF")
-	} else if os.Getenv("HOME") != "" && util.PathExists(os.Getenv("HOME")) {
-		configDir = filepath.Join(os.Getenv("HOME"), ".config", "incus")
-	} else {
+		return os.Getenv("INCUS_CONF"), nil
+	}
+
+	// Use the platform-specific user configuration directory.
+	// (~/.config on Linux, ~/Library/Application Support on macOS, %AppData% on Windows)
```
bensmrs, line 24: Ah, I’m not sure what to think about that one for macOS. Sure, that *should* be the correct directory, but `.config` is used a lot by CLI applications inherited from the Linux world. A symlink at `.config` pointing to the canonical directory would surely be very useful for macOS users. At least that’s my feeling as a user.

cmd/incus/main.go
```diff
@@ -438,6 +438,59 @@ If you already added a remote server, make it the default with "incus remote swi
...
 }
 
+func getCachePath() (string, error) {
+	// Honor an explicit override.
+	if os.Getenv("INCUS_CACHE") != "" {
+		return os.Getenv("INCUS_CACHE"), nil
+	}
+
+	// Use the platform-specific user cache directory.
+	// (~/.cache on Linux, ~/Library/Caches on macOS, %LocalAppData% on Windows)
```
bensmrs, line 448: For this one though, I don’t think a symlink would be useful.

shared/cliconfig/file.go
stgraber: Sure. I can add a symlink as part of the migration logic.

So new installs would go to the canonical directory but existing installs would have the symlink so folks don't get lost if they're used to messing with it directly (hopefully not).

cmd/incus/main.go
stgraber: Yeah, agreed. We don't really want folks messing too much with .config but sometimes it has to be done. For .cache, there's really no reason to :)

shared/cliconfig/file.go
stgraber: Added the symlink now

bensmrs (approved):
There may be an issue now on macOS if users remove the new config dir, as we’d be moving the symlink to its self-referencing location. I don’t care enough to make this a blocker, though.

Issue #2702, opened 2025-11-26
Routed NIC as Managed Network


ibot3:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

We are using routed NICs. This means for each VM, we need to configure a network device like that:

```
devices:
  enp5s0:
    ipv4.address: <ipv4_without_cidr>
    ipv6.address: <ipv6_without_cidr>
    mtu: '1500'
    nictype: routed
    type: nic
    vrf: <vrf>
```

It's not very userfriendly as there are no checks for the correct IP range or an already used IP-address.
It's also a security risk, because we need to allow non-managed network devices and wrong values can be supplied.

So it would be great so have the option to create a managed network with the type routed, where all the parameters (like vrf or routing table id) can be configured for all Instances using the network.
And additionally that v4/v6 subnet pools can be supplied, where IP-addresses are taken from and tracked via IPAM.
(It should still be possible to pick an address yourself from the pool)

And it would be nice if the creation of an instance with routed NIC would work in the UI, without doing custom YAML things.

### What do you think would need to be added

I am not deep enough into the tech-stack to know what would be required to do and how complicated it is.

stgraber:
In general, I agree that this would be quite nice and useful. The implementation however may get pretty tricky and would likely result in many more folks trying out routed NICs without understanding what else is at play.

Specifically, getting the allocation side of things handled would still need manual work to then configured the address inside of the VM.

I feel like there is quite a lot more we could/should do around routed NICs leading to having this issue be eventually resolved.

ibot3:
A big plus would be if we could get DHCP to work with routed NICs.
However at least dnsmasq doesn't support that.
Maybe Kea DHCP would work, but this would be a bigger change.

This would allow using routed NICs without manual cloud-init network config.

stgraber:
We were looking into an in-process DHCP server a while back for that use case but that's a lot of code and work just for this :)

damex:
i do use it with bird that filters what can be announced and what can not.

maybe routed network could act as a source for default values and hold some prefix(es) ?

eg 10.11.12.0/24{32,32} is in network and device could get a pass if they annouce 10.11.12.1/32 or 10.11.12.111/32 but will get rejected if it is out of prefix bounds like 10.11.13.1/32.

PR #3547, opened 2026-06-26
feat: Add `bgp.ovn.active_chassis_only` config option

ibot3:
When enabled, each incus node only announces prefixes for the ovn networks that also have their active ovn gateway on that node. 

Also generally disables pushing ovn prefixes with no uplink to BGP (as there is never a valid next-hop).

###  Why?

We have a setup where our incus nodes do not have a shared L2 connection but we would still want to use OVN networks with uplink.
My plan was to configure the same transit uplink network on each node, but its not connected between the nodes.
Currently all nodes announce all OVN prefixes. However, this requires that the Uplink network is shared between all nodes, which isn't the case in our setup.

This MR adds the option that only the active OVN chassis announces the OVN prefix.

I am also open to other suggestions on how we can solve our problem.
(We are currently using routed NICs, but this doesn't work well with Kubernetes CAPI, etc. as it requires quite complex configuration of the incus instance, and routed NICs can't be supplied by a managed network (see #2702))

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,5 @@
+export CGO_CFLAGS="-I/home/jakob/go/deps/raft/include/ -I/home/jakob/go/deps/cowsql/include/"
+export CGO_LDFLAGS="-L/home/jakob/go/deps/raft/.libs -L/home/jakob/go/deps/cowsql/.libs/"
+export LD_LIBRARY_PATH="/home/jakob/go/deps/raft/.libs/:/home/jakob/go/deps/cowsql/.libs/"
+export CGO_LDFLAGS_ALLOW="(-Wl,-wrap,pthread_create)|(-Wl,-z,now)"
+export PATH="${PATH}:$(go env GOPATH)/bin"
--- a/doc/config_options.txt
+++ b/doc/config_options.txt
@@ -4012,6 +4012,14 @@ User keys can be used in search.
 
 <!-- config group network_macvlan-common end -->
 <!-- config group network_ovn-common start -->
+```{config:option} bgp.ovn.active_chassis_only network_ovn-common
+:condition: "BGP server"
+:default: "`false`"
+:shortdesc: "Only announce the network's BGP prefixes from the cluster member hosting the active OVN gateway"
+:type: "bool"
+
+```
+
 ```{config:option} bridge.external_interfaces network_ovn-common
 :scope: "local"
 :shortdesc: "Comma-separated list of unconfigured network interfaces to include in the bridge"
--- a/doc/howto/network_bgp.md
+++ b/doc/howto/network_bgp.md
@@ -87,3 +87,14 @@ Set the following configuration options on the uplink network:
 
 Once the uplink network is configured, downstream OVN networks will get their external subnets and addresses announced over BGP.
 The next-hop is set to the address of the OVN router on the uplink network.
+
+### Announce OVN networks only from the active gateway
+
+By default, every cluster member announces an OVN network's prefixes, all using the OVN router's uplink address as the next-hop.
+This relies on the uplink network being a shared layer 2 segment between cluster members, so that the next-hop stays reachable wherever the active OVN gateway runs.
+
+To avoid that layer 2 requirement, set `bgp.ovn.active_chassis_only=true` on the OVN network.
+When enabled, only the cluster member currently hosting the active OVN gateway chassis announces the network's prefixes (network subnets, address forwards, and load balancers).
+On failover, the gaining member starts announcing and the losing member withdraws its prefixes automatically.
+
+This allows each cluster member to use an independent (layer 3 only) uplink: configure a local transit bridge with the same subnet on every member, and let a local routing daemon re-advertise the announced prefixes towards the fabric with next-hop-self.
--- a/internal/server/metadata/configuration.json
+++ b/internal/server/metadata/configuration.json
@@ -4487,6 +4487,15 @@
 			"common": {
 				"keys": [
 					{
+						"bgp.ovn.active_chassis_only": {
+							"condition": "BGP server",
+							"default": "`false`",
+							"longdesc": "",
+							"shortdesc": "Only announce the network's BGP prefixes from the cluster member hosting the active OVN gateway",
+							"type": "bool"
+						}
+					},
+					{
 						"bridge.external_interfaces": {
 							"longdesc": "",
 							"scope": "local",
--- a/internal/server/network/driver_common.go
+++ b/internal/server/network/driver_common.go
@@ -98,6 +98,10 @@ type common struct {
 	status      string
 	managed     bool
 	nodes       map[int64]db.NetworkNode
+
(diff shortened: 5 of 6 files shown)


ibot3:
I wanted to add tests for that scenario, but it seems that OVN isn't tested at all?
Should we create a separate issue for that?

stgraber:
OVN is tested but through daily tests run on Jenkins instead.
https://github.com/lxc/lxc-ci has the tests.

stgraber:
I wonder if it wouldn't be cleaner to have something like `ipv4.ovn.independent` and `ipv6.ovn.independent` to more clearly declare that setup and then have BGP react to it.

It should also be possible to just do it automatically in some scenarios. For example if the uplink is an Incus managed bridge (like `incusbr0`) rather than a `physical` type network, we can already assume that each server will be independent and that this behavior is then warranted.

ibot3:
I could imagine that even if you have a shared L2 uplink network, one could still want to enable that option to have the traffic routed more efficient (directly to the node having the gw).

So I would be fine with renaming it, but maybe "independent" wouldn't fit that good in the mentioned setup.

stgraber:
> I could imagine that even if you have a shared L2 uplink network, one could still want to enable that option to have the traffic routed more efficient (directly to the node having the gw).

That's not how this works :)

On a shared L2, all Incus servers advertise the EXACT same route over BGP, the next-hop is the virtual router's address on the uplink network so all traffic always flow through the server running that virtual router.

The reason for having those prefixes advertised by all servers is redundancy. If something goes wrong with the Incus daemon itself, the traffic will still flow correctly as the other active servers can still advertise it.

The only routes that we advertise only from specific servers are those tied to a resource that only exists on that server, specifically instances. That's because we can generally assume that a problem preventing that server from advertising those instance-specific routes may also be affecting said instance and that it probably shouldn't be getting traffic (particularly important in ECMP setups).

ibot3:
> On a shared L2, all Incus servers advertise the EXACT same route over BGP, the next-hop is the virtual router's address on the uplink network so all traffic always flow through the server running that virtual router.

But still, if the OVN Gateway of an OVN network resides at Node B and traffic for a VM (on Node A) in that network comes in over Node A (because advertised on all nodes), The flow is `Fabric -> Node A -> Node B -> Node A` which is one hop more than required.
However, the fix would be to give the route of the active chassis a better MED or similar so that is preferred by the fabric while still having redundancy.

So my point vanished and I think "independent" could be a fitting naming. 

But as you just mentioned it, the redundancy for the L3-only setup isn't that optimal, as it highly depends on the incus daemon (while the ovn on a node where the incusd crashed still works, so the prefixes aren't announced somewhere else).
It probably survives a usual restart (because I hope routes are not withdrawn then), but a longer downtime of the incusd would be problematic..
But maybe out of scope for this PR

stgraber:
> But still, if the OVN Gateway of an OVN network resides at Node B and traffic for a VM (on Node A) in that network comes in over Node A (because advertised on all nodes), The flow is `Fabric -> Node A -> Node B -> Node A` which is one hop more than required.

No it's not.

Say your uplink is 10.100.100.0/24 and your virtual router for the network got 10.100.100.10 and the network itself is 10.200.0.0/24, every server will then advertise:

`10.200.0.0/24 via 10.100.100.10`

Incus never advertises itself as the next hop, it advertises the IP address of the OVN router within the uplink network as the next hop. The router it advertises to sits on that L2 so the traffic correctly goes `Fabric -> Node running the logical router -> Node running the instance`.

stgraber:
```
stgraber@orilla-d13:~ (incus:s-dcmtl-cluster/routing)$ incus network list --project core default
+---------+------+---------+----------------+------------------------+------------------------------+---------+---------+
|  NAME   | TYPE | MANAGED |      IPV4      |          IPV6          |         DESCRIPTION          | USED BY |  STATE  |
+---------+------+---------+----------------+------------------------+------------------------------+---------+---------+
| default | ovn  | YES     | 10.65.122.1/24 | 2602:fc62:a:1000::1/64 | Network for regular services | 18      | CREATED |
+---------+------+---------+----------------+------------------------+------------------------------+---------+---------+
stgraber@orilla-d13:~ (incus:s-dcmtl-cluster/routing)$ incus exec frr01 bash
root@frr01:~# vtysh 

Hello, this is FRRouting (version 10.6.1).
Copyright 1996-2005 Kunihiro Ishiguro, et al.

frr01# show ipv6 route 2602:fc62:a:1000::1/64
Routing entry for 2602:fc62:a:1000::/64
  Known via "bgp", distance 200, metric 0, best
  Last update 5d13h46m ago
  Flags: Recursion iBGP Selected 
  Status: Installed 
  * 2602:fc62:a:100::100, via eth-uplink, weight 1
    2602:fc62:a:100::100, via eth-uplink (duplicate nexthop removed), weight 1
    2602:fc62:a:100::100, via eth-uplink (duplicate nexthop removed), weight 1

frr01# 
```

stgraber:
Or even more clearly:
```
frr01# show bgp ipv6 2602:fc62:a:1000::1/64
BGP routing table entry for 2602:fc62:a:1000::/64, version 1345772
Paths: (3 available, best #1, table default)
  Not advertised to any peer
  Local
    2602:fc62:a:100::100 from 2602:fc62:a:101::100 (10.0.0.100)
      Origin IGP, localpref 100, weight 1000, valid, internal, multipath, best (Router ID), rpki validation-state: invalid
      Last update: Sun Jun 21 03:11:28 2026
  Local
    2602:fc62:a:100::100 from 2602:fc62:a:101::101 (10.0.0.101)
      Origin IGP, localpref 100, weight 1000, valid, internal, multipath, rpki validation-state: invalid
      Last update: Sun Jun 21 03:25:10 2026
  Local
    2602:fc62:a:100::100 from 2602:fc62:a:101::102 (10.0.0.102)
      Origin IGP, localpref 100, weight 1000, valid, internal, multipath, rpki validation-state: invalid
      Last update: Sun Jun 21 03:52:15 2026
```

ibot3:
Ah sorry, you are totally right.
I was too deep in our layer-3 only setup where the fabric wouldn't be directly connected to the uplink network.

ibot3:
The hard dependency on incusd running for the networking to work made me think about everything again and I came to the conclusion that this wouldn't be a good solution.

To remove this dependency I would suggest another change:
Supporting OVNs [dynamic routing](https://docs.ovn.org/en/latest/topics/dynamic-routing/architecture.html#introduction) options in incus, so that the LR can be configured with these options:

`dynamic-routing="true", dynamic-routing-redistribute=connected,lb, dynamic-routing-v4-prefix-nexthop="<uplink addr>", dynamic-routing-vrf-id="<vrf-id>",`

This lets OVN directly install the active prefixes of a node into a selected routing table on the host.
A local Routing Daemon can then distribute the routes to the fabric.
So the routes are available, even if incus daemon doesn't run.

I already tested this locally in my test setup and can confirm it works, so I will spin up a new PR for that can then this one can probably be closed.

ibot3:
See #3557

PR #3548, opened 2026-06-26
Fix typo in documented behavior in setting `restricted.containers.privilege` to `isolated`

Hye-Dev:
This PR documents the behavior of setting `restricted.containers.privilege` to `isolated` on a project in the configurable options index.

**Additional Context:**
Recently, #3545 was merged, fixing a discrepancy in the configurable options index of the behavior of `security.idmap.isolated` when setting `restricted.containers.privilege` to `isolated` on a project.

This fix actually introduced a similar discrepancy to the documented behavior of `security.privileged`, stating `isolated` prevents `security.privileged` from being set to `false`, when in reality, it prevents `security.privileged` from being set to `true`.

### DIFF
--- a/cmd/incusd/api_project.go
+++ b/cmd/incusd/api_project.go
@@ -1660,7 +1660,7 @@ func projectValidateConfig(s *state.State, config map[string]string) error {
 		// Possible values are `unprivileged`, `isolated`, and `allow`.
 		//
 		// - When set to `unprivileged`, this option prevents setting {config:option}`instance-security:security.privileged` to `true`.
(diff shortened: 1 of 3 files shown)


Hye-Dev:
I am working on making a dictionary of normalized objects for all of the configurable options for Incus, which would be the obvious path forward for  a foundation of a broader re-structuring of the configurable options index, as well as preventing issues like this in the future.

At the same time, I do actually *like* how the configurable options are sort of defined in the code where they are relevant and automatically generated. Maybe it would be better to explore a superior way to locally define & generate these things so we can keep that simplicity (the in-progress dictionary could be used as a base for migrating that work when ready, however, it's massive so that won't be comprehensive for quite a while).

PR #3557, opened 2026-06-28
feat: Support OVN dynamic routing

ibot3:
Add two new config options to networks that can be uplink network for OVN networks: `ovn.dynamic_routing` and `ovn.dynamic_routing.vrf.id`

When configured, dynamic-routing [0] will be enabled on OVN networks using this uplink network. With that, prefixes of ovn networks and load-balancers are added to the configured host routing table (`vrf.id`) of the active OVN chassis.

[0] https://docs.ovn.org/en/latest/topics/dynamic-routing/architecture.html

### Why?

We have a setup where our incus nodes do not have a shared L2 connection, but all run a local routing daemon.
Currently, incus requires a shared L2 uplink network for OVN networks.
My plan was to configure an uplink bridge on each node, which is not directly connected to the fabric nor other nodes.
Instead, the local routing daemon distributes the routes coming directly from the OVN active gateway.

Using the bgp integration is no good option, as this would create a hard dependency on the incus daemon running.
(Only one host (=the active gateway) can announce the prefix)

### DIFF
--- a/doc/config_options.txt
+++ b/doc/config_options.txt
@@ -3860,6 +3860,23 @@ User keys can be used in search.
 ```
 
 <!-- config group network_bridge-common end -->
+<!-- config group network_bridge-ovn start -->
+```{config:option} ovn.dynamic_routing network_bridge-ovn
+:condition: "standard mode"
+:default: "`false`"
+:shortdesc: "Have downstream OVN networks export their prefixes into a host VRF on the active gateway chassis"
+:type: "bool"
+
+```
+
+```{config:option} ovn.dynamic_routing.vrf.id network_bridge-ovn
+:condition: "`ovn.dynamic_routing`"
+:shortdesc: "Routing table ID of the pre-created host VRF OVN exports into (required when `ovn.dynamic_routing` is enabled)"
+:type: "integer"
+
+```
+
+<!-- config group network_bridge-ovn end -->
 <!-- config group network_forward-common start -->
 ```{config:option} target_address network_forward-common
 :shortdesc: "Default target address for anything not covered through a port definition"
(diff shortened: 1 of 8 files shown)


ibot3:
I also took a look on how to test this in lxc-ci, but it seems that ovn comes from `https://ppa.launchpadcontent.net/stgraber/ovn-stable/ubuntu/` which does not have a recent enough ovn version.
I don't know why you chose not to use the ubuntu-shipped version, but even with that, we would need to switch to Ubuntu 26.04

ibot3:
Just discovered a bad caveat: 
When there are multiple OVN gateways on the same chassis with the same vrf id, only one prefix makes it into the routing table.
That limits the number of ovn networks per vrf to 1, which is quite frustrating.
I have to think about that.

stgraber:
We should move the daily tests to Debian 13 and using https://github.com/zabbly/ovn for the packages. That PPA is effectively dead at this point, we just haven't spent the time to update the tests yet.

stgraber:
@ibot3 any update on this one?

ibot3:
I am currently in the process of getting the change into OVN. 

But I also wanted to check if OVN has some option to allow multiple active gateway chassis for the same network, as this would give a better redundancy.

ibot3:
The fix is now upstream at OVN:
https://github.com/ovn-org/ovn/commit/92c7ab5d1cb6c70bbb1c9352215550b7db1a97cf

But it will probably land in OVN 26.09 and therefore not in the versions shipped by default by current Debian/Ubuntu releases.
@stgraber what do you think about the MR in general? Would you be happy with it?

I would suggest adding a note to the documentation describing the limitation with older (<26.09) OVN versions.

ibot3:
@stgraber Ping :)

stgraber:
I'd probably keep it simpler with just `ovn.routes_vrf` as the config option, if not set, it's disabled, if set to a VRF, it's enabled. Keeps the pattern mostly in line with what we already have with ingress_mode.

Otherwise, we'll want the branch split into our usual set of commits for a feature like this, basically:
 - API extension
 - OVN database addition (ovn_nb)
 - One commit per network driver
 - Commit for the doc tutorial
 - Commit to pick up the auto-generated config/metadata files

PR #3562, opened 2026-06-28
github: Add a build workflow

stgraber:
This builds "incusd" and "incus" on a self-hosted Debian 13 builder.

The reason for producing this artifact is so we can then pull it on IncusOS test systems. The goal is to slowly refresh our physical testing lab from Ubuntu deployed by MAAS to regular IncusOS boxes that can be factory reset and the new version side-loaded onto them for testing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,93 @@
+name: Build
+on:
+  push:
+    branches:
+      - main
+      - stable-*
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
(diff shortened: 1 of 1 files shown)


stgraber:
This PR has the workflow enabled on pull-request too. That will get disabled prior to merging but is needed to make sure this actually works on a self-hosted runner.

stgraber:
Links to working runs (as the PR now disabled it, keeping it just for on merge):
 - https://github.com/lxc/incus/actions/runs/28332351860/job/83932619686?pr=3562
 - https://github.com/lxc/incus/actions/runs/28332351860/job/83932619705?pr=3562

bensmrs (approved):

Issue #3563, opened 2026-06-28
DRM Native Context as a GPU option


cmspam:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

Incus already has a few GPU options for VMs (physical passthrough, mdev, mig, and sriov). They all hand a whole GPU, or a vendor slice of one, to a VM.

With qemu 11 and virglrenderer,  drm_native_context gives VMs full GPU acceleration (OpenGL, Vulkan, and video encode/decode) from a single GPU, without needing vendor SR-IOV or mdev support. I think it would be a good GPU option.

### What do you think would need to be added

Right now I set it up manually with the raw.qemu options:
- raw.qemu.conf adds a virtio-vga-gl device with blob resources and drm_native_context turned on.
- raw.qemu adds an egl-headless display.
- On the host I install virglrenderer and the QEMU virtio-gpu GL display module.
- In the guest I just use the normal mesa driver. On AMD (my iGPU) a stock userspace already works, and I get OpenGL, Vulkan, and VAAPI video inside the VM.

I think the qemu options would be useful to many people if integrated into an Incus GPU option.  It is quite easy to integrate, so I have prepared a draft PR, in case you'd like to integrate it.  The only downside is it requires virglrenderer and very recent qemu (11+) -- IncusOS and official zabbly builds maybe don't have virglrenderer, but virglrenderer isn't a big package or a difficult build, so I think it's a better candidate than virgl or venus which require full mesa, could be included easily, and I think a lot of people may want video transcoding or other GPU features in a VM without full passthrough.

PR #3564, opened 2026-06-28
Add a native-context GPU type for VMs (virtio-gpu DRM native context)

cmspam:
This implements the feature request in #3563: a new `native-context` option for the `gpu` device on VMs.

Incus already has GPU options that hand a whole GPU, or a vendor slice of one, to a VM (physical, mdev, mig, sriov). This adds a different kind: with QEMU 11 and virglrenderer, `drm_native_context` gives a VM full GPU acceleration (OpenGL, Vulkan, and video encode/decode) from a single GPU, with no vendor SR-IOV or mdev support, and the host keeps using the GPU. One GPU can be shared by several VMs.

I had been setting this up by hand with the raw.qemu options and it works well, so I thought it would be worth offering as a proper GPU option.

## Usage

```
incus config device add v1 gpu0 gpu gputype=native-context
```

You can optionally pick a card with the usual selectors (id, pci, vendorid, productid), and set the host-visible blob window with `blob.size` (default 2GiB). In the guest you just use the normal mesa driver.

## What it does

- A new `native-context` gputype (VM only). When attached, the default emulated GPU becomes a GL-capable virtio-gpu (`virtio-vga-gl` on x86, `virtio-gpu-gl-pci` elsewhere) with blob and drm_native_context, and QEMU is started with `-display egl-headless,gl=on[,rendernode=...]`. It is not a PCI passthrough, so the host keeps the GPU and it can be shared.
- Requires QEMU 11.0.0 or newer (checked, with a clear error otherwise).
- The AppArmor profile is granted access to the host DRM render nodes only when a native-context device is attached.

## Requirements

- virglrenderer on the host. IncusOS and the zabbly builds may not ship it today, but it is a small package and an easy build, and unlike virgl or venus it does not need host-side mesa at all. I think it is a good fit for people who want video transcoding or other GPU features in a VM without full passthrough.
- QEMU 11 or newer.

## Testing

Tested on an AMD Radeon 780M (Phoenix iGPU). A normal Arch VM with stock mesa got, all from the one host GPU:

- Vulkan via RADV (deviceName "AMD Radeon 780M Graphics (RADV PHOENIX)")
- VAAPI H264 and HEVC decode and encode
- OpenGL 4.6

I have only tested AMD. Intel should also work, since i915 already supports native context, though I have not tested it. Intel Xe also works with patches (https://github.com/cmspam/xe-native-context-enablement)

The change is split into the usual commits: the API extension, the device, the QEMU wiring, the AppArmor change, the reference doc, and the regenerated config metadata. Happy to adjust anything.

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -3256,3 +3256,12 @@ upgrade.
 
 ## `network_allocations_network`
 Adds the `network` field to the network allocations API response.
+
+## `gpu_native_context`
+
+This adds a new `native-context` `gputype` for `gpu` devices on virtual machines.
+Instead of passing through a PCI device, it gives the VM an accelerated `virtio-gpu`
+that uses DRM native context, so the guest's own driver drives the host GPU
+directly, with graphics and video acceleration. One host GPU can be shared by several
+VMs and the host keeps using it. The `blob.size` device option sets the
+host-visible blob window (default 2GiB). Requires QEMU 11.0.0 or newer.
--- a/doc/config_options.txt
+++ b/doc/config_options.txt
@@ -333,6 +333,44 @@ With VMs, this option supports mounting file system disk devices and paths withi
 ```
 
 <!-- config group devices-gpu_mig end -->
+<!-- config group devices-gpu_native_context start -->
+```{config:option} blob.size devices-gpu_native_context
+:default: "2GiB"
+:required: "no"
+:shortdesc: "Size of the host-visible blob memory window for the `virtio-gpu` device"
+:type: "string"
+
+```
+
+```{config:option} id devices-gpu_native_context
+:required: "no"
+:shortdesc: "The DRM card ID of the GPU device"
+:type: "string"
+
+```
+
+```{config:option} pci devices-gpu_native_context
+:required: "no"
+:shortdesc: "The PCI address of the GPU device"
+:type: "string"
+
+```
+
+```{config:option} productid devices-gpu_native_context
+:required: "no"
+:shortdesc: "The product ID of the GPU device"
+:type: "string"
+
+```
+
+```{config:option} vendorid devices-gpu_native_context
+:required: "no"
+:shortdesc: "The vendor ID of the GPU device"
+:type: "string"
+
+```
+
+<!-- config group devices-gpu_native_context end -->
 <!-- config group devices-gpu_physical start -->
 ```{config:option} gid devices-gpu_physical
 :default: "0"
--- a/doc/reference/devices_gpu.md
+++ b/doc/reference/devices_gpu.md
@@ -15,6 +15,7 @@ The following types of GPUs can be added using the `gputype` device option:
 - [`mdev`](gpu-mdev) (VM only): Creates and passes a virtual GPU through into the instance.
 - [`mig`](gpu-mig) (container only): Creates and passes a MIG (Multi-Instance GPU) through into the instance.
 - [`sriov`](gpu-sriov) (VM only): Passes a virtual function of an SR-IOV-enabled GPU into the instance.
+- [`native-context`](gpu-native-context) (VM only): Gives the VM GPU acceleration through `virtio-gpu` DRM native context, without passing the GPU through.
 
 The available device options depend on the GPU type and are listed in the tables in the following sections.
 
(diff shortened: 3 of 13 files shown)


stgraber:
Looks like the doc needs some escaping to pass the tests and the code is using some define+check syntax which we don't allow.

cmspam:
@stgraber Thanks for looking at this. I took care of what you mentioned. Spellcheck passes, and define+check eliminated. I also added details about virglrenderer version requirements in the documentation. Happy to adjust anything else if needed.

jochumdev:
Tested this works great!

Note that I had to install these packages on ArchLinux:

virglrenderer
qemu-hw-display-virtio-gpu-pci
qemu-ui-egl-headless

Having those not installed makes `qemu-system-x86_64 -device virtio-gpu-gl-pci,help` crash and the system will not start cause of that.

I installed `qemu-full` after all :)

---

FPS is low in `virt-viewer` for my favorite game `Beyond all Reason`, have to check direct rendering or something.

stgraber:
I've been working on and off on this one for a little bit. Got the package to behave, will limit to Debian 13 and Ubuntu 26.04 just to avoid all mesa/GL stacks.

I'm also going to tweak the AppArmor rules because here I needed to add:
```
    deny capability sys_admin,
    deny /var/lib/incus/.cache/ wklx,
    /usr/share/drirc.d/ r,
    /usr/share/drirc.d/** r,
    /usr/share/glvnd/egl_vendor.d/ r,
    /usr/share/glvnd/egl_vendor.d/** r,
    /usr/share/libdrm/** r,
    /etc/glvnd/egl_vendor.d/ r,
    /etc/glvnd/egl_vendor.d/** r,
```

Once I've done that and the static analysis looks happy, I'll merge it.

cmspam:
> I've been working on and off on this one for a little bit. Got the package to behave, will limit to Debian 13 and Ubuntu 26.04 just to avoid all mesa/GL stacks.
> 
> I'm also going to tweak the AppArmor rules because here I needed to add:
> 
> ```
>     deny capability sys_admin,
>     deny /var/lib/incus/.cache/ wklx,
>     /usr/share/drirc.d/ r,
>     /usr/share/drirc.d/** r,
>     /usr/share/glvnd/egl_vendor.d/ r,
>     /usr/share/glvnd/egl_vendor.d/** r,
>     /usr/share/libdrm/** r,
>     /etc/glvnd/egl_vendor.d/ r,
>     /etc/glvnd/egl_vendor.d/** r,
> ```
> 
> Once I've done that and the static analysis looks happy, I'll merge it.

Kernel 6.13 or higher is needed for drm-native-context to work, anyway, so unless someone was running an old debian/ubuntu with a new kernel, it wouldn't work anyway.

Does this mean we may see support for this feature in IncusOS with the gpu-support application?

stgraber:
Currently the gpu-support application only really brings in the firmware files, not any library, so that would probably need some tweaking.

cmspam:
> Currently the gpu-support application only really brings in the firmware files, not any library, so that would probably need some tweaking.

Indeed, without patching qemu so that it doesn't need opengl/egl-headless, to support amd/intel, I believe an enablement application would require:

  | Package | Version | Installed | .deb |
  |---|---|---:|---:|
  | `libllvm19` | 1:19.1.7 | 124 MB | 26 MB |
  | `mesa-libgallium` | 25.0.7 | 41 MB | 9.6 MB |
  | `libvirglrenderer1` | 1.3.0 | 2.1 MB | ~0.4 MB |
  | `libglx-mesa0` | 25.0.7 | 0.5 MB | 0.14 MB |
  | `libegl-mesa0` | 25.0.7 | 0.41 MB | 0.13 MB |
  | `libgl1-mesa-dri` | 25.0.7 | 0.24 MB | 46 KB |
  | `libgbm1` | 25.0.7 | 0.13 MB | 45 KB |
  | `libgl1` / `libglx0` / `libegl1` (libglvnd) | 1.7.0 | ~0.9 MB | ~0.16 MB |
  | `libdrm2` / `-amdgpu1` / `-intel1` / `-common` | 2.4.124 | ~0.44 MB | ~0.13 MB |
  | **Total** | | **~170 MB** | **~36 MB** |

libvirglrenderer 1.3.0 would be a compiled build since the version in trixie is too old, and would also have to match what you compile qemu against.

I'm not sure how you feel about the size for including into IncusOS, but it would be very cool to have access to the feature. Optionally you could add vulkan drivers as well and then support for venus would be available too.

stgraber:
Yeah, we'll see. First we'll want that in the Zabbly packages (which I'm working on), then get that into Incus 7.3, have some folks use it and report back, then we can look at IncusOS.

cmspam:
Sounds good. Feel free to let me know if there's anything I can do to help.

stgraber:
Alright, got it all tested and working. Next daily Zabbly package will include virglrenderer and the QEMU config tweaks. Incus change here mostly worked but I had to expand the AppArmor rules and add logic to handle the ownership of /dev/dri/ on systems with an `incus` user.

stgraber:
Anyway, got it all running on a test AMD laptop, getting pretty smooth 70 FPS in Unigine Heaven from inside the VM, so that's behaving as expected.

cmspam:
Very cool.  Happy to know it works with the zabbly builds!

I had compeltely overlooked AppArmor and /dev/dri ownership, because it didn't apply to my testing environment. Sorry about that.

nekoteai:
I wondered can this feature work with snapshot?

PR #3579, opened 2026-07-03
Fix concurrent image unpacking on ceph

DarkressX:
The Bug:
I was seeing the error:
`Failed creating instance from image: Failed to run: rbd --id admin --cluster ceph --image-feature exclusive-lock --image-feature object-map --image-feature fast-diff --image-feature layering clone ceph-default/image_f7cb36e58c37989e8073fb579874cf251d3e94e6077e182579d1e195e2c8874b_ext4.block@readonly     
  ceph-default/virtual-machine_0199f257-a97e-7c91-bee4-8537f74ce4e7_test.block: exit status 2 (2026-07-02T12:21:06.068+0000 7efc667fc6c0 -1 librbd::image::RefreshRequest: failed to locate snapshot: readonly\n2026-07-02T12:21:06.068+0000 7efc667fc6c0 -1 librbd::image::OpenRequest: failed to find snapshot   
  readonly\n2026-07-02T12:21:06.072+0000 7efc55ffb6c0 -1 librbd::image::CloneRequest: 0x5561bd98bda0 handle_open_parent: failed to open parent image: (2) No such file or directory\nrbd: clone error: (2) No such file or directory)`
during concurrent VM creation on multiple cluster members from the same image.
This caused a failure during the creation.

The problem was that incus  (via the ceph rbd driver) considers an image to be a viable clone source, once its in the DB and the image has been created on ceph.
This unfortunately is not enough as the image needs to have the `readonly` snapshot and this snapshot also needs to be protected.

Since the protection is the last stage that happens after a fresh unpack, I implemented a check, that check that polls until the image has the protected `readonly` snapshot. Only then is it considered to be a viable clone source.
This of course only happens if the image is detected and the snapshot is not ready yet.

The ugly downcasting is there, because i haven't worked with the other storage drivers yet.
As soon as it's implemented there as well the casting can be safely removed.

### DIFF
--- a/internal/server/storage/backend.go
+++ b/internal/server/storage/backend.go
@@ -3884,6 +3884,48 @@ func (b *backend) UnmountInstanceSnapshot(inst instance.Instance, op *operations
 	return err
 }
 
+// waitImageCloneSourceReady blocks until the optimized image volume's clone source is ready to be cloned from.
+func (b *backend) waitImageCloneSourceReady(imgVol drivers.Volume) error {
+	ctx, cancel := context.WithTimeout(context.TODO(), 10*time.Minute)
+	defer cancel()
+
+	for {
+		// Only drivers that build a separate clone source (a protected readonly snapshot) after
+		// unpacking are affected by this race and implement the readiness check. For all other drivers
+		// the clone source is the image volume itself, so there is nothing to wait for.
+		ready, err := b.driver.IsImageCloneSourceReady(imgVol)
+		if err != nil {
+			return err
+		}
+
+		if ready {
+			return nil
+		}
+
+		// Detect creator abort/revert so we fail fast instead of blocking until the timeout: if
+		// the base image volume or its cluster DB row has disappeared, the creating member gave up.
+		exists, err := b.driver.HasVolume(imgVol)
+		if err != nil {
+			return err
+		}
+
+		dbVol, err := VolumeDBGet(b, api.ProjectDefaultName, imgVol.Name(), drivers.VolumeTypeImage)
+		if err != nil && !response.IsNotFoundError(err) {
+			return err
+		}
+
+		if !exists || dbVol == nil {
+			return fmt.Errorf("Image %q clone source is unavailable because its creation was aborted on another cluster member; please retry", imgVol.Name())
+		}
+
+		select {
+		case <-time.After(1 * time.Second):
+		case <-ctx.Done():
+			return fmt.Errorf("Timed out waiting for image %q clone source to become ready: %w", imgVol.Name(), ctx.Err())
+		}
+	}
+}
+
 // EnsureImage creates an optimized volume of the image if supported by the storage pool driver and the volume
 // doesn't already exist. If the volume already exists then it is checked to ensure it matches the pools current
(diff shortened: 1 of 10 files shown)


internal/server/storage/backend.go
```diff
@@ -3884,6 +3884,64 @@ func (b *backend) UnmountInstanceSnapshot(inst instance.Instance, op *operations
...
 }
 
+// imageCloneSourceReadyTimeout is the maximum time EnsureImage will wait for another cluster member
+// to finish preparing a shared image volume's clone source before giving up. It must comfortably
+// exceed the worst-case image unpack time; creator failure is detected actively (see
+// waitImageCloneSourceReady) so this is only a backstop.
+const imageCloneSourceReadyTimeout = 10 * time.Minute
+
+// imageCloneSourceReadyPoll is the interval between clone source readiness checks.
+const imageCloneSourceReadyPoll = 1 * time.Second
```
stgraber: I usually don't do const when they're only used once.

I've seen LLMs (Claude mostly) that really push for this pattern, but having to scroll back up just to see the value just makes things more annoying. That is unless it's used in a bunch of places or is exported, but neither is true here.

DarkressX: Generally I would agree, however magic numbers should also be discourages from being used in the code.
In this case it should be fine, as its obviously just a time duration and therefore not really a magic number.
I removed them

stgraber: Yeah, agreed. So far magic numbers for us have mostly been ioctls and the like which we've concentrated as exported const in the `linux` package.

stgraber:
Hmm, looks like another comment I had prepared got lost somewhere.

On top of the `const` thing which you've fixed, I was also pointing out that you could add the new function to the driver interface and put a placeholder that always returns true in the `common` driver.

With that you wouldn't need to do the casting dance as all drivers would have the function, then those that need extra care like the Ceph driver can implement the function with added logic.

In general all the local storage drivers should be fine as they operate under a server-wide lock avoiding the issue. It's really the network ones that may run into the issue of the image appearing ready before it actually is (ceph, linstor, lvmcluster, truenas).

DarkressX:
> Hmm, looks like another comment I had prepared got lost somewhere.
> 
> On top of the `const` thing which you've fixed, I was also pointing out that you could add the new function to the driver interface and put a placeholder that always returns true in the `common` driver.
> 
> With that you wouldn't need to do the casting dance as all drivers would have the function, then those that need extra care like the Ceph driver can implement the function with added logic.
> 
> In general all the local storage drivers should be fine as they operate under a server-wide lock avoiding the issue. It's really the network ones that may run into the issue of the image appearing ready before it actually is (ceph, linstor, lvmcluster, truenas).

Yeah I was thinking that too, but wasn't sure whats cleaner.
I'll change that up too and just use placeholder functions in the other drivers for now.

DarkressX:
Alright, so i've added it to the interface, and added the `IsImageCloneSourceReady` function to the interface and guarded the call behind a check, so that it only runs for remote storage drivers (lvm included).
So far there only is an implementation for the ceph rbd driver, all other remote drivers have a basic implementation with TODO comments.

PR #3582, opened 2026-07-03
Fallback to sending agent state data through vserial ringbuffer

### LINKED ISSUE #3196: Add fallback communication logic to incus-agent
Currently the incus-agent expects to receive connections over vsock and to be able to similarly connect to the host over vsock (for /dev/incus API). We have limited support to connect over HTTPS as a fallback mechanism but that only really works when run on an Incus-managed network where we can predict the IP address of the instance.

At FuturFusion, most of our customers seem to mostly care about the agent running for reporting of IP configuration as well as OS information. The other features like file transfers, command execution and to an extent, the extra resource information are often disabled by policy anyway.

As we're working with pretty old platforms (see #3195), vsock is often unavailable and even HTTPS could be a challenge due to those systems usually not having general network connectivity (they're EOL and very much unpatched).

The one thing that works quite reliably is access to the virtio serial device, so the idea is to extend incus-agent to have a fallback mode where when vsock isn't detected, it will start sending its state struct (OS, IP, ...) as JSON serialized data through the serial port alongside its usual heartbeat.

Incus can then pick that up and hold it in memory (with a reasonable size limit to avoid potential abuses).
Then when the state information is requested from Incus, that data can be used as a substitute for the agent API call.

masnax:
Closes #3196

I used 16KiB for the ring buffer size after testing a VM with 9 NICs attached creating a 4KiB state object. Most of them didn't have addresses set, so once that's accounted for (as well as multiple addresses per NIC), I think 16KiB should be more than sufficient for most real-world use cases.

### DIFF
--- a/internal/server/instance/drivers/qmp/monitor.go
+++ b/internal/server/instance/drivers/qmp/monitor.go
@@ -22,7 +22,7 @@ var (
 )
 
 // RingbufSize is the size of the agent serial ringbuffer in bytes.
-var RingbufSize = 16
+var RingbufSize = 1024 * 16
 
 // EventAgentStarted is the event sent once the agent has started.
 var EventAgentStarted = "AGENT-STARTED"
--- a/internal/server/instance/drivers/qmp/monitor.go
+++ b/internal/server/instance/drivers/qmp/monitor.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/lxc/incus/v7/internal/linux"
+	"github.com/lxc/incus/v7/shared/api"
 	"github.com/lxc/incus/v7/shared/logger"
 	"github.com/lxc/incus/v7/shared/util"
 )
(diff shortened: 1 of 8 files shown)


internal/server/instance/drivers/qmp/monitor.go
```diff
@@ -22,7 +22,7 @@ var (
 )
 
 // RingbufSize is the size of the agent serial ringbuffer in bytes.
-var RingbufSize = 16
+var RingbufSize = 1024 * 16
```
stgraber, line 26: Will need to validate that cross-version live-migration works as expected (VMs started with 16 bytes buffer being migrated to Incus with a default of 16k)

stgraber:
Looks good to me. I'll just do a quick test here with a cluster to make sure we can live migrate between both ring buffer sizes. If not, we'll need to add some state checking logic.

stgraber (approved):

internal/server/instance/drivers/qmp/monitor.go
stgraber: Confirmed to not fail migration, we're good.

stgraber (approved):

PR #3587, opened 2026-07-05
Use persistent Logical Switch Ports with OVN

stgraber:
Up until now Incus would create and delete OVN ports at every NIC device start/stop event.

This would keep OVN looking very clean with only active instances getting database records, but it also comes with some annoying downsides.

The main one being that when using dynamic allocation, restarting a bunch of instance makes it unlikely that they'll all get their previous IP back. Incus has a volatile key to try to restore those addresses on a best-effort basis but when dealing with a large amount of instances coming up at once, a bunch are bound to change addresses.

Instead, we're now keeping an OVN LSP for the instance's lifetime but marking it as disabled when the instance is stopped.

### DIFF
--- a/internal/server/network/driver_ovn.go
+++ b/internal/server/network/driver_ovn.go
@@ -5511,7 +5511,7 @@ func (n *ovn) InstanceDevicePortIPs(instanceUUID string, deviceName string) ([]n
 	return devIPs, nil
 }
 
-// InstanceDevicePortStop deletes an instance device port from the internal logical switch.
+// InstanceDevicePortStop disables an instance device port and removes its routes and NAT rules.
 func (n *ovn) InstanceDevicePortStop(ovsExternalOVNPort networkOVN.OVNSwitchPort, opts *OVNInstanceNICStopOpts) error {
 	// Decide whether to use OVS provided OVN port name or internally derived OVN port name.
 	instancePortName := ovsExternalOVNPort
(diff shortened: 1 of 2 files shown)


internal/server/network/ovn/ovn_nb_actions.go
```diff
@@ -3056,10 +3056,23 @@ func (o *NB) UpdatePortGroupACLRules(ctx context.Context, portGroupName OVNPortG
 	return nil
 }
 
-// AddLogicalSwitchQoSRules applies a set of rules to the specified logical switch port.
+// AddLogicalSwitchQoSRules replaces the set of QoS rules assigned to the specified logical switch port.
```
bensmrs: Then maybe the function should be renamed

internal/server/network/driver_ovn.go
```diff
@@ -4742,6 +4746,21 @@ func (n *ovn) InstanceDevicePortAdd(instanceUUID string, deviceName string, devC
 		}
 	}
 
+	// Create the logical switch port in disabled state.
```
bensmrs, line 4749: What’s the motivation behind that, if the port is created anyway if missing?

bensmrs (approved):

internal/server/network/ovn/ovn_nb_actions.go
stgraber: Hmm, indeed, at this point `SetLogicalSwitchQoSRules` would make more sense.

internal/server/network/driver_ovn.go
stgraber: Accounting mostly. I'd rather we run out of IP addresses upfront, not at instance startup.

bensmrs: Yeah, good enough :)

internal/server/network/ovn/ovn_nb_actions.go
stgraber: Fixed!

PR #3588, opened 2026-07-05
Various bugfixes

stgraber:


cmd/incusd/instance_console.go
```diff
@@ -259,6 +259,9 @@ func (s *consoleWs) doConsole() error {
 			conn := s.conns[-1]
 			s.connsLock.Unlock()
 
+			// Bound control message size to avoid unbounded reads.
+			conn.SetReadLimit(ws.ControlMessageMaxSize)
```
bensmrs, line 263: What is this solving exactly?

bensmrs (approved):

cmd/incusd/instance_console.go
stgraber: Prevents a bad client from sending us a massive control message which we'd then fully load into memory.

So it's DoS protection though only for already fully authenticated clients.

PR #3596, opened 2026-07-06
incusd/network/ovn: Fix NAT for network forward default targets

### LINKED ISSUE #3578: `ipvX.address.external` non functional with Forward

### Incus system details

```yaml
config:
  cluster.https_address: 10.196.163.13:8443
  core.https_address: 10.196.163.13:8443
  network.ovn.ca_cert: SENSITIVE
  network.ovn.client_cert: SENSITIVE
  network.ovn.client_key: SENSITIVE
  network.ovn.northbound_connection: ssl:[10.196.163.11]:6641,ssl:[10.196.163.12]:6641,ssl:[10.196.163.13]:6641
api_extensions_count: 529
api_status: stable
api_version: "1.0"
auth: trusted
public: false
auth_methods:
  - tls
auth_user_name: root
auth_user_method: unix
environment:
  addresses:
    - 10.196.163.13:8443
  architectures:
    - x86_64
    - i686
  certificate: ""
  driver: lxc | qemu
  driver_version: 7.0.0 | 11.0.2
  firewall: nftables
  kernel: Linux
  kernel_architecture: x86_64
  kernel_features: {}
  kernel_version: 6.12.90+deb13.1-amd64
  lxc_features: {}
  os_name: Debian GNU/Linux
  os_version: "13"
  project: infra-block-dev0b
  server: incus
  server_clustered: true
  server_event_mode: full-mesh
  server_name: 2282152-eos-dev0b-n003
  server_pid: 2811403
  server_version: "7.2"
  storage: ceph | dir | zfs
  storage_version: 18.2.7 | 1 | 2.3.2-2
  storage_supported_drivers:
    - name: ceph
      version: 18.2.7
      remote: true
    - name: cephfs
      version: 18.2.7
      remote: true
    - name: cephobject
      version: 18.2.7
      remote: true
    - name: dir
      version: "1"
      remote: false
    - name: truenas
      version: 0.7.7
      remote: true
    - name: zfs
      version: 2.3.2-2
      remote: false
    - name: btrfs
      version: "6.14"
      remote: false
    - name: lvm
      version: 2.03.31(2) (2025-02-27) / 1.02.205 (2025-02-27) / 4.48.0
      remote: false
---
ovn-nbctl 25.09.0
Open vSwitch Library 3.6.0
DB Schema 7.12.0
```

### Instance details

_No response_

### Instance log

_No response_

### Current behavior

This is essentially a continuation of #2921 and the fix in #3397.

What's different now is that the `dnat_and_snat` rule is successfully created, but since a forward must be in-place for this to happen when an instance has `ipvX.address.external` set, two load-balancers with empty port and protocol definitions are also created in OVN.

```
~# ovn-nbctl lr-lb-list incus-net248-lr
UUID                                    LB                  PROTO      VIP               IPs
b7e7c7c4-0ada-42de-ac05-9b82a49c9cfc    incus-net248-lb-               <EXTERNAL_IP>    192.168.126.2
f3cdecca-b3e4-4ff9-b829-5224d034e8bf    incus-net248-lb-               <EXTERNAL_IP>     192.168.126.2
```

These are essentially what break outbound traffic. When we remove these manually via `ovn_nbctl lr-lb-del`, outbound traffic is correctly restored for an instance.

### Expected behavior

When we create the necessary forward like so:

```
description: ""
config:
  target_address: 192.168.126.2
ports: []
listen_address: <EXTERNAL_IP>
location: ""
```

and an instance has it's external ip set

```
devices:
  eth0:
    ipv4.address: 192.168.126.2
    ipv4.address.external: <EXTERNAL_IP>
    type: nic
```

Ideally, Incus should NOT create any LB's in OVN, since this configuration points to a direct `dnat_to_snat` forward. Creating a forward as a necessity for this case could be a bit confusing for a user. Which is why we could maybe skip the forward dependency entirely?

### Steps to reproduce

1. Create a OVN Network, e.g. `test-net-1`
2. Create a new VM instance in `test-net-1`, e.g. `test-vm-1`
3. Then create a network forward in `test-net-1` with your external ip as listener and `test-vm-1` ip address as default target, e.g. `1.1.1.1`

```diff
+description: ""
+config:
+  target_address: <TEST_VM_1_ADDRESS>
+ports: []
+listen_address: 1.1.1.1
+location: ""
```

4. Then add this external ip as a `.external` in `test-vm-1` config

```diff
devices:
  eth0:
+    ipv4.address.external: 1.1.1.1
```

5. Try to curl anything public from the instance, e.g. `curl v4.ident.me` -> it will time-out

0xk1f0:
Addresses #3578

A network forward with only a `target_address` (no port maps) generated a portless VIP that got duplicated into both the `tcp` and `udp` `Load_Balancer` objects per forward.
To my knowledge, when that same address was also used by `ipv4/ipv6.address.external`'s `snat` rule, OVN's UNSNAT stage will skip reversing the egress SNAT for TCP/UDP traffic on that address, because the address was now both a NAT `external_ip` and a `Load_Balancer` VIP. Outbound TCP/UDP from the instance would silently break because of this.

This now changes the following:
- A forward's default `target_address` is now implemented as a plain router `dnat` NAT rule. It never touches the Load_Balancer table, so it can't collide with any `snat` rule on the same address, and it still correctly works with one (equivalent to `dnat_and_snat`) without either side needing to know about the other.
- Port maps are untouched, they still create and populate `Load_Balancer`'s, correctly split by protocol.
- `ipv4/ipv6.address.external` goes back to always creating a plain `snat` rule, which since a forward is required for this to be set, complements the `dnat` rule added by the forward.

So a forward with a default target no longer creates any `Load_Balancer` entry, and pairing it with `ipv4/ipv6.address.external` on the same address now gives correct bidirectional NAT.

### DIFF
--- a/internal/server/network/driver_ovn.go
+++ b/internal/server/network/driver_ovn.go
@@ -4776,33 +4776,6 @@ func (n *ovn) hasDHCPv4Reservation(dhcpReservations []iprange.Range, ipAddress n
 	return false
 }
 
-// forwardHasDefaultTarget reports whether a network forward exists for the given listen address
-// and has a default target address configured.
-func (n *ovn) forwardHasDefaultTarget(listenAddress string) (bool, error) {
-	var hasDefaultTarget bool
-
-	err := n.state.DB.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
-		dbForward, err := dbCluster.GetNetworkForward(ctx, tx.Tx(), n.ID(), listenAddress)
-		if err != nil {
-			return err
-		}
-
-		forward, err := dbForward.ToAPI(ctx, tx.Tx())
-		if err != nil {
-			return err
-		}
-
-		hasDefaultTarget = forward.Config["target_address"] != ""
-
-		return nil
-	})
-	if err != nil {
-		return false, err
-	}
-
-	return hasDefaultTarget, nil
-}
-
 // InstanceDevicePortStart sets up and enables an instance device port on the internal logical switch.
 // Accepts a list of ACLs being removed from the NIC device (if called as part of a NIC update).
 // Returns the logical switch port name and a list of IPs that were allocated to the port for DNS.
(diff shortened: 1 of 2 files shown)


stgraber:
@0xk1f0 needs to be rebased on the recent LSP change.

0xk1f0:
@stgraber Done, should be good to go!

PR #3600, opened 2026-07-06
Add support for BGP unnumbered peers

stgraber:


bensmrs (approved):
Way cleaner than I expected :)

PR #3605, opened 2026-07-06
 incusd/instance/qemu: Handle missing kvm64

### LINKED ISSUE #3603: VM feature checks fail on qemu 10.x (RHEL/AlmaLinux 10): CPU model 'kvm64' not found

### Incus system details

```yaml
On AlmaLinux 10 (qemu-kvm 10.1.0), VM support is disabled: Unable to run feature checks during QEMU initialization:

Failed to query CPU model: kvm64 not found. qemu ≥10 removed the kvm64 model; driver_qemu.go still calls QueryCPUModel("kvm64"). Please use a model present in qemu 10 (e.g. qemu64) for the x86_64 baseline probe

Required information (incus info):
   server_version: "7.2"   (also 6.23)
   os_name: AlmaLinux   os_version: "10.2"
   kernel_version: 6.12.0-211.29.1.el10_2.x86_64
   driver: lxc | qemu     driver_version: 6.0.6 | 10.1.0
   storage: btrfs
   incus config show <vm> / incus info --show-log <vm>: not applicable — the VM never gets created, because VMs are disabled daemon-wide (see below), so no instance exists.
```

### Instance details

_No response_

### Instance log

_No response_

### Current behavior

With qemu-kvm 10.1.0, the qemu driver is not operational:

Unable to run feature checks during QEMU initialization:

 Failed to query CPU model: GenericError: CPU model 'kvm64' not found
 Instance type not operational  driver=qemu  err="QEMU failed to run feature checks"
 qemu ≥10 removed the legacy kvm64 CPU model, but internal/server/instance/drivers/driver_qemu.go still calls monitor.QueryCPUModel("kvm64") for the x86_64 baseline probe (present in current main).

### Expected behavior

VMs are supported on qemu 10.x. The baseline probe should use a model that exists in qemu 10 (e.g. qemu64), or fall back when kvm64 is absent.

### Steps to reproduce

1. Install AlmaLinux 10 (qemu-kvm 10.1.0) and Incus (e.g. COPR neelc/incus).
2. incus admin init --minimal
3. incus launch images:ubuntu/24.04 v1 --vm → fails; incus info shows the qemu driver as not operational with the kvm64 error.

Suggested fix: one line — QueryCPUModel("kvm64") → QueryCPUModel("qemu64") (verified working on AlmaLinux 10 / qemu 10.1)

glowf1sh:
Fixes #3603 

qemu >= 10.0 removed the legacy `kvm64` CPU model. The host-CPU baseline probe
in the qemu driver still requests `kvm64`, so on qemu 10.x the feature check
fails ("CPU model 'kvm64' not found") and the qemu driver is marked not
operational — VMs are disabled entirely on AlmaLinux/RHEL 10 (qemu-kvm 10.1.0).

This switches the x86_64 baseline probe to `qemu64`, which is the generic
baseline still present in qemu 10.x. Verified on AlmaLinux 10.2 / qemu-kvm
10.1.0: the qemu driver becomes operational again.

Note: 

qemu64 is itself marked deprecated in qemu 10.1, so a longer-term fix may
prefer to fall back gracefully when the probed baseline model is absent, rather
than hard-coding another named model.

### DIFF
--- a/internal/server/instance/drivers/driver_qemu.go
+++ b/internal/server/instance/drivers/driver_qemu.go
@@ -10379,7 +10379,11 @@ func (d *qemu) checkFeatures(hostArch int, qemuPath string) (map[string]any, err
 	if hostArch == osarch.ARCH_64BIT_INTEL_X86 {
 		model, err := monitor.QueryCPUModel("kvm64")
 		if err != nil {
-			return nil, err
+			// Fallback to qemu64 if kvm64 is missing (RHEL).
+			model, err = monitor.QueryCPUModel("qemu64")
+			if err != nil {
+				return nil, err
+			}
 		}
 
 		cpuFlags := map[string]bool{}


stgraber:
The correct fix here is to lookup `qemu64` as the fallback if `kvm64` isn't found.

It's only RHEL that's been playing the aggressive deprecation game. Upstream QEMU hasn't deprecated either of those CPU types (https://www.qemu.org/docs/master/about/deprecated.html).

glowf1sh:
> The correct fix here is to lookup `qemu64` as the fallback if `kvm64` isn't found.
> 
> It's only RHEL that's been playing the aggressive deprecation game. Upstream QEMU hasn't deprecated either of those CPU types (https://www.qemu.org/docs/master/about/deprecated.html).

true :(

Thanks — I'll push an updated version that keeps `kvm64` and only falls back to `qemu64` when it isn't available.

Context on why I'm on 7.x in the first place: AlmaLinux 10's QEMU build ships without `virtio-9p-pci` (and `virtio-sound-pci`), so on 6.23 a VM fails to start because its 9p config drive device is rejected ("`virtio-9p-pci` is not a valid device model name"). 7.x's virtio-fs config drive / device feature-gating is exactly what handles that, so 7.x is the version I need here — 6.23 can't do VMs on this QEMU.

The problem: when I build **7.2** from the release tarball on AlmaLinux 10 (cowsql 1.15.9, go-cowsql v1.22.0, RPM spec derived from the neelc/incus one that builds a working 6.23), the database layer misbehaves — 
every *get-by-name* lookup returns "not found" while `list` works:

      $ incus storage list        # shows it
      | bulk | btrfs | CREATED |
      $ incus storage show bulk
      Error: not found

Same for `profile show default`; instance creation and `incus admin init --minimal` ("profile already exists") fail too. The **identical build of 6.23** is completely fine (containers, init, and instance *creation* all work — it's only the VM *boot* that hits the missing QEMU device above).

Is there a build-side requirement that changed between 6.23 and 7.x for EL/non-Ubuntu builds (cowsql/libsqlite3 build tags, dqlite/raft version)? Happy to open a separate issue with a full repro.

stgraber:
@glowf1sh you're almost certainly using a broken Go build command.

Incus 7.2 switched from gorilla/mux to the native Go muxer.
There are some build parameters which completely break that new muxer, most notably setting GO111MODULE=off or having GODEBUG contain httpmuxgo121=1

stgraber:
@glowf1sh note that I updated your branch for you with the approach I suggested, just waiting for one more test to pass then it will get merged.

stgraber (approved):

PR #3609, opened 2026-07-07
Support persistent bitmaps for raw-format block devices

### LINKED ISSUE #3472: Support persistent bitmaps for raw-format block devices

Currently, we support bitmaps both persistent and transient.
A persistent ones are only supported when qcow2 is used as an underlying block device format.

From the discussion [1] I learned that QEMU supports a scheme, when qcow2 file is used as a container to store metadata (like bitmaps, in our case), while block device data is stored in a "data file". The idea here is to use that way, to support persistent bitmaps for non-qcow2 case.

I have done a draft of a very ugly implementation [see here](https://gist.github.com/mihalicyn/6b1e433cd37cc813315098bc94f4dfc5) and [here](https://github.com/mihalicyn/incus/commit/ba181a92f95e1235a8b662b01d0cbca08cfaae93) to check if this approach works and there is no hidden obstacles along the way. And yes, I can confirm that it works :-)

Link: https://lists.gnu.org/archive/html/qemu-devel/2021-03/msg07428.html [1]

_No response_

presztak:
Closes: #3472

### DIFF
--- a/internal/server/device/config/device_runconfig.go
+++ b/internal/server/device/config/device_runconfig.go
@@ -21,17 +21,18 @@ type RunConfigItem struct {
 
 // MountEntryItem represents a single mount entry item.
 type MountEntryItem struct {
-	DevName     string      // The internal name for the device.
-	DevPath     string      // Describes the block special device or remote filesystem to be mounted.
-	BackingPath []string    // Describes the block special device to be mounted as backing drive for qcow2.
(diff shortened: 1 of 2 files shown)


stgraber:
I'll need to test that on one of my test clusters to see how the live migration handling works for existing VMs that were started without that logic.

stgraber:
Tested and QEMU doesn't seem to mind at all.

PR #3623, opened 2026-07-11
Various bugfixes

stgraber:


internal/server/instance/drivers/driver_lxc.go
```diff
@@ -7702,9 +7702,23 @@ func (d *lxc) FileSFTPConn() (net.Conn, error) {
...
+	// Record the socket file identity so cleanup doesn't remove a newer socket.
+	forkfileInfo, err := os.Stat(forkfilePath)
+	if err != nil {
+		_ = forkfileListener.Close()
+		return nil, err
+	}
+
+	removeForkfileSocket := func() {
+		info, err := os.Stat(forkfilePath)
+		if err == nil && os.SameFile(info, forkfileInfo) {
```
bensmrs, line 7714: It can technically still race after that condition passes. Can’t we suffix the socket name with something unique?

stgraber: Not really because the whole point of forkfile is that we have subsequent requests re-use the existing socket and connect to it.

forkfile has a built-in timeout where it exits 10s since the last user interaction.
The issue we're sometimes seeing is when you get a new incoming connection spawning a new daemon just as the previous one is getting cleaned up.

That can trigger the cleanup logic of the existing daemon to remove the socket of the new one, causing the error we're seeing.

The fix makes sure we're deleting the correct thing to avoid that particular race.
There's still a potential race if someone connects to the old daemon just as it's exiting, but that's a much smaller race and not one we've ever seen any reports of.

bensmrs: Yeah, that was the race I was talking about

bensmrs (approved):

PR #3639, opened 2026-07-14
Add support for multiple authorizers

### LINKED ISSUE #3488: Support for using OpenFGA for authorization with TLS client certs

TLS certs are used when building automation around Incus and it would provide more security if we could limit a TLS certs to have more restrictive granular permissions.

Its currently only possible to limit a TLS cert to a specific project.
It would be good to allow TLS certs to be configured so they can be validated against OpenFGA.
This would use one of the fields in the cert as the username passed to OpenFGA.

presztak:
This PR adds the ability to define multiple authorizers for different authentication methods. A specific authorizer can be set per method using these keys:
```
authorization.client.default: deny
authorization.client.unix: scriptlet
authorization.client.tls: allow
authorization.client.tls-restricted: openfga
authorization.client.oidc: openfga
```

In addition, two new authorizers were added: `allow` and `deny`, which effectively allow and deny all requests respectively.
For the case of OpenFGA mediating TLS clients, a new configuration key `authorization.openfga.tls.identifier` was added to decide whether to mediate access based on the TLS fingerprint or the certificate's name.

Closes: #3488

### DIFF
--- a/cmd/incusd/patches.go
+++ b/cmd/incusd/patches.go
@@ -15,6 +15,7 @@ import (
 	linstorClient "github.com/LINBIT/golinstor/client"
 
 	internalInstance "github.com/lxc/incus/v7/internal/instance"
+	"github.com/lxc/incus/v7/internal/server/auth"
 	"github.com/lxc/incus/v7/internal/server/backup"
 	"github.com/lxc/incus/v7/internal/server/certificate"
 	"github.com/lxc/incus/v7/internal/server/cluster"
@@ -102,6 +103,7 @@ var patches = []patch{
 	{name: "linstor_tune_rs_discard_granularity", stage: patchPostDaemonStorage, run: patchLinstorDiscardGranularity},
 	{name: "storage_lvmcluster_qcow2_overhead", stage: patchPostDaemonStorage, run: patchGenericStorage},
 	{name: "authorization_openfga_config_keys", stage: patchPreDaemonStorage, run: patchAuthorizationOpenFGAConfigKeys},
+	{name: "authorization_client_routes_from_driver", stage: patchPreDaemonStorage, run: patchAuthorizationClientRoutesFromDriver},
 }
 
 type patchRun func(name string, d *Daemon) error
@@ -1873,6 +1875,49 @@ func patchAuthorizationOpenFGAConfigKeys(_ string, d *Daemon) error {
 
 		if len(changes) == 0 {
 			return nil
+		}
+
+		err = tx.UpdateClusterConfig(changes)
+		if err != nil {
+			return fmt.Errorf("Failed updating cluster config: %w", err)
+		}
+
+		return nil
+	})
+}
+
+// patchAuthorizationClientRoutesFromDriver seeds authorization.client.* config
+// from the current authorizer class, so an upgrade keeps enforcing the same driver.
+func patchAuthorizationClientRoutesFromDriver(_ string, d *Daemon) error {
+	return d.db.Cluster.Transaction(context.TODO(), func(ctx context.Context, tx *db.ClusterTx) error {
+		config, err := tx.Config(ctx)
+		if err != nil {
+			return fmt.Errorf("Failed getting cluster config: %w", err)
+		}
+
+		// If any explicit route is already set, the routing table has been
+		// configured directly; leave it untouched.
+		for _, class := range []string{"default", "unix", "tls", "tls-restricted", "oidc"} {
+			if config["authorization.client."+class] != "" {
+				return nil
+			}
+		}
+
+		// Determine the implicitly selected driver, matching the legacy startup
+		// precedence: scriptlet over OpenFGA over the default (TLS) driver.
+		var changes map[string]string
+		switch {
+		case config["authorization.scriptlet"] != "":
+			changes = map[string]string{
+				"authorization.client.tls":            auth.DriverScriptlet,
+				"authorization.client.tls-restricted": auth.DriverScriptlet,
+				"authorization.client.oidc":           auth.DriverScriptlet,
+			}
+
+		case config["authorization.openfga.api.url"] != "" && config["authorization.openfga.api.token"] != "" && config["authorization.openfga.store.id"] != "":
+			changes = map[string]string{
+				"authorization.client.oidc": auth.DriverOpenFGA,
+			}
 		}
 
 		err = tx.UpdateClusterConfig(changes)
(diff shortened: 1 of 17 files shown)


test/suites/oidc.sh
```diff
@@ -8,6 +8,7 @@ test_oidc() {
     spawn_oidc
     incus config set "oidc.issuer=http://127.0.0.1:$(cat "${TEST_DIR}/oidc.port")/"
     incus config set "oidc.client.id=device"
+    incus config set "authorization.client.oidc=tls"
```
stgraber: This one looks wrong to me.

stgraber:
I still need to fully review the commit introducing most of the actual logic, but just looking at this from a higher level, I'm wondering what to do with the `tls` authorizer.

`tls` clients (unrestricted) are currently handled the same as with the `allow` authorizer.
Only the `tls-restricted` clients are effectively using the `tls` authorizer today.

I think that we should make the `tls` authorizer only be valid for `tls-restricted` as it's the only scenario where it actually does something.

We should also clearly document that setting `tls-restricted` to something other than `tls` will effectively ignore the `projects` field on the certificate entry. That's a behavior that we need to make using the `openfga` or `scriptlet` authorizers be flexible enough for restricted certificates, but we need a clear warning in the documentation as it may not be obvious and users may incorrectly assume that their projects list will be respected.

I'd have said we should just prevent that particular combination, having Incus reject any authorization request for a restricted certificate with a list of projects set, but I don't think that's necessarily the right thing to do as the projects list is exposed to the scriptlet authorizer, so the authorizer may actually make use of that list, it's just not guaranteed that it will (openfga won't).

stgraber:
I also think we need to make a specific carve out for the root user over the UNIX socket so it's not possible for a user to break it. Incus internally relies on the root user being able to interact freely with the API.

stgraber:
@presztak so I think the actions are:
 - Restrict the `tls` authorized to only `tls-restricted`
 - Add a clear WARNING in the documentation that setting `tls-restricted` to a value other than `tls` MAY lead to the per-certificate project list being ignored (dependent on authorizer)
 - Allow the root user to always interact with the UNIX socket API regardless of configured authorizer
 - Add another WARNING in the documentation to be very careful with the `default` authorizer, only ever setting it last after confirming that the rest works correctly as setting it to say `openfga` when OpenFGA is misconfigured would prevent any interaction with Incus other than the local root user UNIX access from above

presztak:
> * Allow the root user to always interact with the UNIX socket API regardless of configured authorizer

So `authorization.client.unix` key should be removed?

test/suites/oidc.sh
presztak: Could you elaborate? Are you referring to: `Restrict the tls authorizer to only tls-restricted`?

stgraber:
> So `authorization.client.unix` key should be removed?

No, it's just the root user that needs to get a special treatment.
It still makes sense to be able to use an authorizer for non-root users.

test/suites/oidc.sh
presztak: Set to `allow`.

presztak:
@stgraber , all comments addressed.

cmd/incusd/patches.go
```diff
@@ -1884,4 +1886,61 @@ func patchAuthorizationOpenFGAConfigKeys(_ string, d *Daemon) error {
...
+				"authorization.client.tls-restricted": auth.DriverTLS,
+			}
+
+		default:
+			changes = map[string]string{
+				"authorization.client.default":        auth.DriverAllow,
+				"authorization.client.unix":           auth.DriverAllow,
+				"authorization.client.tls":            auth.DriverAllow,
+				"authorization.client.tls-restricted": auth.DriverTLS,
+			}
```
stgraber: I don't think we should be adding in `authorization.XYZ` keys for systems that don't currently have either openfga or scriptlet configured. We should let them just use the default logic (that should match what's shown here) so it matches what you get on a fresh install (which also wouldn't have any of those keys set).

test/suites/oidc.sh
```diff
@@ -8,6 +8,7 @@ test_oidc() {
     spawn_oidc
     incus config set "oidc.issuer=http://127.0.0.1:$(cat "${TEST_DIR}/oidc.port")/"
     incus config set "oidc.client.id=device"
+    incus config set "authorization.client.oidc=allow"
```
stgraber: Why is that needed?

If our default when no `authorization.client` is set is to `allow` except for `tls-restricted` which uses `tls`, then this shouldn't be needed.

stgraber:
So I think we need to make sure that a system that's not using openfga/scriptlet which upgrades to a version of Incus with this logic gets the same configuration and behavior as a clean system (that is, no `authorization` keys set and internal behavior of unix=allow, oidc=allow, tls=allow, tls-restricted=tls).

Anyone with configured `openfga` or `scriptlet` will get the new `authorization` configuration keys and the relevant `authorization.client` config. I think I'd want to keep those configs as light as possible, so for `openfga`, only set `authorization.client.oidc=openfga` and for `scriptlet`, only set `authorization.client.oidc=scriptlet`, `authorization.client.tls=scriptlet` and `authorization.client.tls-restricted=scriptlet`.

We shouldn't need to set any of the other keys on upgrade.

test/suites/oidc.sh
presztak: No, this is no longer needed, it's a leftover.

cmd/incusd/patches.go
presztak: Removed.

doc/authorization.md
```diff
@@ -89,3 +93,62 @@ Additionally, two optional functions can be defined so that users can be listed
...
+The `root` user is always allowed to interact with the API over the Unix
+socket, regardless of the configured authorization method. This is what allows
+a server locked out by a bad routing configuration to be recovered.
+```
+
+For example, to have OpenFGA authorize OIDC clients while restricted TLS clients
+keep using TLS authorization:
+
+    incus config set authorization.client.oidc=openfga
+    incus config set authorization.client.tls-restricted=tls
```
stgraber: I don't think that's a good example as `tls-restricted` defaults to `tls`.

Maybe instead show:
 - oidc=openfga
 - tls-restricted=scriptlet

stgraber:
@presztak small doc tweak (example). I also think we need to add an extra warning to the documentation about setting `authorization.openfga.XYZ` and `authorization.scriptlet` not actually doing anything unless one or more `authorization.client` points to it.

That's to avoid folks updating existing scripts that just set the `openfga.XYZ` keys or the scriptlet key who would then just rename them to `authorization.openfga.XYZ` and not understand that they will have all clients allowed because of the `allow` default value.

doc/authorization.md
presztak: Fixed.

PR #3648, opened 2026-07-16
device: Add I/O limits for unix-block devices

fivetime:
## Summary

- add `limits.read` and `limits.write` to `unix-block` devices
- reuse Incus' existing cgroup I/O enforcement path
- apply limits during normal/API device start and optional-device add events
- clear limits when a device is detached while the instance keeps running
- advertise support through `unix_block_limits`

## Lifecycle details

A Cinder-style API attach to a running container executes `Start()`. A `required=false` device whose source appears later executes the registered Unix-device add callback. Both paths resolve the Incus-managed host device node and queue limits for its major:minor. The remove callback and `Stop()` clear `io.max` before removing the node, preventing a later device that reuses the same major:minor from inheriting stale limits.

The implementation is isolated in `unix_block_limits.go`; it does not change the existing disk limit parser and stores no volatile device state.

## Validation

- `go test ./internal/server/device ./internal/version`
- complete `incusd` build
- required/API-attached RBD device: 10 MB/s read and 5 MB/s write limits matched direct-I/O measurements
- delayed `required=false` loop device: limits appeared when the host node was created and disappeared with the container device when the host node was removed

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -3279,3 +3279,9 @@ is controlled by the new `port-forward` agent feature.
 
 A matching `incus port-forward` command is added to the client, providing
 a local TCP listener which forwards every connection to the instance.
+
+## `unix_block_limits`
+
+This adds the `limits.read` and `limits.write` configuration keys to
+`unix-block` devices. These behave similarly to their `disk` device
+equivalents, accepting either a byte/s value or an IOPS value.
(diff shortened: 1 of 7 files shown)


stgraber:
@fivetime commits appear to be missing the required Signed-off-by line, see CONTRIBUTING.md for details.

stgraber (approved):

Issue #2627, opened 2025-11-07
Allow adding custom certificates to the OVMF DB


bensmrs:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

Importing certificates to the SecureBoot DB is a bit tedious, as one has to go through the OVMF UI and manually browse filesystems to find and import certificates.

### What do you think would need to be added

It would be nice to have additional keys, `security.secureboot.db` and `security.secureboot.dbx`, that take a list of certificates to add to the DB and DBX.
The feature would need to rework the NVRAM initialization, although I’m not yet 100% sure of all the implications (and I’m not sure of the available Go tooling for that).

stgraber:
If you can somehow implement it, it'd definitely be nice to have it but last I checked, the NVRAM format wasn't exactly the most trivial thing to interact with from Go :)

bensmrs:
Well at least I can try, you can assign me to it; I may have some time for that in the coming weeks.

stgraber:
I seem to remember $someone having a rough implementation of the NVRAM structures in python. No idea how good those were and if they actually work in all situations given EDK2 has various version of the NVRAM (2MB and 4MB if my memory serves).

stgraber:
https://github.com/awslabs/python-uefivars is what I had in mind

bensmrs:
Ah, we’re allowing 2M images in Incus?

stgraber:
We're trying to avoid them like the plague but we don't always control what a distro may be shipping :)

bensmrs:
Rethinking about it, also adding keys to edit the NVRAM in order to set EFI variables would be great (admittedly, for very niche use cases). If I’m writing an EDK2 NVRAM parser, I might as well exploit it at its full potential. EFI vars have a hierarchical structure (owner GUID or familiar name that translates to its GUID, then variable name) and I’m not completely sure how it should be encoded in Incus configuration format. Would something along the lines of:
```yaml
nvram.{8BE4DF61-93CA-11D2-AA0D-00E0-98032B8C}.Boot0080: base64_encoded_string
```
or its equivalent:
```yaml
nvram.Efi.Boot0080: base64_encoded_string
```
be good?

stgraber:
Given we should be in control of what key namespace is supported (as we're not on hardware with vendor specific extensions), the latter should be fine, though I'd probably want it to be EFI rather than Efi ;)

[...]

PR #3662, opened 2026-07-18
Add NVRAM inspection

bensmrs:
Related-to: #2627


shared/api/instance_nvram.go
```diff
@@ -0,0 +1,39 @@
...
+	"encoding/base64"
+	"time"
+)
+
+// UEFIVar represents a UEFI variable.
+//
+// swagger:model
+//
+// API extension: nvram.
+type UEFIVar struct {
```
stgraber: Should be InstanceNVRAM instead of UEFIVar

shared/uefi/guid.go
```diff
@@ -0,0 +1,175 @@
+package uefi
```
stgraber, line 1: Is this package ever imported by `client/` or something else in `shared/`?
If not, then it should be `internal/uefi`.

From a quick scan in this PR, I'm not seeing any reason for it to be shared, but maybe you have something in subsequent planned PRs which will require it be shared.

client/incus_instances.go
```diff
@@ -3282,3 +3282,102 @@ func (r *ProtocolIncus) CreateInstanceBitmap(name string, bitmap api.StorageVolu
 
 	return nil
 }
+
+func (r *ProtocolIncus) getInstanceNVRAM(name string, guid string, varName string, accept string) (*http.Response, error) {
```
stgraber, line 3286: All of those functions need to have an API extension check added.

shared/api/instance_nvram.go
bensmrs: That’s a single variable, not the whole NVRAM. I also noticed later that I put the wrong extension name.

shared/uefi/guid.go
bensmrs: I’d like the package to be usable outside of Incus actually

client/incus_instances.go
bensmrs: Yeah, noticed that today and fixed in my local working copy.

shared/api/instance_nvram.go
stgraber: InstanceNVRAMVariable or InstanceNVRAMEntry then?

stgraber (changes requested):
I must admit having only skimmed over the very long list of GUIDs and the specific parsers, but that all looks very good to me, just a couple of naming/structure nits.

shared/uefi/guid.go
stgraber: Alright, we can do that for now. I think this is one of those packages that's sufficiently separate from Incus itself that it may warranty its own repository/package, but we can have it as shared/uefi until then.

bensmrs:
Not expecting a very careful review of the dissectors, as it’s an interpretation of a very debatably-specified standard.
I added a few self-imposed nits when working on the followup PR, which I’ll also push.

shared/api/instance_nvram.go
bensmrs: WFM

shared/uefi/guid.go
bensmrs: Yeah, that was the gist of the discussion at the bar that you heard about :)

bensmrs:
Pushed for easy diffing, rebasing now and adding the DELETE endpoint in a bit.

bensmrs:
The deletion endpoint may need further work to clear authenticated variables from the `certdb`, but that’ll be for another PR that will handle signatures and certificates specifically.

PR #3694, opened 2026-07-23
incusd/seccomp: Report FUSE mount helper failures

fivetime:
Report contextual errors from the intercepted FUSE mount helper and retain its stderr in the seccomp debug context. The existing syscall fallback behavior is unchanged. Tested with: go test ./internal/server/seccomp ./cmd/incusd

### DIFF
--- a/internal/server/seccomp/seccomp.go
+++ b/internal/server/seccomp/seccomp.go
@@ -2207,6 +2207,7 @@ func (srv *Server) HandleMountSyscall(c Instance, siov *Iovec) int {
 	}
 
 	if err != nil {
+		ctx["err"] = err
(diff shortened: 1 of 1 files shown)


stgraber:
Reworked this one quite a bit as the logic was pretty wrong.

stderr in forksyscall is meant as a way to send back an errno, it's not meant to be used for general purpose fprintf. The change as it was made would cause a mix of errno and string which could confuse parsers.

The Go error handling logic was also needlessly complex when the main point is to make sure the error is caught and included in the context.

So updated the PR to do just that. If we want more detailed error logging, more significant rework of forksyscall would be needed.

PR #3695, opened 2026-07-23
incusd/instance/lxc: Fall back after CRIU pre-dump failure

fivetime:
Treat CRIU pre-dump as an optional optimization. Remove an incomplete generation, preserve the last successful parent and tell the receiver to proceed to the final dump instead of aborting migration. Tested with: go test ./internal/server/instance/drivers

### DIFF
--- a/internal/server/instance/drivers/driver_lxc.go
+++ b/internal/server/instance/drivers/driver_lxc.go
@@ -6410,7 +6410,7 @@ func (d *lxc) MigrateSend(args instance.MigrateSendArgs) error {
 							return err
 						}
 
-						preDumpDir = fmt.Sprintf("%03d", preDumpCounter)
+						preDumpDir = loopArgs.dumpDir
 						preDumpCounter++
 					}
 				} else {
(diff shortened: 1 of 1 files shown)


stgraber:
@mihalicyn can you take look? I'd like the opinion of CRIU upstream on this one.

mihalicyn:
> Treat CRIU pre-dump as an optional optimization. Remove an incomplete generation, preserve the last successful parent and tell the receiver to proceed to the final dump instead of aborting migration. Tested with: go test ./internal/server/instance/drivers

same question as in #3696 ;)

@fivetime btw, why 2 separate PRs are needed for this?

stgraber:
Going to close both PRs for now as it definitely looks like this is papering over some CRIU issue which we'd rather fix on the CRIU side. If upstream CRIU was recommending the fallback, that'd be different, but it's clearly not the case here.

stgraber:
Going to close both PRs for now as it definitely looks like this is papering over some CRIU issue which we'd rather fix on the CRIU side. If upstream CRIU was recommending the fallback, that'd be different, but it's clearly not the case here.

PR #3696, opened 2026-07-23
incusd/instance/lxc: Retry final CRIU dump without parent

fivetime:
If an incremental final CRIU dump fails while the source remains running, remove the incomplete final dump and retry once without a parent image. Tested with: go test ./internal/server/instance/drivers

### DIFF
--- a/internal/server/instance/drivers/driver_lxc.go
+++ b/internal/server/instance/drivers/driver_lxc.go
@@ -6436,9 +6436,21 @@ func (d *lxc) MigrateSend(args instance.MigrateSendArgs) error {
 						Function:     "migration",
 					}
 
-					// Do the final CRIU dump. This is needs no special handling if
-					// pre-dumps are used or not.
-					dumpSuccess <- d.migrate(&criuMigrationArgs)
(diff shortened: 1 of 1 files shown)


stgraber:
@mihalicyn can you take look? I'd like the opinion of CRIU upstream on this one.

mihalicyn:
Hi @fivetime,

> If an incremental final CRIU dump fails while the source remains running, remove the incomplete final dump and retry once without a parent image. Tested with: go test ./internal/server/instance/drivers

Please could you explain in more detail why this logic is needed? `criu dump` rarely fails, and when it does, it usually means that CRIU does not support certain kernel resources used by the process tree being dumped. In other words, excluding the parent process and performing a full dump does not resolve errors of this type. Otherwise, I would suspect a bug in CRIU itself - such an issue should be properly documented and reported to the developers. Could you tell me more about your specific case?

stgraber:
Going to close both PRs for now as it definitely looks like this is papering over some CRIU issue which we'd rather fix on the CRIU side. If upstream CRIU was recommending the fallback, that'd be different, but it's clearly not the case here.

PR #3702, opened 2026-07-23
Various bugfixes

stgraber:


cmd/incus/info.go
```diff
@@ -328,13 +328,29 @@ func (c *cmdInfo) renderCPU(cpu api.ResourcesCPUSocket, prefix string) {
...
+		if len(clusters) > 1 && (i == 0 || core.Cluster != cpu.Cores[i-1].Cluster) {
+			fmt.Printf(prefix+"  - "+i18n.G("Cluster %d")+"\n", core.Cluster)
+		}
+
+		fmt.Printf(prefix+coreIndent+"  - "+i18n.G("Core %d")+"\n", core.Core)
+		fmt.Printf(prefix+coreIndent+"    "+i18n.G("Frequency: %vMhz")+"\n", core.Frequency)
+		fmt.Print(prefix + coreIndent + "    " + i18n.G("Threads:") + "\n")
 		for _, thread := range core.Threads {
-			fmt.Printf(prefix+"      - "+i18n.G("%d (id: %d, online: %v, NUMA node: %v)")+"\n", thread.Thread, thread.ID, thread.Online, thread.NUMANode)
+			fmt.Printf(prefix+coreIndent+"      - "+i18n.G("%d (id: %d, online: %v, NUMA node: %v)")+"\n", thread.Thread, thread.ID, thread.Online, thread.NUMANode)
```
bensmrs, line 353: I know it was already there, but it feels pretty ugly to me to mix concatenation and `Printf`

bensmrs (approved):
LGTM, although an additional look from someone more well-versed in that dark magic could be good.

cmd/incus/info.go
stgraber: The reason is so we use always the same prefix regardless of whether we fmt.Print or fmt.Printf, then only the rest of the line changes.

PR #3707, opened 2026-07-24
Fix bug in s3 bucket authorization

ra-beer:
When using local s3 buckets (eg btrfs) in clustered mode of incus, bucket permissions are create with the location.
When performing actions that require can_view or can_edit the request for the authorization has no location and fails.
This results in that you can generate buckets but without ability to use, show or edit them.
The fix is oriented at the workaround with volumes for code consistency.
For fairness I don´t test the fix and I get a little bit help from AI.

Signed-off-by: Ramona Beermann <ra-beer@berryit.de>

### DIFF
--- a/cmd/incusd/daemon.go
+++ b/cmd/incusd/daemon.go
@@ -413,7 +413,46 @@ func allowPermission(objectType auth.ObjectType, entitlement auth.Entitlement, m
 			return nodes[0].Name
 		}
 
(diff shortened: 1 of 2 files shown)


stgraber:
Please respect CONTRIBUTING.md, especially the part about our expectation for AI assisted work.

This pull request includes completely hallucinated code as you're calling `GetStorageBucketNodes` which isn't a function that exists at all.

In the future, if you don't understand the code and especially if you can't build and test your fix, please stick to just filing an issue and we'll take it from there.

PR #3715, opened 2026-07-26
storage: Bound Ceph RBD unmap operations

fivetime:
The regular \
bd unmap\ path can block indefinitely when the kernel side of a mapped RBD device is stuck. Bound the normal unmap call, reject timeout-based recovery for mounted filesystem volumes, and fall back to \
bd device unmap --options force\ for detached mapped devices when the regular unmap times out.\n\nA forced RBD unmap can complete asynchronously while open references remain, so this returns once the kernel accepts the removal request instead of looping until EINVAL.\n\nTesting:\n- \go test ./internal/server/storage/drivers\ on Windows: not applicable, Linux/cgo quota package is excluded by build tags.\n- Linux container attempt reached package compilation but the ad-hoc container lacked full Incus CI dependencies, specifically \cowsql.h\. This patch is limited to Ceph helper code and should be covered by upstream CI.

### DIFF
--- a/internal/server/storage/drivers/driver_ceph_utils.go
+++ b/internal/server/storage/drivers/driver_ceph_utils.go
@@ -1,6 +1,7 @@
 package drivers
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -242,7 +243,9 @@ func (d *ceph) rbdUnmapVolume(vol Volume, unmapUntilEINVAL bool) error {
 	ourDeactivate := false
 
 again:
-	_, err := subprocess.RunCommand(
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	_, err := subprocess.RunCommandContext(
+		ctx,
 		"rbd",
 		"--id", d.config["ceph.user.name"],
 		"--cluster", d.config["ceph.cluster_name"],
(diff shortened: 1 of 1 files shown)


stgraber:
@fivetime can you provide more details on your Ceph and Incus deployment, how frequently you've run into issues with the unmap logic and why this is needed?

Because in almost a decade of running LXD and Incus clusters on Ceph, I've never had an issue with `rbd unmap`. The only cases where I've seen it fail are cases where the logic proposed here would at best cause data loss and at worst a full kernel panic.

fivetime:
Thanks, that concern makes sense.

Some context: this did not come from a normal Incus cluster workload. The setup is an OpenStack/Nova integration where each compute node runs a standalone Incus daemon against Ceph RBD-backed storage. We hit this while doing failure-injection around live migration / failed restore cleanup in our fork, including cases where a target had created or claimed an RBD mapping and then the migration path failed before normal teardown completed.

So far this is not something we have seen frequently in normal operation; it was exposed by repeated destructive migration tests. The intent of the patch was to avoid an incusd operation being stuck forever behind a wedged `rbd unmap` subprocess.

I agree that the `--options force` fallback is too aggressive for upstream generic Incus semantics. Even if the filesystem mount check passes, there can still be open references or kernel-side state that make forcing the device removal unsafe.

Would a narrower version be acceptable that only bounds the regular `rbd unmap` call and returns a clear timeout error, without attempting `rbd device unmap --options force`? If not, I can close this PR and keep any force-unmap handling as a site-specific/fork operational workaround rather than upstream behavior.

stgraber:
Yeah, we definitely don't want an `rbd unmap` to be stuck forever, so having it timeout after 30s would be appropriate here.

Issue #3513, opened 2026-06-20
Disable NAT between incus bridges on the same host


candlerb:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

Currently, traffic which originates on one incus bridge, and is routed to another incus bridge on the same host, is NAT'd to the incus host's own address.

Specific example:

* bridge `incusbr0` has IP address 100.125.0.1/22
* bridge `wifi0` has IP address 100.126.0.1/22

The problem is that traffic from any container on `wifi0`, to any container on `incusbr0`, has its source address changed from 100.126.X.X to 100.125.0.1

This obscures the true source in logs, makes debugging harder, and breaks anything which depends on seeing the true client IP address (the particular example I hit was RADIUS clients.conf)

### What do you think would need to be added

At the moment, incus creates NAT rules like this:

```
table inet incus {
	chain pstrt.incusbr0 {
		type nat hook postrouting priority 100; policy accept;
		ip saddr 100.125.0.0/22 ip daddr != 100.125.0.0/22 masquerade
...
	chain pstrt.wifi0 {
		type nat hook postrouting priority 100; policy accept;
		ip saddr 100.126.0.0/22 ip daddr != 100.126.0.0/22 masquerade
```

This means that traffic leaving any bridge is NAT'd, even if it's going to another incus bridge.

I can see some possibilities for how to change this.

## 1. Make incus no-NAT between its own bridges

This could be implemented by:

* Creating an nftables [named set](https://wiki.nftables.org/wiki-nftables/index.php/Sets) containing all the bridge subnets
* Change the generated rules to use that address set

```
table inet incus {
	chain pstrt.incusbr0 {
		type nat hook postrouting priority 100; policy accept;
		ip saddr 100.125.0.0/22 ip daddr != @incus masquerade
```

The address set would have to be updated when incus bridges are created, destroyed or changed.

This is technically a backwards-incompatible change (does anybody *depend* on the current NAT behaviour??) but it could be put behind a flag

## 2. Make the incus no-NAT configurable per network

For example:

```
incus network set incusbr0 ipv4.nat.exclude_destinations="100.64.0.0/0 @myownset"
```

This configuration would override the auto-generated rule to:

```
table inet incus {
	chain pstrt.incusbr0 {
		type nat hook postrouting priority 100; policy accept;
		ip saddr 100.125.0.0/22 ip daddr != 100.64.0.0/0 ip daddr != @myownset masquerade
```

[...]

PR #3721, opened 2026-07-27
Fix NAT rule duplication and stale exclusions between managed bridge networks

jochumdev:
Since #3513, every managed bridge carries a NAT exclusion rule per
sibling bridge, and starting or reconfiguring any bridge re-applies the
NAT rules of all the others. That refresh appended to the existing
chain rather than replacing it, so the pstrt chains of the other
bridges gained a duplicate rule set each time, growing without bound.
The template now uses the same add/flush pattern as the proxy NAT and
ACL templates.

Deleting a bridge also left its exclusion rule behind in the other
bridges' chains. Beyond the leftover rule, an unmanaged interface later
created under the same name would have its traffic skip NAT entirely.
The remaining bridges are now refreshed on deletion.


stgraber:
I submitted an alternate (much much smaller) implementation.

PR #3728, opened 2026-07-27
Add support for ACME EAB

### LINKED ISSUE #2183: [ACME] pass raw parameter to the lego binary

`incus` currently restricts configuring ACME requests with a few `options`. I would like to have more flexibility in managing certificate updates from my private CA.
For example I would like to us `ACME External Account Binding (eab)` in my environment, and have ability to issue `incus cluster certificate` with multiple SANs.

Adding `acme.lego_raw_params` option and passing this string to the `lego` command-line would be helpful.

stgraber:
Closes #2183

### DIFF
--- a/doc/.wordlist.txt
+++ b/doc/.wordlist.txt
@@ -80,6 +80,7 @@ EB
 Ebit
 eBPF
 ECDHE
+EAB
 ECDSA
 ECMP
 EDK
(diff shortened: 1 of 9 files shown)


stgraber:
Since both requests in the Lego issue were about enabling EAB, it felt like it'd be a fair bit cleaner to just support that natively rather than allow arbitrary arguments to be passed through.

bensmrs (approved):

Issue #3059, opened 2026-03-20
Near-live migration for containers


lady-melinoe:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

Currently, there isn't a good solution for moving large containers quickly and with low downtime (other than getting better network cards or a better internet plan lol)

### What do you think would need to be added

Something along the lines of:

snapshot  the container
send initial snapshot to target (e.g. btrfs send/receive)
repeatedly take incremental snapshots and keep sending deltas until the remaining diff is small
stop the container
take final snapshot and send it over
start the container up on the target host

This should, based on my (admittedly janky) testing, significantly reduce downtime compared to a full shutdown-move-start loop, and also would work on instances with things that CRIU doesn't support, like instances with network interfaces)

I’ve only tested this kind of workflow manually with btrfs, but I assume similar approaches could work with other backends that support incremental send (e.g. ZFS).

lady-melinoe:
(I accidentally posted this issue at https://github.com/lxc/incus-os/issues/981 first and then realised i misclicked the wrong repo. oof.)

stgraber:
incus copy source:my-container target:my-container --refresh
incus copy source:my-container target:my-container --refresh
incus stop source:my-container
incus copy source:my-container target:my-container --refresh
incus start target:my-container
incus delete -f source:my-container

bensmrs:
We may be able to walk around this one by allowing to define aliases as sequences of commands, hence allowing OP to define an `incremental-migrate` alias doing essentially what you wrote.

lady-melinoe:
~~Is there a good way to get that to work with a cluster? i tried while using --target, and with target:, but I can't seem to get it to work lol.~~

Edit: i managed to make a script that works well enough for my use case, but something with the safeguards built in and not relying on janky renaming would be nice

```
#!/usr/bin/env bash

if [ -z "$1" ] && [ -z "$2" ]; then
    echo "Usage: $0 <source-container> <target-host>"
    echo "!!! Error: Missing both source and target arguments."
    exit 1
fi

if [ -z "$1" ]; then
    echo "Usage: $0 <source-container> <target-host>"
    echo "!!! Error: Missing source container argument."
    exit 1
fi

if [ -z "$2" ]; then
    echo "Usage: $0 <source-container> <target-host>"
    echo "!!! Error: Missing target host argument."
    exit 1
fi

src="$1"
tgt="$2"

[...]

PR #3730, opened 2026-07-28
Server-side near-live migration

presztak:
Changes:

- Add `instanceShutdownOrForceStop`, which attempts a graceful instance shutdown and falls back to a forceful stop if that fails
- Allow cluster-internal instance changes on evacuated members
- Implement server-side near-live migration
- Add support for near-live migration during evacuation
- Block near-live migration for instances with dependent volumes

Part of: #3059


stgraber:
Going to hold that one for 7.4 just to be on the safe side.

cmd/incusd/instance_post_nearlive.go
```diff
@@ -0,0 +1,265 @@
...
+	l.Debug("Starting instance on target member")
+
+	startOp, err := target.UpdateInstanceState(instName, api.InstanceStatePut{Action: "start"}, "")
+	if err == nil {
+		err = startOp.Wait()
+	}
+
+	if err != nil {
+		return fmt.Errorf("Failed starting instance on target member: %w", err)
+	}
```
stgraber: The double Debug logging seems a bit odd.

stgraber:
Let's make this opt-in, we'll want a `cluster.evacuation=refresh-migrate` which is fully opt-in for now. Then at the CLI level, we'll want `--stateless --refresh` combined with a running instance to kick in the process, both server to server (standalone) and within a cluster.

cmd/incusd/instance_post_nearlive.go
presztak: Fixed.

internal/server/instance/drivers/util.go
```diff
@@ -127,6 +131,48 @@ func ParseMemoryStr(memory string) (valueInt int64, err error) {
...
+	pool, err := storagePools.LoadByName(s, rootDiskConf["pool"])
+	if err != nil {
+		return fmt.Errorf("Failed loading instance storage pool: %w", err)
+	}
+
+	driverName := pool.Driver().Info().Name
+	if driverName == "dir" || driverName == "lvm" {
+		return fmt.Errorf("Near-live migration isn't supported on %q storage pools", driverName)
+	}
+
```
stgraber: Also need the same check on all the dependent volumes.
Should also use a proper boolean flag in the driver Info rather than hardcode a list of driver names here.

internal/server/instance/drivers/driver_common.go
```diff
@@ -1064,6 +1064,10 @@ func (d *common) canMigrate(inst instance.Instance) string {
 		return "live-migrate"
 	}
 
+	if NearLiveMigrationSupported(d.state, inst) == nil {
+		return "refresh-migrate"
+	}
+
```
stgraber: I thought we said we'd make it opt-in for now?

If it's meant to be opt-in (through cluster.evacuate) then it shouldn't be returned here.

internal/server/instance/drivers/util.go
stgraber: Actually, you said dependent volumes will be in a follow-up PR. So I guess for now we should error-out if the instance has any dependent volumes.

stgraber: I see you have the check in https://github.com/lxc/incus/pull/3730/changes#diff-b37ea46fd3fcac113ccb2de3a8403c463ea787beeeccc5bb9be091313090ccf9R150 so that may be enough.

presztak: Yes, this is enough for now, but you're reminding me that we need to add this condition for every dependent disk in my follow-up PR. Thanks.

internal/server/instance/drivers/driver_common.go
presztak: Yep, this is what we agreed. I added `refresh-migrate` as an option for `cluster.evacuate` but I forgot to remove that condition. Fixed.

stgraber (approved):

PR #3734, opened 2026-07-28
incus/launch: Link to the supported instance types

gagath:
Having to consult the documentation to find the list of supported instances is a bit painful. Add a link to the supported instances in the help message as well to avoid having to dig through the documentation first.

### DIFF
--- a/cmd/incus/launch.go
+++ b/cmd/incus/launch.go
@@ -34,6 +34,7 @@ incus launch images:debian/12 c2 < config.yaml
 
 incus launch images:debian/12 c3 -t aws:t2.micro
     Create and start a container named "c3" using the same size as an AWS t2.micro (1 vCPU, 1GiB of RAM)
+    Find the list of supported instance types here: https://images.linuxcontainers.org/meta/instance-types/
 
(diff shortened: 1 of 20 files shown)


cmd/incus/launch.go
```diff
@@ -34,6 +34,7 @@ incus launch images:debian/12 c2 < config.yaml
 
 incus launch images:debian/12 c3 -t aws:t2.micro
     Create and start a container named "c3" using the same size as an AWS t2.micro (1 vCPU, 1GiB of RAM)
+    Find the list of supported instance types here: https://github.com/dustinkirkland/instance-type
```
stgraber: This should link to https://images.linuxcontainers.org/meta/instance-types/ instead as Dustin's repo is quite outdated and we're looking at refreshing that date ourselves using current cloud providers documentation.

gagath: This page is so much better, thanks! It also feels more official and stable enough to be mentioned in the CLI help. I will update the PR soon-ish to use this new link for both the command line help as well as the current documentation.

gagath:
Thanks!

Issue #3735, opened 2026-07-28
RunningInUserNS can misdetect systemd PrivateUsers=full as the initial user namespace


aleivag:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

I ran into this while debugging container startup failures under systemd 260 with `PrivateUsers=full`.

The current user namespace detection logic infers whether the process is in the initial user namespace by reading `/proc/self/uid_map` and checking if 4294967295 is on it.

That assumes the only namespace that can contain 4294967295 UIDs is the initial namespace. However, uid_map describes ID mapping, not namespace identity. A non-initial user namespace can have a full identity map.

At meta we hit this hard while upgrading systedm >=260... I already went and fixed crun for podman and while fixiing moby/sys (https://github.com/moby/sys/issues/239#issuecomment-5097002354) for runc and docker, they recommended to also come here to fix this, so here i am 

### What do you think would need to be added

The more robust check seems to be comparing /proc/self/ns/user to the kernel's initial user namespace inode, PROC_USER_INIT_INO (0xEFFFFFFD), rather than parsing /proc/self/uid_map. This is what systemd does, and (now after https://github.com/containers/crun/issues/2150) is what crun

PR #3736, opened 2026-07-28
internal/linux: detect initial user namespace by inode

aleivag:
The current uid_map heuristic treats a private user namespace with an identity uid_map as the initial user namespace. systemd 260 can create that shape with PrivateUsers=full, causing callers to believe they can perform initial-namespace-only operations.

Use /proc/self/ns/user's inode instead and compare it with the kernel's PROC_USER_INIT_INO value. This detects the initial user namespace directly instead of inferring it from the uid_map layout.

Related: https://github.com/lxc/incus/issues/3735

### DIFF
--- a/internal/linux/userns.go
+++ b/internal/linux/userns.go
@@ -1,34 +1,20 @@
 package linux
 
 import (
-	"bufio"
-	"fmt"
-	"os"
-
-	"github.com/lxc/incus/v7/shared/logger"
+	"syscall"
 )
+
+// See PROC_USER_INIT_INO in include/uapi/linux/nsfs.h.
+const procUserInitIno = 0xEFFFFFFD
 
 // RunningInUserNS returns true if the current process is running inside a user namespace.
 func RunningInUserNS() bool {
-	file, err := os.Open("/proc/self/uid_map")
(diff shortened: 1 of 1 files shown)


internal/linux/userns_test.go
```diff
@@ -0,0 +1,31 @@
+package linux
```
stgraber: Drop the pointless test, we don't need a test to make sure that "1 != 2" :)

internal/linux/userns.go
```diff
@@ -1,34 +1,22 @@
...
-	var a, b, c int64
-	_, _ = fmt.Sscanf(line, "%d %d %d", &a, &b, &c)
-	if a == 0 && b == 0 && c == 4294967295 {
-		return false
-	}
+	return inodeInUserNS(st.Ino)
+}
 
-	return true
+func inodeInUserNS(inode uint64) bool {
```
stgraber: With the test dropped, just move that into RunningInUserNS.

aleivag: yup, without the tests... makes no sense having this here... I updated the code

stgraber (approved):

PR #3742, opened 2026-07-29
Fix SFTP connections

jochumdev:
Wait a full minute instead of 10s for sFTP connections and Align the buffer/package size used by the client and server (cmd/incusd, incus-agent) for sFTP transfers.

The first fixes slow clients in overloaded tests, the second WOULD fix `io.Copy()` breaking at ~800 MiB of a transfer.

Related: https://github.com/pkg/sftp/issues/658

### DIFF
--- a/cmd/incus-agent/sftp.go
+++ b/cmd/incus-agent/sftp.go
@@ -71,7 +71,7 @@ func (r *sftpServe) Render(w http.ResponseWriter) error {
 	}
 
 	// Start sftp server.
-	server, err := sftp.NewServer(conn, sftp.WithAllocator(), sftp.WithServerWorkingDirectory(osBaseWorkingDirectory))
+	server, err := sftp.NewServer(conn, sftp.WithAllocator(), sftp.WithServerWorkingDirectory(osBaseWorkingDirectory), sftp.WithMaxTxPacket(128*1024))
 	if err != nil {
 		return nil
 	}
--- a/cmd/incusd/main_forkfile.go
+++ b/cmd/incusd/main_forkfile.go
@@ -237,7 +237,7 @@ func (c *cmdForkfile) run(_ *cobra.Command, args []string) error {
 			mu.Unlock()
 
 			// Spawn the server.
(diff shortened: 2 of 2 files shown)


cmd/incusd/main_forkfile.go
```diff
@@ -160,7 +160,7 @@ func (c *cmdForkfile) run(_ *cobra.Command, args []string) error {
 	// Automatically shutdown after inactivity.
 	go func() {
 		for {
-			time.Sleep(10 * time.Second)
+			time.Sleep(time.Minute)
```
stgraber: Nope, we're not going to be doing that. It's already annoying enough having other operations get stuck on a leftover forkfile daemon for a few seconds, we definitely don't want to start waiting an entire minute since the last connection terminated.

cmd/incusd/main_forkfile.go
```diff
@@ -237,7 +237,8 @@ func (c *cmdForkfile) run(_ *cobra.Command, args []string) error {
 			mu.Unlock()
 
 			// Spawn the server.
-			server, err := sftp.NewServer(conn, sftp.WithAllocator())
+			// Allow responses large enough for the 128KiB read requests sent by the Incus client.
```
stgraber: No need for that comment, the code clearly already states the same.

cmd/incus-agent/sftp.go
```diff
@@ -71,7 +71,8 @@ func (r *sftpServe) Render(w http.ResponseWriter) error {
 	}
 
 	// Start sftp server.
-	server, err := sftp.NewServer(conn, sftp.WithAllocator(), sftp.WithServerWorkingDirectory(osBaseWorkingDirectory))
+	// Allow responses large enough for the 128KiB read requests sent by the Incus client.
```
stgraber: Same, no need for the comment.

cmd/incusd/main_forkfile.go
jochumdev: Why are there leftover forkfiles? Do we not close them as we should?

PR #3743, opened 2026-07-29
More speed for compression/decompression with pigz

jochumdev:
Switch gzip usage to klauspost/pgzip (parallel gzip) in the client,
incusd response rendering, and the qemu driver's state
save/restore path for faster compression.

Prefer the external `pigz` binary for tar.gz decompression when
available, and treat its output as reproducible the same as gzip's
when publishing images.

### DIFF
--- a/cmd/incusd/images.go
+++ b/cmd/incusd/images.go
@@ -133,7 +133,12 @@ var imagePublishLock sync.Mutex
 var imageTaskMu sync.Mutex
 
 func compressFile(compress string, infile io.Reader, outfile io.Writer) error {
-	reproducible := []string{"gzip"}
+	// Compressors with reproducible output and the flags needed for it.
+	reproducible := map[string][]string{
+		"gzip": {"-n"},
+		"pigz": {"-n", "-m"},
+	}
+
 	var cmd *exec.Cmd
 
 	// Parse the command.
@@ -183,8 +188,9 @@ func compressFile(compress string, infile io.Reader, outfile io.Writer) error {
 			args = append(args, fields[1:]...)
 		}
 
-		if slices.Contains(reproducible, fields[0]) {
-			args = append(args, "-n")
+		flags, ok := reproducible[fields[0]]
+		if ok {
+			args = append(args, flags...)
 		}
 
 		cmd := exec.Command(fields[0], args...)
--- a/cmd/incusd/images.go
+++ b/cmd/incusd/images.go
@@ -188,6 +188,15 @@ func compressFile(compress string, infile io.Reader, outfile io.Writer) error {
 			args = append(args, fields[1:]...)
 		}
 
+		// Prefer pigz over gzip when available.
+		if fields[0] == "gzip" {
+			_, err := exec.LookPath("pigz")
+			if err == nil {
+				fields[0] = "pigz"
+				args = append(args, "-p", strconv.Itoa(archive.CompressionThreads()))
+			}
+		}
+
 		flags, ok := reproducible[fields[0]]
 		if ok {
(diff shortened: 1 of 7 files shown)


cmd/incusd/images.go
```diff
@@ -133,7 +133,7 @@ var imagePublishLock sync.Mutex
 var imageTaskMu sync.Mutex
 
 func compressFile(compress string, infile io.Reader, outfile io.Writer) error {
-	reproducible := []string{"gzip"}
+	reproducible := []string{"gzip", "pigz"}
```
stgraber: That's incorrect, `pigz -n` isn't reproducible, `pigz -nm` may be reproducible but that's not how the code will invoke it.

jochumdev: I'll fix that.

jochumdev:
I'm using this to move everything else to pigz:

```
incus config set images.compression_algorithm=pigz
incus config set backups.compression_algorithm=pigz
```

Do you think we can detect if `pigz` is installed and replace `gzip` automatically with that?

stgraber:
So in general, I'm opposed to this change. Incus is used in pretty large and busy environment, the last thing you want in such an environment is for a user's random non-urgent request like a backup or image download to suddenly spawn 128 goroutines and add load to every single CPU thread on the system.

I don't mind that users set it directly through `compression_algorithm` when it's available on the system (it often isn't as `gzip` is usually the security supported compressor) but having it applied everywhere is going to trade overall running instance experience and debugability for speed of non-urgent operations which I don't think is a good idea.

stgraber:
So currently from that branch, I'd only consider including the reproducibility change for pigz, once it's corrected to pass the correct flags.

All the other codepaths are used by the incus daemon and will lead to a large number of goroutines being spawned, making server debugging quite messy (we commonly graph and dump the list of goroutines when investigating cluster performance issues) and will cause a significantly larger system impact during background image updates or similar operations due to now loading every CPU thread rather than just one, actively slowing down every instance on the system rather than allowing the scheduler to deal with a single busy thread.

jochumdev:
I see the problem with `pigz` overloading the OS, can you think of a standardized way so we can set `-p` with it? Maybe as well with a `nice`?

Can we switch the oci decompress path to use `images.compression_algorithm` as well?

stgraber:
I think a sane pigz limit would probably be:
 - Host has 1 to 3 threads => 1
 - Host has 4 to 15 threads => 2
 - Host has 16 threads or more => 4

If we know that `images.compression_algorithm` is `pigz` then using it with the limits above during decompress would be fine. The problem is going to be how to plumb that from the global database (either server or project config depending) through to `archive.Unpack`.

Probably pass it as a preferredCompressor or something, the problem will be with DetectCompress as that's called from quite a lot of different places, including other external projects, so not super keen on changing that function's signature.

That means we'd need Unpack to handle the substitution which is really making this whole thing uglier and uglier...

jochumdev:
Can't we auto opt-in for pigz in case it is installed?

jochumdev:
NICE! Many Thanks!

jochumdev:
Is it my output or do I miss something here?

swappy-20260730_174527

```sh
root@ict-daily:~# incus version
Client version: 7.2
Server version: 7.2
root@ict-daily:~# incus config show
config:
  core.https_address: '[::]:8443'
```

stgraber:
pigz never does parallel decompression, only compression.

stgraber:
decompression is always two threads, so still faster than gzip, but doesn't need a flag to limit it.

jochumdev:
Then sorry for the noise.

PR #3747, opened 2026-07-30
Fix resource leak, deadlock and bulk query performance

stgraber:


stgraber:
@masnax can you take a look at the DB code? I'm not particularly good at messing with the generator though the output looks correct and does allow for the `OBJECT_id IN (1,2,3)` WHERE clause that we want to get much better performance on those queries.

cmd/incusd/instance_patch.go
```diff
@@ -200,12 +200,17 @@ func instancePatch(d *Daemon, r *http.Request) response.Response {
 			return err
 		}
 
-		profileConfigs, err := cluster.GetAllProfileConfigs(ctx, tx.Tx())
+		profileIDs := make([]int, 0, len(profiles))
+		for _, profile := range profiles {
+			profileIDs = append(profileIDs, profile.ID)
+		}
+
+		profileConfigs, err := cluster.GetConfig(ctx, tx.Tx(), "profiles", "profile", cluster.ConfigFilter{ReferenceID: profileIDs})
```
bensmrs: Those repeated calls could deserve their own `GetReferencedProfileXXX` or something like that.

bensmrs (commented):
Not approving because I’m not at ease with the DB code, which needs a more experienced reviewer.

internal/server/db/cluster/devices.mapper.go
```diff
@@ -133,17 +146,35 @@ func GetDevices(ctx context.Context, db tx, parentTablePrefix string, parentColu
...
 
 			configFilters = append(configFilters, *filter)
 		}
 	}
 
-	config, err := GetConfig(ctx, db, parentTablePrefix+"_devices", parentColumnPrefix+"_device", configFilters...)
-	if err != nil {
-		return nil, err
+	config := map[int]map[string]string{}
+	if len(filters) == 0 || len(objects) > 0 {
```
masnax: `if len(filters) == 0` seems unnecessary here. If `objects` is empty, the result of `GetConfig` will never be returned anyway, so no point in performing the query.

stgraber: done!

cmd/incusd/instance_patch.go
stgraber: added a couple of helpers to handle that!

bensmrs (approved):
If the DB stuff is good, then LGTM

PR #3748, opened 2026-07-30
Add NVRAM rebuild

bensmrs:


bensmrs:
I can rename `rebuild` to `reset` if that’s more suitable.

bensmrs:
Thanks!

stgraber:
As discussed, I've moved the logic to be a repair action accessible through `incus low-level repair NAME rebuild-nvram`. The rest of the logic is identical. I've updated the API extension text to match.

PR #3750, opened 2026-07-30
Security fixes for Incus 7.3

stgraber:


cmd/incusd/daemon.go
```diff
@@ -352,24 +358,26 @@ func allowPermission(objectType auth.ObjectType, entitlement auth.Entitlement, m
 			}
 
 			if objectType == auth.ObjectTypeProfile {
-				projectName = project.ProfileProjectFromRecord(p)
+				return project.ProfileProjectFromRecord(p)
 			} else if objectType == auth.ObjectTypeStorageBucket {
```
bensmrs, line 362: Nit: you can drop the `else`s

bensmrs (approved):
Approving since my remark isn’t that important.

cmd/incusd/daemon.go
stgraber: Ah yeah, good point. Kinda surprised that golangci-lint didn't complain :)

PR #3753, opened 2026-07-31
Update gomod

stgraber:


stgraber:
I'd normally just push those directly as part of the release, but the go.mod update list is long enough that I want to see a test run.

PR #3754, opened 2026-07-31
Client-side near-live migration

presztak:
Near-live migration for the cluster <-> cluster case.

Part of: https://github.com/lxc/incus/issues/3059

### DIFF
--- a/cmd/incus/copy.go
+++ b/cmd/incus/copy.go
@@ -347,6 +347,11 @@ func (c *cmdCopy) copyOrMove(cmd *cobra.Command, src *u.Parsed, dst *u.Parsed, k
 			dstServer = dstServer.UseTarget(c.flagTarget)
 		}
 
+		// A stateless move of a running container is performed as a near-live migration.
+		if srcServer.HasExtension("container_incremental_copy") && dstServer.HasExtension("container_incremental_copy") && move && !stateful && c.flagRefresh && !instanceOnly && entry.StatusCode == api.Running && entry.Type == "container" {
+			return c.nearLiveMoveInstance(srcServer, dstServer, srcInstanceName, dstInstanceName, entry, &args)
+		}
+
 		op, err = dstServer.CopyInstance(srcServer, *entry, &args)
 		if err != nil {
 			return err
--- a/cmd/incus/move_nearlive.go
+++ b/cmd/incus/move_nearlive.go
@@ -1,0 +1,172 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	incus "github.com/lxc/incus/v7/client"
+	"github.com/lxc/incus/v7/cmd/incus/color"
+	"github.com/lxc/incus/v7/internal/i18n"
+	internalUtil "github.com/lxc/incus/v7/internal/util"
+	"github.com/lxc/incus/v7/shared/api"
+	cli "github.com/lxc/incus/v7/shared/cmd"
+	"github.com/lxc/incus/v7/shared/revert"
+)
+
+// nearLiveMoveDeleteSnapshot removes a pre-copy snapshot.
+func nearLiveMoveDeleteSnapshot(server incus.InstanceServer, instName string, snapName string, warn bool) {
+	op, err := server.DeleteInstanceSnapshot(instName, snapName)
+	if err == nil {
+		err = op.Wait()
+	}
+
+	if err != nil && warn {
+		fmt.Fprintf(os.Stderr, "%s "+i18n.G("Failed to delete pre-copy snapshot %q of instance %q: %s")+"\n", color.WarningPrefix, snapName, instName, err)
+	}
+}
+
+// nearLiveMoveCopyRound performs a single copy of the instance onto the target server.
+func (c *cmdCopy) nearLiveMoveCopyRound(srcServer incus.InstanceServer, dstServer incus.InstanceServer, entry *api.Instance, args *incus.InstanceCopyArgs, format string) error {
+	op, err := dstServer.CopyInstance(srcServer, *entry, args)
+	if err != nil {
+		return err
+	}
+
+	progress := cli.ProgressRenderer{
+		Format: format,
+		Quiet:  c.global.flagQuiet,
+	}
+
+	_, err = op.AddHandler(progress.UpdateOp)
+	if err != nil {
+		progress.Done("")
+		return err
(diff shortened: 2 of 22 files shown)


cmd/incus/move_nearlive.go
```diff
@@ -0,0 +1,215 @@
...
+	"github.com/lxc/incus/v7/cmd/incus/color"
+	"github.com/lxc/incus/v7/internal/i18n"
+	internalUtil "github.com/lxc/incus/v7/internal/util"
+	"github.com/lxc/incus/v7/shared/api"
+	cli "github.com/lxc/incus/v7/shared/cmd"
+	"github.com/lxc/incus/v7/shared/revert"
+)
+
+// nearLiveMovePrecopyRounds is how many times the instance is copied while it keeps running.
+const nearLiveMovePrecopyRounds = 2
```
stgraber: No need for a const when it's only used in exactly one place.

cmd/incus/move_nearlive.go
```diff
@@ -0,0 +1,215 @@
...
+
+// nearLiveMoveInstanceUUID returns the instance's UUID, generating a random one if it has none.
+func nearLiveMoveInstanceUUID(entry *api.Instance) (string, error) {
+	instUUID := entry.Config["volatile.uuid"]
+	if instUUID != "" {
+		return instUUID, nil
+	}
+
+	return internalUtil.RandomHexString(16)
+}
```
stgraber: Not sure a standalone function is needed when used in exactly one place.

cmd/incus/move_nearlive.go
```diff
@@ -0,0 +1,215 @@
...
+
+// nearLiveMoveSnapshotName returns the name to use for the pre-copy snapshot of a round.
+func nearLiveMoveSnapshotName(instUUID string, round int) string {
+	return fmt.Sprintf("move-of-%s-%d", instUUID, round)
+}
```
stgraber: Same with this one.

cmd/incus/move_nearlive.go
```diff
@@ -0,0 +1,215 @@
...
+		err = op.Wait()
+	}
+
+	if err != nil && warn {
+		fmt.Fprintf(os.Stderr, "%s "+i18n.G("Failed to delete pre-copy snapshot %q of instance %q: %s")+"\n", color.WarningPrefix, snapName, instName, err)
+	}
+}
+
+// nearLiveMoveDeleteInstance removes a partially transferred copy.
+func nearLiveMoveDeleteInstance(server incus.InstanceServer, name string) {
```
stgraber: Same here

cmd/incus/move_nearlive.go
```diff
@@ -0,0 +1,215 @@
...
+		err = op.Wait()
+	}
+
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "%s "+i18n.G("Failed to delete partial copy of instance %q on the target server: %s")+"\n", color.WarningPrefix, name, err)
+	}
+}
+
+// nearLiveMoveStopInstance cleanly shuts the instance down, falling back to a forced stop.
+func nearLiveMoveStopInstance(server incus.InstanceServer, name string) error {
```
stgraber: Same here

cmd/incus/move_nearlive.go
```diff
@@ -0,0 +1,215 @@
...
+
+	if err != nil {
+		return fmt.Errorf(i18n.G("Failed to stop instance %q: %w"), name, err)
+	}
+
+	return nil
+}
+
+// nearLiveMoveSupported checks whether a stateless move of a running container can be performed as a near-live migration.
+func (c *cmdCopy) nearLiveMoveSupported(srcServer incus.InstanceServer, dstServer incus.InstanceServer, entry *api.Instance, move bool, stateful bool, instanceOnly bool, refresh bool) (bool, error) {
```
stgraber: Same

stgraber:
Logic generally seems fine but the over-use of small functions actually makes it harder for me to follow and understand the flow.

The use of functions and const all makes sense if they're called multiple times (we don't want to duplicate code or const) but it's not the case here and it just makes it harder to follow.

cmd/incus/move_nearlive.go
presztak: Done.

cmd/incus/move_nearlive.go
presztak: Done.

cmd/incus/move_nearlive.go
presztak: Done.

cmd/incus/move_nearlive.go
presztak: Fixed.

cmd/incus/move_nearlive.go
presztak: Done.

cmd/incus/move_nearlive.go
presztak: Fixed.

cmd/incus/copy.go
```diff
@@ -347,6 +347,18 @@ func (c *cmdCopy) copyOrMove(cmd *cobra.Command, src *u.Parsed, dst *u.Parsed, k
...
+		nearLive := false
+		if srcServer.HasExtension("container_incremental_copy") && dstServer.HasExtension("container_incremental_copy") {
+			if move && !stateful && c.flagRefresh && !instanceOnly && entry.StatusCode == api.Running && entry.Type == "container" {
+				nearLive = true
+			}
+		}
+
+		if nearLive {
+			return c.nearLiveMoveInstance(srcServer, dstServer, srcInstanceName, dstInstanceName, entry, &args)
+		}
```
stgraber: Wouldn't this be much simpler if you just called the `return c.nearLiveMoveInstance` from the spot where you currently set `nearLive = true`?

You wouldn't need the variable at all or the second if statement.
The nested if statement can also be folded into the first if statement.

presztak: ah, I jumped to fix this too quickly, but the nested if statement was intentional. I thought it would be cleaner since we have several conditions and wanted to keep them separated.

stgraber:
@presztak in your follow-up PR can you add system testing for this stuff? It should be possible to test it easily enough when running the testsuite on ZFS or btrfs.

PR #3755, opened 2026-07-31
docs: Clarify the sentence about Incus owning the ZFS pool/dataset

ldesgoui:
Hello,
The original sentence has tripped me up and it seems I'm not the only one because I can see it quoted a few times on the forums.
I don't think this change is perfect but I feel like it improves the clarity a little bit.
Thank you


ldesgoui:
Ah, I manually added Signed-off-by in the commit description but Github also did, I'll clone and fix the commit

stgraber (approved):

PR #3762, opened 2026-08-01
Various bugfixes

stgraber:


cmd/incus/remote_unix.go
```diff
@@ -283,8 +283,8 @@ func (h remoteProxyHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	// Handle /1.0 internally (saves a round-trip).
-	if r.RequestURI == "/1.0" || strings.HasPrefix(r.RequestURI, "/1.0?project=") {
+	// Handle GET /1.0 internally (saves a round-trip).
+	if r.Method == http.MethodGet && (r.RequestURI == "/1.0" || strings.HasPrefix(r.RequestURI, "/1.0?project=")) {
```
bensmrs: Why check `RequestURI` and not `Path`?

bensmrs (commented):
I’m honestly not sure I get the caching thing. A server update doesn’t invalidate the cache, does it? Is the road-trip discount worth a wrong server struct reporting?

stgraber:
Caching /1.0 cuts down the number of requests and therefore latency in half, it's the most significant speed improvement that comes from using the keep-alive feature. 

It reporting slightly outdated /1.0 is usually not a problem. The big problem was that it was swallowing non-GET requests. 

Then since that needed fixing, I figured we may was well invalidate the cache when we are the ones making a direct change to the /1.0 data.

The keep-alive proxy is usually meant to run for up to a minute (I do 10s) after its last request, so once the user's script or Ansible runs completes, it exits and we'll get a clean state next time around.

cmd/incus/remote_unix.go
stgraber: We want to detect specific full patterns in the URL. Path doesn't include the parameters. 

Specifically, we don't want to cache a request that contains ?target

bensmrs: And would this code actually prevent doing `?project=foo&target=bar`?

bensmrs:
Thanks for the clarification, I forgot the keepalive duration.

stgraber:
@bensmrs did you mean to approve?

bensmrs:
I commented your reply in the discussion :)

stgraber:
Oh Github, didn't update the discussion, had to refresh, fun fun

cmd/incus/remote_unix.go
stgraber: Hmm, nope and I just checked, that would actually be the order of the parameters that would be sent...

Alright, guess I'll do it with full parameters parsing then ;)

stgraber: Fixed

stgraber:
@bensmrs alright, got that one sorted and added another fix I've come up with following some customer related fun earlier today.

bensmrs (approved):

PR #3764, opened 2026-08-03
Fix/oci dns

jochumdev:


jochumdev:
My decision to not update `resolv.conf` when any `oci.dns` key is present isn't that good, how do people resolve without knowing the gateway?

We could add all networks the instance is connected to, to `oci.dns.nameservers` but that is a decision I don't want to take.

PR #3767, opened 2026-08-03
incusd/cluster: Stop instances in place when evacuation has no target

stgraber:


stgraber:
That's pretty useful for me as I have a cluster with mixed architectures. For arm64 I only have a single server, so there is nowhere for it to evacuate to.

bensmrs (approved):

PR #3768, opened 2026-08-03
Implement security.protect.start

stgraber:


internal/server/instance/drivers/driver_common.go
```diff
@@ -965,6 +965,11 @@ func (d *common) validateStartup(stateful bool, statusCode api.StatusCode) error
 		return errors.New("Requested architecture isn't supported by this host")
 	}
 
+	// Check if instance is start protected.
+	if util.IsTrue(d.expandedConfig["security.protection.start"]) {
+		return errors.New("Instance is protected")
```
bensmrs: I think the error message is quite confusing to the user.

bensmrs (changes requested):

internal/server/instance/drivers/driver_common.go
stgraber: Pushed an extra commit to clarify them all as we had a similar error on delete.

bensmrs (approved):

PR #3769, opened 2026-08-04
Fix resolv.conf generation in forknet dhcp

stgraber:
Closes #3764


jochumdev:
This looks like a very clean fix to me.

Locking + per value exclusion. Nice!

presztak (approved):

PR #3784, opened 2026-08-06
client: Event connection liveness and Disconnect fixes

jochumdev:
Two independent fixes to the client-side event connection lifecycle.
They can be reviewed and dropped separately.

### DIFF
--- a/client/incus_events.go
+++ b/client/incus_events.go
@@ -113,11 +113,16 @@ func (r *ProtocolIncus) getEvents(allProjects bool, eventTypes []string) (*Event
 
 			r.eventListenersLock.Lock()
 			r.eventConnsLock.Lock()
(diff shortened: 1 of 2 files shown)


stgraber:
I'll take care of fixing and landing this.

client/incus_events.go
```diff
@@ -119,11 +151,37 @@ func (r *ProtocolIncus) getEvents(allProjects bool, eventTypes []string) (*Event
 		}
 	}()
 
+	// Spawn a pinger so an idle but healthy connection doesn't hit the read deadline.
```
stgraber: Most of this logic is unnecessary. We already have a pinger on the server side sending a keepalive every 10s. So all you need is for the client to close on inactivity from the server.

stgraber (approved):

PR #3785, opened 2026-08-06
Fix event ordering

jochumdev:
Every event was written to each listener in its own go routine, so events
raced each other and arrived in whatever order the scheduler picked.
Anything reconstructing state from the event stream ends up wrong — the
symptom that found this was `project-updated` arriving before
`project-created`.

The Go client did the same thing when calling handlers, so ordering there
is opt-in via `SetOrdered()` rather than changing behavior under existing
consumers.

Also fixes `project-updated` being sent before `projectChange` applied the
update, which meant it fired even when the update was rejected.

### DIFF
--- a/cmd/incusd/api_project.go
+++ b/cmd/incusd/api_project.go
@@ -624,10 +624,15 @@ func projectPut(d *Daemon, r *http.Request) response.Response {
 		return response.BadRequest(err)
 	}
 
+	err = projectChange(r.Context(), s, project, req)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
 	requestor := request.CreateRequestor(r)
 	s.Events.SendLifecycle(project.Name, lifecycle.ProjectUpdated.Event(project.Name, requestor, nil))
 
-	return projectChange(r.Context(), s, project, req)
+	return response.EmptySyncResponse
 }
 
 // swagger:operation PATCH /1.0/projects/{name} projects project_patch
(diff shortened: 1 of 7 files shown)


cmd/incusd/api_project.go
```diff
@@ -624,10 +624,7 @@ func projectPut(d *Daemon, r *http.Request) response.Response {
 		return response.BadRequest(err)
 	}
 
-	requestor := request.CreateRequestor(r)
-	s.Events.SendLifecycle(project.Name, lifecycle.ProjectUpdated.Event(project.Name, requestor, nil))
-
-	return projectChange(r.Context(), s, project, req)
+	return projectChange(r.Context(), s, project, req, request.CreateRequestor(r))
```
stgraber: Instead of passing through the requestor to projectChange, just handle the error from projectChange in the caller and doe the SendLifecycle call after that.

cmd/incusd/api_project.go
```diff
@@ -749,14 +746,11 @@ func projectPatch(d *Daemon, r *http.Request) response.Response {
 		}
 	}
 
-	requestor := request.CreateRequestor(r)
-	s.Events.SendLifecycle(project.Name, lifecycle.ProjectUpdated.Event(project.Name, requestor, nil))
-
-	return projectChange(r.Context(), s, project, req)
+	return projectChange(r.Context(), s, project, req, request.CreateRequestor(r))
```
stgraber: same

stgraber:
I'll take care of fixing and landing this.

PR #3790, opened 2026-08-07
Expose image location in cluster

### LINKED ISSUE #3788: Expose location for images

Most of the Incus resources, that are tied to a single server within an Incus cluster expose the location information when the resource is queried over the API. For images, this is currently not the case, which prevents Operations Center from collecting this information. Beside of consistency and completeness requirements, this particular information would also be helpful to decide, when it is save for server to be removed from an existing cluster. The removal of a server from a cluster should be prevented within Operations Center, if there are images, which are only available on this particular server.

The location information is already present in the DB. The only things, that are needed to be add are:

* Add it to the API (most likely as a list of locations, since images might be available on multiple servers)
* Ensure, that this information is available in the client functions `GetImages` and `GetImagesAllProjects`

Related issue: https://github.com/FuturFusion/operations-center/issues/813

stgraber:
Closes #3788

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -3366,3 +3366,8 @@ running container is migrated this way.
 It also adds `refresh-migrate` as a configuration option to `cluster.evacuate`
 and as an evacuation mode, which forces refresh migration of instances during
 cluster evacuation.
+
+## `image_locations`
+
+Adds a new `locations` field to the image API, listing the names of the
+cluster members which have a local copy of the image.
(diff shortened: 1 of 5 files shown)


breml (approved):
LGTM

bensmrs (approved):
Never a huge fan of the term “location” being pluralized, but that’s a matter of taste I guess.

stgraber:
> Never a huge fan of the term “location” being pluralized, but that’s a matter of taste I guess.

Yeah, in this case I was just following precedent as we have it already for storage pools and networks

PR #3794, opened 2026-08-07
Add support for dependent disks during in-cluster near-live migration

### LINKED ISSUE #3059: Near-live migration for containers

presztak:
Closes: #3059

### DIFF
--- a/cmd/incusd/instance_post_nearlive.go
+++ b/cmd/incusd/instance_post_nearlive.go
@@ -15,6 +15,7 @@ import (
 	"github.com/lxc/incus/v7/internal/server/instance"
 	"github.com/lxc/incus/v7/internal/server/instance/instancetype"
 	"github.com/lxc/incus/v7/internal/server/operations"
+	"github.com/lxc/incus/v7/internal/server/response"
 	"github.com/lxc/incus/v7/internal/server/state"
 	storagePools "github.com/lxc/incus/v7/internal/server/storage"
 	"github.com/lxc/incus/v7/internal/version"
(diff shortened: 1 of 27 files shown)


stgraber:
@presztak I really don't like the added shared/api options. We shouldn't have anything internal be exposed to the public API in that way.

Is there no way for us to detect what we need internally considering all of this is happening with a shared database and with the client user-agent and certificate set to the cluster internal one so we can detect that this isn't a user request but is in fact a cluster-internal operation.

presztak:
@stgraber Another option I thought, would be to detect that this is a `copy` request originating from a different node, I mean, an inter-cluster request. However, reusing `copy` request elsewhere would then cause us to always skip shared volumes.

presztak:
quick update, on the server side, we're now using `migration` directly instead of `copy`. This avoids exposing new external fields related to dependent disk migration, and has reduced the number of commits to 4. I've moved the PR back to draft, since we're still missing support for near-live migration when the remote is on shared storage but dependent disks are local.

cmd/incusd/instance_post_nearlive.go
```diff
@@ -29,20 +33,23 @@ func nearLiveMigrationSupported(s *state.State, inst instance.Instance, req api.
...
-	dependent := false
 	err := inst.ForEachDependentDiskType(func(dev deviceConfig.DeviceNamed) error {
-		dependent = true
+		diskPool, err := storagePools.LoadByName(s, dev.Config["pool"])
+		if err != nil {
+			return fmt.Errorf("Failed loading storage pool: %w", err)
+		}
+
+		driverName := diskPool.Driver().Info().Name
+		if driverName == "dir" || driverName == "lvm" {
```
stgraber: Can we please make that a DriverInfo property so we don't need to hardcode drivers name.

test/suites/clustering_move.sh
```diff
@@ -195,8 +195,31 @@ EOF
         INCUS_DIR="${INCUS_ONE_DIR}" incus info c4 | grep -q "Location: node2"
         INCUS_DIR="${INCUS_ONE_DIR}" incus info c4 | grep -q "Status: STOPPED"
 
+        # Dependent volumes are moved along with the instance.
```
stgraber, line 289: Can we add a test running on ceph and linstor which validates the tricky edge cases?

Specifically, I'd like to see two more tests:
 - Instance on shared storage, one dependent volume on local, one dependent volume on shared, one regular volume on shared
 - Instance on local storage, one dependent volume on local, one dependent volume on shared, one regular volume on shared

I'd also prefer that the test writes a basic test file to the instance and to each of the dependent and regular volumes just prior to the move and confirms that the files exist with the same content following the move.

cmd/incusd/instance_post_nearlive.go
presztak: Make sense. Done.

stgraber:
@presztak what's with that "test" commit? :)

presztak:
@stgraber it's just for testing purpose for now as I've hit issues on linstor driver.

test/suites/clustering_move.sh
presztak: Tests added. One test for linstor is failing so we need to wait for: https://github.com/lxc/incus/pull/3866

internal/server/storage/drivers/driver_cephfs.go
```diff
@@ -94,6 +94,7 @@ func (d *cephfs) Info() Info {
 		OptimizedImages:              false,
 		PreservesInodes:              false,
 		Remote:                       d.isRemote(),
+		NearLiveMigration:            true,
```
stgraber: instances can't be on cephfs, so that should be a no

internal/server/storage/drivers/driver_truenas.go
```diff
@@ -105,6 +105,7 @@ func (d *truenas) Info() Info {
 		OptimizedBackups:             false,
 		PreservesInodes:              false,
 		Remote:                       d.isRemote(),
+		NearLiveMigration:            true,
```
stgraber: Should be false as we don't have a good server to server incremental mechanism.

internal/server/storage/drivers/driver_lvm.go
```diff
@@ -156,6 +156,7 @@ func (d *lvm) Info() Info {
 		OptimizedImages:              d.usesThinpool(), // Only thinpool pools support optimized images.
 		PreservesInodes:              false,
 		Remote:                       d.isRemote(),
+		NearLiveMigration:            d.clustered,
```
stgraber: Should be false as we don't have a good server to server incremental mechanism.

internal/server/storage/drivers/driver_linstor.go
```diff
@@ -294,6 +294,7 @@ func (d *linstor) Info() Info {
 		DefaultVMBlockFilesystemSize: deviceConfig.DefaultVMBlockFilesystemSize,
 		Buckets:                      false,
 		Remote:                       d.isRemote(),
+		NearLiveMigration:            true,
```
stgraber: Should be false as we don't have a good server to server incremental mechanism.

internal/server/storage/drivers/driver_ceph.go
```diff
@@ -93,6 +93,7 @@ func (d *ceph) Info() Info {
 		OptimizedImages:              true,
 		PreservesInodes:              false,
 		Remote:                       d.isRemote(),
+		NearLiveMigration:            true,
```
stgraber: Should be false as we don't have a good server to server incremental mechanism.

stgraber:
@presztak the Info flag seems wrong to me. It's meant to indicate whether an efficient transfer of incremental state is available or not (the requirement for near live migration).

Only ZFS and btrfs offer that. The rest all rely on rsync/dd which aren't suitable for this use case.

Remote pools within a cluster also don't need the flag set to true anyway since the data isn't going to be moving at all. But if you were to go cross-cluster between remote storage drivers on both ends, none of our drivers have an efficient way to transfer those snapshots so they shouldn't be treated as handling near live migration.

presztak:
@stgraber, ok, I did it that way because the initial condition was:
```
if poolInfo.Name == "dir" || poolInfo.Name == "lvm" {
    return fmt.Errorf("Near-live migration isn't supported for dependent disk %q on %q storage pools", dev.Name, poolInfo.Name)
}
```
So `NearLiveMigration` was `true` for everything except `dir` and `lvm`. But I see your point now, and I've made adjustments based on your feedback.

internal/server/storage/drivers/driver_cephfs.go
presztak: Fixed.

internal/server/storage/drivers/driver_truenas.go
presztak: Fixed.

internal/server/storage/drivers/driver_lvm.go
presztak: Fixed.

internal/server/storage/drivers/driver_linstor.go
presztak: Fixed.

internal/server/storage/drivers/driver_ceph.go
presztak: Fixed.

PR #3797, opened 2026-08-07
client: Replace SetOrdered with AddChannel

jochumdev:
`AddChannel` lets a consumer receive events in order without a goroutine
per event, and since it hands back a channel it can be consumed with a
plain `for range` loop — no extra channel and forwarding goroutine on the
consumer side.

`AddHandler` is untouched and incusd keeps using it for operations.

A reader that falls too far behind is dropped rather than having events
silently skipped, so a stale view can't go unnoticed.


doc/events.md
```diff
@@ -98,8 +101,8 @@ type: lifecycle
...
 | `image-alias-updated`                  | The configuration for an image alias has changed.                     | `target`: the original instance.                                                                     |
-| `instance-agent-started`               | The instance agent has connected to the host.                        |                                                                                                      |
-| `instance-agent-stopped`               | The instance agent has disconnected from the host.                   |                                                                                                      |
+| `instance-agent-started`               | The instance agent has connected to the host.                         |                                                                                                      |
+| `instance-agent-stopped`               | The instance agent has disconnected from the host.                    |                                                                                                      |
```
stgraber, line 103: Looks unrelated, should probably be its own commit.

jochumdev: Sorry, my damn auto format on save.

doc/events.md
```diff
@@ -22,8 +22,8 @@ reconnecting, anything being tracked from the event stream should be re-fetched,
 disconnected are not resent.
```
stgraber, line 22: Why is there a doc change in this commit?

stgraber: That refers to [client: Add EventListener.AddChannel](https://github.com/lxc/incus/pull/3797/commits/e757cb64997a1bf237db4c10bfadec34e7f11c61)

jochumdev:
I have written a fork of `incus/v7/client` means I have no need for this, if incus has no use for this drop it.

Issue #3804, opened 2026-08-09
Allow multiple --envionment-file arguments


AJRepo:
### Is there an existing issue for this?

- [x] There is no existing issue for this feature

### What are you currently unable to do

Setting up different containers with different environmental variables and sometimes there is overlap between containers where container1 might have some similar and some different variables to container2. 

A simple example would be container1 has user1 and user2 while container2 has user2 and user3. So one would run: 

```
incus launch images:XXX container1 --environment-file user1.env --environment-file user2.env 
incus launch images:XXX container2 --environment-file user2.env --environment-file user3.env 
```

### What do you think would need to be added

Modify `create.go` to change the variable from a string to an array and update the parsing logic. 

I have done this and will submit a PR momentarily.

PR #3805, opened 2026-08-09
Add multiple --environment-file args

AJRepo:
Allow the flag --environement-file to be called multiple times with different files as in:

incus launch images:debian/13 env-test \
--environment-file test1.env \
--environment-file test2.env

See Feature Request #3804 

Signed-off-by: AJRepo <AJRepo@users.noreply.github.com>

### DIFF
--- a/cmd/incus/create.go
+++ b/cmd/incus/create.go
@@ -28,7 +28,7 @@ type cmdCreate struct {
 
 	flagConfig          []string
 	flagDevice          []string
-	flagEnvironmentFile string
+	flagEnvironmentFile []string
 	flagEphemeral       bool
 	flagNetwork         string
 	flagProfile         []string
@@ -63,7 +63,7 @@ incus launch images:debian/12 v2 --vm -d root,size=50GiB -d root,io.bus=nvme
 	cli.AddStringArrayFlag(cmd.Flags(), &c.flagProfile, "profile|p", i18n.G("Profile to apply to the new instance"))
 	cli.AddStringArrayFlag(cmd.Flags(), &c.flagDevice, "device|d", i18n.G("New key/value to apply to a specific device"))
 	cli.AddBoolFlag(cmd.Flags(), &c.flagEphemeral, "ephemeral|e", i18n.G("Ephemeral instance"))
-	cli.AddStringFlag(cmd.Flags(), &c.flagEnvironmentFile, "environment-file", "", "", i18n.G("Include environment variables from file"))
+	cli.AddStringArrayFlag(cmd.Flags(), &c.flagEnvironmentFile, "environment-file", i18n.G("Include environment variables from file"))
 	cli.AddStringFlag(cmd.Flags(), &c.flagNetwork, "network|n", "", "", i18n.G("Network name"))
 	cli.AddStringFlag(cmd.Flags(), &c.flagStorage, "storage|s", "", "", i18n.G("Storage pool name"))
 	cli.AddStringFlag(cmd.Flags(), &c.flagType, "type|t", "", "", i18n.G("Instance type"))
@@ -188,8 +188,12 @@ func (c *cmdCreate) create(conf *config.Config, parsed []*u.Parsed, launch bool)
 		configMap = map[string]string{}
 	}
 
-	if c.flagEnvironmentFile != "" {
-		envMap, err := readEnvironmentFile(c.flagEnvironmentFile)
+	for _, envFile := range c.flagEnvironmentFile {
+		if envFile == "" {
+			continue
+		}
+
+		envMap, err := readEnvironmentFile(envFile)
 		if err != nil {
 			return nil, err
 		}
--- a/cmd/incus/create.go
+++ b/cmd/incus/create.go
@@ -59,11 +59,11 @@ incus launch images:debian/12 v2 --vm -d root,size=50GiB -d root,io.bus=nvme
 
 	cmd.Aliases = []string{"init"}
 	cmd.RunE = c.run
-	cli.AddStringArrayFlag(cmd.Flags(), &c.flagConfig, "config|c", i18n.G("Config key/value to apply to the new instance"))
-	cli.AddStringArrayFlag(cmd.Flags(), &c.flagProfile, "profile|p", i18n.G("Profile to apply to the new instance"))
-	cli.AddStringArrayFlag(cmd.Flags(), &c.flagDevice, "device|d", i18n.G("New key/value to apply to a specific device"))
+	cli.AddStringArrayFlag(cmd.Flags(), &c.flagConfig, "config|c", i18n.G("Config key/value to apply to the new instance (may be passed multiple times)"))
+	cli.AddStringArrayFlag(cmd.Flags(), &c.flagProfile, "profile|p", i18n.G("Profile to apply to the new instance (may be passed multiple times)"))
+	cli.AddStringArrayFlag(cmd.Flags(), &c.flagDevice, "device|d", i18n.G("New key/value to apply to a specific device (may be passed multiple times)"))
 	cli.AddBoolFlag(cmd.Flags(), &c.flagEphemeral, "ephemeral|e", i18n.G("Ephemeral instance"))
-	cli.AddStringArrayFlag(cmd.Flags(), &c.flagEnvironmentFile, "environment-file", i18n.G("Include environment variables from file"))
+	cli.AddStringArrayFlag(cmd.Flags(), &c.flagEnvironmentFile, "environment-file", i18n.G("Include environment variables from file (may be passed multiple times)"))
 	cli.AddStringFlag(cmd.Flags(), &c.flagNetwork, "network|n", "", "", i18n.G("Network name"))
 	cli.AddStringFlag(cmd.Flags(), &c.flagStorage, "storage|s", "", "", i18n.G("Storage pool name"))
 	cli.AddStringFlag(cmd.Flags(), &c.flagType, "type|t", "", "", i18n.G("Instance type"))
--- a/cmd/incus/copy.go
+++ b/cmd/incus/copy.go
@@ -57,9 +57,9 @@ The pull transfer mode is the default as it is compatible with all server versio
 	))
 
 	cmd.RunE = c.run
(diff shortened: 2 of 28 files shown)


AJRepo:
Tested and works. Here is the testing code: 

```bash
$cat <<EOF > test1.env
FIRST_VAR=apple
SHARED_VAR=original_value
EOF

$cat <<EOF > test2.env
SECOND_VAR=banana
SHARED_VAR=overridden_value
EOF

$~/go/bin/incus launch images:debian/13 env-test \
  --environment-file test1.env \
  --environment-file test2.env
Launching env-test
Retrieving image: Unpacking image

$~/go/bin/incus config show env-test | grep environment
environment.FIRST_VAR: apple
environment.SECOND_VAR: banana
environment.SHARED_VAR: overridden_value

```

cmd/incus/create.go
```diff
@@ -63,7 +63,7 @@ incus launch images:debian/12 v2 --vm -d root,size=50GiB -d root,io.bus=nvme
 	cli.AddStringArrayFlag(cmd.Flags(), &c.flagProfile, "profile|p", i18n.G("Profile to apply to the new instance"))
 	cli.AddStringArrayFlag(cmd.Flags(), &c.flagDevice, "device|d", i18n.G("New key/value to apply to a specific device"))
 	cli.AddBoolFlag(cmd.Flags(), &c.flagEphemeral, "ephemeral|e", i18n.G("Ephemeral instance"))
-	cli.AddStringFlag(cmd.Flags(), &c.flagEnvironmentFile, "environment-file", "", "", i18n.G("Include environment variables from file"))
+	cli.AddStringArrayFlag(cmd.Flags(), &c.flagEnvironmentFile, "environment-file", i18n.G("Include environment variables from file"))
```
bensmrs: Maybe give a hint in the flag description that you can use it several times.

cmd/incus/create.go
```diff
@@ -188,14 +188,19 @@ func (c *cmdCreate) create(conf *config.Config, parsed []*u.Parsed, launch bool)
...
-		envMap, err := readEnvironmentFile(c.flagEnvironmentFile)
-		if err != nil {
-			return nil, err
-		}
-
-		for k, v := range envMap {
-			configMap["environment."+k] = v
+	if len(c.flagEnvironmentFile) > 0 {
+		for _, envFile := range c.flagEnvironmentFile {
+			if envFile == "" {
```
bensmrs: I don’t think we should add this test; because the files are now a list, there shouldn’t be an empty string.

cmd/incus/create.go
```diff
@@ -188,14 +188,19 @@ func (c *cmdCreate) create(conf *config.Config, parsed []*u.Parsed, launch bool)
...
+				continue
+			}
+			envMap, err := readEnvironmentFile(envFile)
+			if err != nil {
+				return nil, err
+			}
```
bensmrs: Please have a look at the linting errors. You should leave an empty line after each block.

cmd/incus/create.go
```diff
@@ -188,14 +188,19 @@ func (c *cmdCreate) create(conf *config.Config, parsed []*u.Parsed, launch bool)
...
+			// Merge the keys into our master map
```
bensmrs: I’d remove that comment; it’s pretty trivial to understand what your code does ;)

bensmrs (changes requested):

cmd/incus/create.go
AJRepo: If this is something you feel should be changed, I am happy to do so, however, I think that change should be consistent across all of the ArrayFlags (flagConfig, flagProfile,FlagDevice,flagEnvironmentFile) 

Perhaps prepending "Multiple: " to each of their descriptions?

cmd/incus/create.go
AJRepo: Will do. 

I don't see those linting errors though. What linter and .golangci.yml are you using? I tried with both go-staticcheck and golangci-lint using the .golangci.yml in the root of this codebase and neither gave me linting errors about these lines. I added "- wsl_v5" under  enable, but still got no errors related to whitespace.

cmd/incus/create.go
bensmrs: Eh… WDYT @stgraber?

cmd/incus/create.go
bensmrs: `make static-analysis`, or look at what the red `Tests / Code` checks say in the CI.
Also, please sign your commits to solve the `Commits / Signed-off-by (DCO)` check.

AJRepo: There seems to be a problem with the DCO CI check as I did sign the commits as usual. 

---

when I run `make static-analysis` I don't get notified about those block issues.

```
$ make static-analysis 
[deleted lines]
Checking for golangci-lint errors...
cmd/incus/action.go:1: : # github.com/lxc/incus/v7/cmd/incus [github.com/lxc/incus/v7/cmd/incus.test]
cmd/incus/export_test.go:23:13: exp.Command undefined (type *cmdExport has no field or method Command, but does have method command) (typecheck)
package main
1 issues:
* typecheck: 1
run-parts: test/lint/golangci.sh exited with return code 1
make: *** [Makefile:296: static-analysis] Error 1
```

Did you want that fixed in this PR? If yes, as exp.command or exp.Command?  If no - let me know if you want me to refactor off of that new point or want a separate PR, etc.

cmd/incus/create.go
AJRepo: I asked another coder who suggested "(can be used multiple times)" and that is now in the updated PR with all your other requests. Thanks!

cmd/incus/create.go
bensmrs: > There seems to be a problem with the DCO CI check as I did sign the commits as usual.

My guess is that you’re supposed to sign at the last line of the commit message.

> when I run make static-analysis I don't get notified about those block issues

Because you fixed them; this is the command that the CI executes.

> cmd/incus/export_test.go:23:13: exp.Command undefined (type *cmdExport has no field or method Command, but does have method command) (typecheck)

This file doesn’t exist in our codebase.

cmd/incus/create.go
bensmrs: I’d say to keep the initial description as-is, and do a single commit to fix all the multiple array descriptions at once; it’ll be easier to track this way.

bensmrs: Please squash your commits, do an additional commit as I described, then update the translation templates with `make i18n`, then `git commit -sm "i18n: Update translation templates" po/`.

cmd/incus/create.go
AJRepo: > This file doesn’t exist in our codebase.

my bad. I was experimenting with adding a test and forgot to remove it. 

> My guess is that you’re supposed to sign at the last line of the commit message.

I also see github changed how anonymous emails work. I'll try that too.

cmd/incus/create.go
AJRepo: will do.

cmd/incus/create.go
AJRepo: DCO issue resolved. 

I still don't get linting notifications with whitespace/newline after if blocks using `make static-analysis` even if I remove those newlines. But I'll mark this as resolved since they are in there anyway.

cmd/incus/create.go
AJRepo: I think I have done as you suggested: The last force-push now has 3 commits:

1. Changed to array
2. Changed text description
3. i18n update

---

I still don't understand why running `make static-analysis` gives a different result when I run it locally vs the git push to github. 

I'm running the same version of golangci-lint
Github CI logs say: 
`golangci/golangci-lint info found version: 2.12.2 for v2.12.2/linux/amd64`

Locally: 
```
$ golangci-lint --version
golangci-lint has version 2.12.2 built with go1.26.2 from c0d3ddc9 on 2026-05-06T11:07:58Z
```

In theory both should be using the same `.golangci.yml` ? 

Running `make static-analysis` locally  doesn't barf at newlines after `if blocks` (like it does on github). 

Ont the other hand, my local running of it gives me errors like 

```
Checking for golangci-lint errors...
../incuslxc.rebase/test/mini-oidc/storage/client.go:11:1: Comment should end in a period (godot)
        // we use the default login UI and pass the (auth request) id
^
../incuslxc.rebase/test/mini-oidc/storage/client.go:16:1: Comment should end in a period (godot)
        // clients to be used by the storage interface
^
../incuslxc.rebase/test/mini-oidc/storage/client.go:21:1: Comment should end in a period (godot)
// this could also be your database model
^
```

But it passes in github: https://github.com/lxc/incus/actions/runs/31409360991/job/93523493293?pr=3805#step:15:6983 

---

Anyway - it's all green in the CI checks so closing this bit to start the review.

AJRepo: I figured out why I didn't see it locally.  

I'm documenting here for others who might run into similar issues. 

.golangci.yml excludes `path: "^test/mini-oidc/storage"`  i.e.

```yaml
 ...
exclusions:
    generated: lax
    rules:
      - linters:
          - revive
        source: '^//generate-database:mapper '
      - linters:
          - revive
        text: "avoid package names that conflict with Go standard library package names"
        path: "^internal/io/"
      - linters:
          - revive
          - godot
        path: "^test/mini-oidc/storage/"
...
```

Note that "^test" regexp is for at the beginning of the path. 

I had done a `cp -r old_dir new_dir` of the repo dir as a working directory and was working in a soft link directory. 

These issues compounded: 
1.  The previous make kept a full path of the old_dir cached.  
2.  `pwd -P` was different than `pwd`
3.  my browser denied scripts from productionresultssa*.blob.core.windows.net; (which displays some of the error reports on github)

1 and 2 above meant  `make static-analysis` had in the path a traversal that  went upwards first and then back down . Thus the directories started with  `../${pwd  -P}`  which meant `../directory/test` wouldn't be excluded, and thus it was linting  in the test directory and stopping with errors there before getting to the code I was editing. 

To fix: I ran

```bash
$ cd "$(pwd -P)"
$ golangci-lint cache clean
$ go clean -cache -modcache
$ make static-analysis
```

Allowing browser scripts from the productionresults...blob.core.windows.net allowed all the errors on github CI to be shown. 

With the above changes I got all the same results locally as seen on github.

PR #3806, opened 2026-08-10
Fix for various recent test failures

stgraber:


stgraber:
As part of my work on expanding the testsuite, I've done some digging on recent test failures (since Incus 7.3) and those were the easy-enough ones to fix, hopefully we get a more reliable testsuite out of it.

internal/server/storage/drivers/driver_linstor_utils.go
```diff
@@ -861,18 +874,27 @@ func (d *linstor) deleteResourceDefinitionFromSnapshot(vol Volume) error {
 		return err
 	}
 
-	resourceDefinition, err := d.getResourceDefinition(vol, false)
+	// Delete every matching resource definition, as an interrupted past deletion can leave more than one behind.
```
bensmrs, line 877: With your reuse logic, even if a resource definition is left behind, it should be reused and there should thus only be one, no?

internal/server/storage/drivers/driver_linstor_volumes.go
```diff
@@ -1037,13 +1037,21 @@ func (d *linstor) MountVolumeSnapshot(snapVol Volume, op *operations.Operation)
 		return d.MountVolumeSnapshot(fsVol, op)
 	}
 
-	// Create a new temporary resource definition from the snapshot
-	err = d.createResourceDefinitionFromSnapshot(snapVol, snapVol)
+	// Reuse any existing resource definition, e.g. from a concurrent mount or an interrupted cleanup.
```
bensmrs, line 1040: I’m not sure how concurrent mounts would behave; are we guaranteed that it’s read-only? I’m not seeing `ro` as a mount option but it may be hidden somewhere…

cmd/incusd/api_internal.go
bensmrs, line 1: I’m curious if that solves a problem or if it’s just additional safety.

cmd/incusd/api_cluster.go
```diff
@@ -2705,12 +2705,18 @@ findLeader:
 		return fmt.Errorf("Failed handing over cluster member role: %w", err)
 	}
 
-	_, _, err = client.RawQuery("POST", "/internal/cluster/handover", post, "")
-	if err != nil {
-		return err
+	// Retry for a while if another configuration change is in progress, as
+	// leaving without a successful handover can cost the cluster its quorum.
+	for range 30 {
```
bensmrs, line 2710: Well if the impact is potentially that critical, is there any reason for not choosing anything bigger?

cmd/incusd/api_internal.go
stgraber: This commit/change fixes a few cases where we've seen a `incus admin sql global` call in a testsuite run fail due to the database being locked (transaction currently being open).

We haven't seen it too much but it has shown up over the past month.

internal/server/storage/drivers/driver_linstor_utils.go
stgraber: It would probably be safe to skip this but it has the side effect of cleaning up anything that's already been left over and it also handles some potential weirdness if unattached custom volumes.

Basically for attached custom volumes or for instances, we route the requests through the system that own the resource, that system then has an in-memory lock to prevent concurrent operations. But when dealing with an unattached custom volume, whichever server receives the request handles it directly, leading to the potential for things piling up a bit if the same operation hits multiple servers.

That's not really likely/possible through normal user interactions, but we've seen more of this kind of stuff happen through tools like Terraform or incus-compose.

internal/server/storage/drivers/driver_linstor_volumes.go
stgraber: Hmm, I believe we're working on a clone of the resource though, so whether read-only or read-write shouldn't really matter?

cmd/incusd/api_cluster.go
stgraber: Normally a raft reconfig stabilizes in less than a second, so giving ourselves 30s here before moving on should be fine.

stgraber: The rare cases where it wouldn't is if the cluster is really FUBAR, such as may happen if the network is super lossy or the other members have a full disk or something, effectively preventing a proper reconfig/election from occurring.

internal/server/storage/drivers/driver_linstor_volumes.go
bensmrs: Exactly, but then if you’re allowing reuse, someone could override something, which then potentially gets used by the other consumer. And that’s ignoring ext4 corruption. I may be missing something in your logic though.

bensmrs: Oh, and LINSTOR doesn’t allow mounting the same volume at different places (except when migrating; there may be something contradicting what I’m saying, but that’s being looked at in the LINSTOR issue)

bensmrs: I forgot about it in my initial comment, and we may have designed the function this way because of this limitation. Although, if possible, we should relax that for mounted snapshots, if we can mount them RO.

stgraber: Hmm, yeah, I'm still not very familiar with Linstor, mostly poking at it when its test failures start getting on my nerves ;)

Checking again, it indeed seems like LINSTOR would prevent the concurrent cross-node use in the first place. Locally we're dealing with a clone and none of our code paths should ever write to that (SnapshotMount is meant for read-only access).

But then we have a clear read-only mount flag on other drivers, so makes sense to add it here too. Will push an extra commit for that.

stgraber: Pushed the extra commit now

bensmrs (approved):

internal/server/storage/drivers/driver_linstor_volumes.go
bensmrs: I think I’ll have to do another pass on the logic (the original one, not your PR) later. But that’s one of those feature codebases where I always go back reluctantly, because the original work was incredibly tiresome.

PR #3812, opened 2026-08-11
Various bugfixes

stgraber:


cmd/incusd/api_cluster.go
```diff
@@ -2678,42 +2678,63 @@ func handoverMemberRole(s *state.State, gateway *cluster.Gateway) error {
...
 		_, _, err = client.RawQuery("POST", "/internal/cluster/handover", post, "")
-		if err == nil || !strings.Contains(err.Error(), "configuration change is already in progress") {
-			return err
+		if err == nil {
+			return nil
 		}
 
-		time.Sleep(time.Second)
+		// Give up right away on errors that won't resolve by retrying.
+		if !strings.Contains(err.Error(), "configuration change is already in progress") && !strings.Contains(err.Error(), "not leader") && !strings.Contains(err.Error(), "Unable to connect") {
```
bensmrs: `ErrNotLeader` is `Not leader` (first letter capitalized), so you’re only matching the `503 not leader` error. Resolve the comment if that’s what you want.

bensmrs: You could also use the actual constants, like `errClusterBusy`.

stgraber: Yeah, there's also a logic issue which is causing the test failures.

I'm going to be exporting the error so it's easier to compare and tweak the handling a bit... We're going to end up with an even longer if ;)

cmd/incusd/api_cluster.go
```diff
@@ -2678,42 +2678,79 @@ func handoverMemberRole(s *state.State, gateway *cluster.Gateway) error {
...
 		_, _, err = client.RawQuery("POST", "/internal/cluster/handover", post, "")
-		if err == nil || !strings.Contains(err.Error(), "configuration change is already in progress") {
-			return err
+		if err == nil {
+			return nil
 		}
 
-		time.Sleep(time.Second)
+		// Give up right away on errors that won't resolve by retrying.
+		if !strings.Contains(err.Error(), cluster.ErrClusterBusy.Error()) && !strings.Contains(err.Error(), cluster.ErrNotLeader.Error()) && !strings.Contains(err.Error(), "503 not leader") && !strings.Contains(err.Error(), "Unable to connect") && !strings.Contains(err.Error(), "Client.Timeout") {
```
bensmrs: I’d go for exact matches when it comes to exported error constants.

internal/server/cluster/membership.go
```diff
@@ -448,7 +448,7 @@ func Join(s *state.State, gateway *Gateway, networkCert *localtls.CertInfo, serv
 			return fmt.Errorf("Failed to join cluster: %w", ctx.Err())
 		default:
 			err = cowsqlClient.Add(ctx, info.NodeInfo)
-			if err != nil && err.Error() == errClusterBusy.Error() {
+			if err != nil && err.Error() == ErrClusterBusy.Error() {
```
bensmrs, line 451: Any reason not to use one of the comparison functions from `errors`?

cmd/incusd/api_cluster.go
stgraber: Can't do that because we're getting them over the network and they're typically wrapped so we can't use things like errors.Is or errors.As, so at best we could do a HasSuffix but that could also be brittle. Because the errors are thankfully pretty distinct, the Contains trick works well enough.

internal/server/cluster/membership.go
stgraber: go-cowsql returns errors as strings so errors.As/errors.Is won't work, only comparing their value works in this scenario.

stgraber:
@bensmrs responded above, but basically we're dealing with strings because some of those errors (e.g. ErrClusterBusy) aren't "our" error, it's actually coming from libraft through cowsql through go-cowsql. We only have the error defined for comparison purposes, we don't return it ourselves and therefore can't directly compare it or use errors.Is/errors.As.

Similar story with some of the others where while we do own them, they're coming from a /internal/XYZ API call and so are received over the network as a concatenated string rather than a native Go error stack from within the same daemon. This again prevents us from using direct comparison or errors.Is/errors.As.

bensmrs:
Alright then, sorry for being pushy but that really felt dirty. Not much we can do about it I guess.

bensmrs (approved):

stgraber:
@bensmrs yeah, this code isn't exactly ideal... @masnax is working on a bit of refactoring around a bunch of this code, moving it directly into go-cowsql so we can re-use it in Operations Center and elsewhere. This may make some of this a bit cleaner.

PR #3813, opened 2026-08-11
Update swagger specification

### LINKED ISSUE #3793: Improve Swagger 2.0 compatibility of generated API specification (doc/rest-api.yaml)
Triggered by https://github.com/FuturFusion/operations-center/issues/988, an effort as been made in https://github.com/FuturFusion/operations-center/pull/994 to improve the Swagger 2.0 compatibility of the generated Swagger 2.0 API specification in Operations Center.

The same fixes can be applied to the Swagger specification of Incus as well. The state of the validity can be verified easily with `swagger validate doc/rest-api.yaml` (with `go install github.com/go-swagger/go-swagger/cmd/swagger@latest`) and in then kept consistent by adding `swagger validate` to the linting scripts (see: https://github.com/FuturFusion/operations-center/blob/main/scripts/lint/api-doc-up-to-date.sh).

In the current state, `swagger validate` shows ~2000 errors and  ~60 warnings.

stgraber:
Fixes swagger warnings/errors, similar to what was by @breml on Operations Center.

### DIFF
--- a/cmd/incusd/images.go
+++ b/cmd/incusd/images.go
@@ -1679,11 +1679,9 @@ func doImagesGet(ctx context.Context, tx *db.ClusterTx, recursion bool, projectN
 //            description: List of endpoints
 //            items:
 //              type: string
-//            example: |-
-//              [
-//                "/1.0/images/06b86454720d36b20f94e31c6812e05ec51c1b568cf3a8abd273769d213394bb",
-//                "/1.0/images/084dd79dd1360fd25a2479eb46674c2a5ef3022a40fe03c91ab3603e3402b8e1"
-//              ]
+//            example:
+//              - /1.0/images/06b86454720d36b20f94e31c6812e05ec51c1b568cf3a8abd273769d213394bb
+//              - /1.0/images/084dd79dd1360fd25a2479eb46674c2a5ef3022a40fe03c91ab3603e3402b8e1
 //    "400":
 //      $ref: "#/responses/BadRequest"
 //    "403":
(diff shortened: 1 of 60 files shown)


breml (approved):
LGTM

cmd/incusd/images.go
```diff
@@ -1121,45 +1121,36 @@ func imageCreateInPool(s *state.State, info *api.Image, storagePool string) erro
...
+//	    items:
+//	      type: string
 //	  - in: header
 //	    name: X-Incus-properties
-//	    description: Descriptive properties
-//	    schema:
-//	      type: object
-//	      additionalProperties:
-//	        type: string
+//	    description: Descriptive properties (URL encoded)
```
bensmrs, line 1138: Isn’t everything URL-encoded?

cmd/incusd/instances_post.go
```diff
@@ -1156,14 +1156,10 @@ func createFromBackup(s *state.State, r *http.Request, projectName string, data
 //	    x-example: default
 //	  - in: body
 //	    name: instance
-//	    description: Instance request
+//	    description: Instance request (or raw backup file)
 //	    required: false
 //	    schema:
 //	      $ref: "#/definitions/InstancesPost"
```
bensmrs, line 1162: At some point we’ll have to think about how to encode in swagger the fact that we accept binary data in some endpoints (OTTOMH backups, images and NVRAM), if we can do better than a simple mention in the description.

bensmrs (approved):
Approving since those are just a nitpick and a comment for the future.

cmd/incusd/instances_post.go
stgraber: Yeah, plain swagger struggles with varying response encoding. OpenAPI can handle it, but then we'd have to find a way to generate a modern OpenAPI spec rather than the older Swagger format.

cmd/incusd/images.go
stgraber: Hmm, no, it's actually correct to flag this one as it's extra weird in the way it's encoded/decoded.

https://github.com/lxc/incus/blob/main/cmd/incusd/images.go#L928

We used to have something similar for aliases but we now support passing just a simple value through so folks playing with curl don't pull their hair out.

https://github.com/lxc/incus/blob/main/cmd/incusd/images.go#L937

bensmrs: Erf, you’re reminding me why I refrain from opening any file called `image.go` and `images.go` in the codebase.

PR #3814, opened 2026-08-11
Port more Jenkins tests to Github Actions (VM)

stgraber:


test/suites/guestapi_vm.sh
```diff
@@ -0,0 +1,111 @@
...
+
+    # guestapi is enabled by default and should work
+    incus exec v1 -- curl -s --unix-socket /dev/incus/sock http://custom.socket/1.0 | jq
+    incus exec v1 -- curl -s --unix-socket /dev/incus/sock http://custom.socket/1.0/devices | jq
+    incus exec v1 -- curl -s --unix-socket /dev/incus/sock http://custom.socket/1.0/config | jq
+    incus exec v1 -- curl -s --unix-socket /dev/incus/sock http://custom.socket/1.0/meta-data | grep -q '#cloud-config'
+
+    # Run sync before forcefully restarting the VM otherwise the filesystem will be corrupted.
+    incus exec v1 -- "sync"
+    incus restart -f v1
```
bensmrs: Why a force restart? Soft rebooting is only gonna take a few more seconds.

test/suites/storage_vm.sh
```diff
@@ -0,0 +1,407 @@
...
+        return
+    fi
+
+    if [ ! -e /dev/kvm ] || ! command -v "qemu-system-$(uname -m)" > /dev/null 2>&1; then
+        echo "==> SKIP: QEMU and KVM needed to run virtual machines"
+        export TEST_UNMET_REQUIREMENT="QEMU and KVM needed to run virtual machines"
+        return
+    fi
+
+    poolDriverList="dir btrfs lvm lvm-thin zfs"
```
bensmrs: Can’t you easily add the remote storage drivers?

bensmrs (approved):
2 nits that repeat throughout the PR

test/suites/guestapi_vm.sh
stgraber: Leftover from the old Jenkins testsuite. I guess we must have had some issues with VM shutdown a while back. I'll switch everything to proper restarts and drop the sync call too.

test/suites/storage_vm.sh
stgraber: I wouldn't say easily, but yeah, probably worth a shot to split that out of the testsuite and make it a matrix job with the various storage drivers, then add ceph and linstor to the mix.

The test VMs are all Debian 13 so the exact setup logic will have to differ from what we have on the Ubuntu 24.04 based Github Actions environment.

internal/server/events/events.go
```diff
@@ -166,17 +166,30 @@ func (s *Server) broadcast(event api.Event, eventSource EventSource) error {
...
 
-	// If a notification hook is present, then call it for locally produced events.
-	// This can be used to send local events to another target (such as an event-hub member).
-	if s.notify != nil && eventSource == EventSourceLocal {
-		s.notify(event)
+	notify := s.notify
+	listeners := make([]*Listener, 0, len(s.listeners))
+	for _, listener := range s.listeners {
+		listeners = append(listeners, listener)
+	}
```
bensmrs, line 173: Why didn’t you use `copy`?

internal/server/storage/drivers/driver_linstor_utils.go
```diff
@@ -931,6 +936,15 @@ func (d *linstor) resizeVolume(vol Volume, sizeBytes int64) error {
...
 	}
 
+	// LINSTOR cannot resize volume definitions with an exact size, so clear the
+	// property (it cannot be re-enabled while resources are deployed).
+	err = linstor.Client.ResourceDefinitions.Modify(context.TODO(), resourceDefinition.Name, linstorClient.GenericPropsModify{
+		DeleteProps: []string{"DrbdOptions/ExactSize"},
+	})
+	if err != nil {
+		return fmt.Errorf("Unable to remove the exact size property: %w", err)
+	}
```
bensmrs, line 946: Wow that one feels weird :)

bensmrs (approved):
Just one nit on `copy` but otherwise LGTM

internal/server/storage/drivers/driver_linstor_utils.go
stgraber: Yeah, that one is a bit of a whacky LINSTOR thing... That property can't be set on resize and can't be set after a resize... So we effectively have to lose the nice accurate size reporting if we go through a resize...

internal/server/events/events.go
stgraber: s.listeners is a map, listeners is a slice.
There may be some helper function these days to get all the map values as a slice which can then be copied, but I went for the oldschool way.

bensmrs: Ok, didn’t notice. Yeah, you have `maps.Keys` that does the job, and we’re using it at other places in the codebase. I won’t bother you with that, but… Those modernisms feel a bit cleaner IMO.

stgraber: Yeah, I wouldn't be surprised if golangci-lint eventually picks up the pattern and forces us to fix them all in one shot. That'd be better than a mix of syntax.

PR #3820, opened 2026-08-11
incusd/instance/qemu: Record maxcpus in boot state

### LINKED ISSUE #3818: VM live migration between two servers fails when the hosts report different CPU counts, because `-smp maxcpus` is host-derived and not recorded in the boot state

### Incus system details

```yaml
IncusOS 202608021451
```

### Instance details

_No response_

### Instance log

_No response_

### Current behavior

For an instance whose `limits.cpu` is a plain count — including the default of no `limits.cpu` at all — `qemuCPU()` takes the non-fixed-topology branch and derives `maxcpus` from the **host's** live CPU thread count (`internal/server/instance/drivers/driver_qemu_templates.go:601-625`):

```go
maxCpus := 64
if int(cpu.Total) < maxCpus {
    maxCpus = int(cpu.Total)
} else if opts.cpuRequested > maxCpus {
    maxCpus = opts.cpuRequested
} else if opts.cpuCount > maxCpus {
    maxCpus = opts.cpuCount
}
entries["maxcpus"] = fmt.Sprintf("%d", maxCpus)
```

`cpu.Total` counts CPU **threads** across all sockets (`shared/api/resource.go:71-73`, incremented per thread at `shared/resources/cpu.go:540`).

`maxcpus` is not a field of `qemuBootState` (`internal/server/instance/drivers/driver_qemu_stateful.go:10-19`), which records `CPUType`, `CPUTopology`, `MachineType` and `MemoryTopology`. Those are replayed on the target rather than recomputed (`driver_qemu.go:1793`, `:1803`). `maxcpus` has no such record, so the migration target recomputes it from its own host and starts QEMU with a different `-smp` line than the source.

QEMU requires both ends of a migration to be launched with the same arguments (https://www.qemu.org/docs/master/devel/migration/main.html). When they are not, the migration fails:

```
qemu-system-x86_64: Missing section footer for 0000:00:1f.0/ICH9LPC
qemu-system-x86_64: load of migration failed: Invalid argument: Section footer error, section_id: 17
```

Because the value is capped at 64, this only bites when at least one host has fewer than 64 threads and the two totals differ. Members of one cluster are normally identical machines, which is why intra-cluster live migration is unaffected. It also fires between two otherwise identical hosts when SMT is disabled on one, since `Total` counts threads.

After a failure the source VM is sometimes left paused (`FROZEN`) and needs `incus unfreeze`. Observed in 2 of 8 failing runs; in the others the source resumed by itself.

**Mechanism.** Traced in QEMU 11.0.3 source; every link below was read directly, but we could not isolate it in a standalone QEMU-to-QEMU reproduction, so treat the section attribution as our best reading rather than a measured fact. The defect and the fix above do not depend on it. `maxcpus` sets the possible-CPU count, which becomes `state->dev_count` (`hw/acpi/cpu.c:225`). The `ICH9LPC` vmstate embeds the PM state (`hw/isa/lpc_ich9.c:814,821`), which carries the subsection `ich9_pm/cpuhp` (`hw/acpi/ich9.c:194-203`), which serialises a variable-length array sized by `dev_count` (`hw/acpi/cpu.c:308-318`; macro at `include/hw/acpi/cpu.h:73-75`). `vmstate_n_elems` reads such counts from each side's own struct, never from the stream (`migration/vmstate.c:84-98`), so the two ends disagree on the element count inside that section. On QEMU 11.0.3 that subsection is always present on q35 (`.needed = cpuhp_needed` -> `mc->has_hotpluggable_cpus`, set true at `hw/i386/pc.c:1654`).

### Expected behavior

Live migration succeeds between hosts with different CPU counts. Recording the computed `maxcpus` in `volatile.vm.boot_state` and replaying it on the target would make the QEMU command line reproducible, exactly as is already done for the CPU model, machine type, CPU topology and memory topology. QEMU permits `maxcpus` greater than the host's CPU count, so replaying a larger value onto a smaller host is safe.

*Suggested Fix:** record the computed `maxcpus` in `qemuBootState` and reuse it on the target, as `CPUType`, `MachineType`, `CPUTopology` and `MemoryTopology` already are.

**Note:** the same host-derived logic exists in Canonical LXD (`lxd/instance/drivers/driver_qemu_templates.go`, `qemuCPU`).

### Steps to reproduce

Two Incus 7.3 servers (QEMU 11.0.3), `hostA` with 8 CPU threads and `hostB` with 18 — any two different totals below 64 — same architecture and CPU model. The two hosts can simply be two VMs given different vCPU counts, which is how this was found; live migration between servers with *equal* thread counts works fine in the same environment, including on the non-shared-storage path, so the nesting itself is not a factor.

```
incus launch images:ubuntu/noble vm1 --vm -c migration.stateful=true -c limits.cpu=1
# the host-derived slot count reaches the guest:
incus exec vm1 -- cat /sys/devices/system/cpu/possible   # 0-7 on hostA, 0-17 on hostB
incus exec vm1 -- dmesg | grep "smpboot: Allowing"       # "Allowing 8 CPUs, 7 hotplug CPUs"

incus move hostA:vm1 hostB:
# Error: Failed migration on source: Error from migration control target:
#   Failed restoring checkpoint from source: QEMU crashed on VM restore:
#   qemu-system-x86_64: Missing section footer for 0000:00:1f.0/ICH9LPC
```

Forcing both ends to agree on `maxcpus`, changing nothing else — `limits.cpu` stays a plain count, so the same code branch and the same `smp-opts` shape are used — makes the identical migration succeed:

```
incus stop vm1
incus config set vm1 raw.qemu.conf '[smp-opts]
maxcpus = "12"'
incus start vm1
incus move hostA:vm1 hostB:   # succeeds
```

`12` matches neither host, so the requirement is that the two ends agree, not that the value match the host. Setting an explicit topology (`limits.cpu=sockets=1,cores=1,threads=1`), which takes the fixed-topology branch and emits no `maxcpus`, also makes it succeed.

stgraber:
Closes #3818

### DIFF
--- a/internal/server/instance/drivers/driver_qemu_templates.go
+++ b/internal/server/instance/drivers/driver_qemu_templates.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/lxc/incus/v7/internal/server/instance/drivers/cfg"
 	"github.com/lxc/incus/v7/shared/osarch"
-	"github.com/lxc/incus/v7/shared/resources"
 )
 
 func writeHeader(sb *strings.Builder, comment string, name string) {
@@ -559,7 +558,7 @@ type qemuNumaEntry struct {
 type qemuCPUOpts struct {
 	architecture     int
 	cpuCount         int
-	cpuRequested     int
+	cpuMaxCpus       int
 	cpuSockets       int
 	cpuCores         int
 	cpuThreads       int
@@ -605,23 +604,8 @@ func qemuCPU(opts *qemuCPUOpts, pinning bool) []cfg.Section {
 		entries["sockets"] = fmt.Sprintf("%d", opts.cpuSockets)
 		entries["cores"] = fmt.Sprintf("%d", opts.cpuCores)
 		entries["threads"] = fmt.Sprintf("%d", opts.cpuThreads)
-	} else {
-		cpu, err := resources.GetCPU()
-		if err != nil {
-			return nil
-		}
-
-		// Cap the max number of CPUs to 64 unless directly assigned more.
-		maxCpus := 64
-		if int(cpu.Total) < maxCpus {
-			maxCpus = int(cpu.Total)
-		} else if opts.cpuRequested > maxCpus {
-			maxCpus = opts.cpuRequested
-		} else if opts.cpuCount > maxCpus {
-			maxCpus = opts.cpuCount
-		}
-
-		entries["maxcpus"] = fmt.Sprintf("%d", maxCpus)
+	} else if opts.cpuMaxCpus > 0 {
+		entries["maxcpus"] = fmt.Sprintf("%d", opts.cpuMaxCpus)
 	}
 
 	sections := []cfg.Section{{
--- a/internal/server/instance/drivers/driver_qemu.go
+++ b/internal/server/instance/drivers/driver_qemu.go
@@ -1799,6 +1799,15 @@ func (d *qemu) start(stateful bool, op *operationlock.InstanceOperation) error {
 		bs.CPUType = cpuType
 	}
 
+	if bs.MaxCPUs == 0 {
+		maxCPUs, err := d.maxCPUs(bs.CPUTopology)
+		if err != nil {
+			return err
+		}
+
+		bs.MaxCPUs = maxCPUs
+	}
+
 	// Setup the memory.
 	if bs.MemoryTopology == nil {
 		// Get the memory topology.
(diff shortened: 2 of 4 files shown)


internal/server/instance/drivers/driver_qemu_templates.go
```diff
@@ -605,23 +604,8 @@ func qemuCPU(opts *qemuCPUOpts, pinning bool) []cfg.Section {
...
-		if int(cpu.Total) < maxCpus {
-			maxCpus = int(cpu.Total)
-		} else if opts.cpuRequested > maxCpus {
-			maxCpus = opts.cpuRequested
-		} else if opts.cpuCount > maxCpus {
-			maxCpus = opts.cpuCount
-		}
-
-		entries["maxcpus"] = fmt.Sprintf("%d", maxCpus)
+	} else if opts.cpuMaxCpus > 0 {
```
bensmrs, line 607: Isn’t it now always greater than 0?

stgraber: I wish :)

Anyway, no, it's not. Because when dealing with those bootstate flags you always need to handle migrations of running VMs that were started prior to the field's introduction. In that scenario, that would lead to a 0 value.

bensmrs: Ah, I always forget about those…

bensmrs (approved):

PR #3828, opened 2026-08-13
lxc: Make incremental memory migration opt-in

fivetime:
The `migration.incremental.memory` configuration is documented with a default of false, but the LXC migration sender currently enables pre-dumps when the key is unset. It also probes the CRIU dirty-memory feature even when the setting is explicitly false.

Treat incremental memory migration as opt-in and only probe CRIU when the expanded configuration is explicitly true. Preserve the existing default iteration count and upper bound.

The driver package tests and `go vet` pass on Linux.

### DIFF
--- a/internal/server/instance/drivers/driver_lxc.go
+++ b/internal/server/instance/drivers/driver_lxc.go
@@ -6015,6 +6015,13 @@ func getCRIULogErrors(imagesDir string, method string) (string, error) {
 
 // Check if CRIU supports pre-dumping and number of pre-dump iterations.
 func (d *lxc) migrationSendCheckForPreDumpSupport() (bool, int) {
+	config := d.ExpandedConfig()
+
+	// Incremental memory migration is opt-in, matching the documented default and avoiding an unnecessary CRIU probe.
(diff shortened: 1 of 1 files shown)


internal/server/instance/drivers/driver_lxc_migration_test.go
```diff
@@ -0,0 +1,43 @@
+package drivers
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestMigrationPreDumpSettings(t *testing.T) {
```
stgraber: This PR is making a bunch of completely needless changes to the code just to allow this test to exist. At the end of the day, all this test does is check if util.IsTrue in fact returns true when passed true.

It's completely pointless and should be removed, allowing the code itself to actually be readable.

fivetime: Agreed. I removed the test and the helper introduced solely for it, and restored the check to the migration control flow directly.

stgraber:

fivetime:

PR #3829, opened 2026-08-13
storage/ceph: Ignore detached devices during RBD sysfs scan

fivetime:
An RBD device can be unmapped after `/sys/devices/rbd` is listed but before its attributes are read. In addition to ENOENT, sysfs returns ENODEV for attributes of a half-removed device.

Treat ENODEV as a vanished scan entry for the pool, name and snapshot attributes, while continuing to return permission and other I/O errors. This prevents an unrelated concurrent unmap from aborting mapped-device discovery.

The storage drivers package tests and `go vet` pass on Linux.

### DIFF
--- a/internal/server/storage/drivers/driver_ceph_utils.go
+++ b/internal/server/storage/drivers/driver_ceph_utils.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	"github.com/google/uuid"
+	"golang.org/x/sys/unix"
 
 	"github.com/lxc/incus/v7/internal/linux"
 	"github.com/lxc/incus/v7/internal/server/db"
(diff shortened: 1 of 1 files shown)


internal/server/storage/drivers/driver_ceph_utils_test.go
```diff
@@ -2,9 +2,31 @@ package drivers
...
 import (
 	"fmt"
+	"io/fs"
 	"testing"
+
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sys/unix"
 )
 
+func TestDetachedRBDSysfsReadError(t *testing.T) {
```
stgraber: Please drop the test, it's completely pointless given it's about as useful as checking that 1 == 1... The function is trivial and contains no real logic worth testing.

fivetime: Agreed. I removed the test and helper and inlined the ENOENT/ENODEV checks at the two sysfs reads.

stgraber:

fivetime:

PR #3831, opened 2026-08-13
lxc: Normalize received CRIU state ownership

fivetime:
CRIU migration state can contain mixed ownership: most images are created by the outer daemon as root, while namespace-owned files retain the source container's shifted IDs. Applying the target idmap directly to that mixed tree leaves the source-shifted files with incorrect target ownership.

Before restore, unshift only files whose UID or GID belongs to the source idmap. Files created by the outer daemon do not map back into the source namespace and are deliberately left unchanged before the existing target shift is applied.

The LXC drivers package tests and `go vet` pass on Linux.

### DIFF
--- a/internal/server/instance/drivers/driver_lxc.go
+++ b/internal/server/instance/drivers/driver_lxc.go
@@ -7213,6 +7213,22 @@ func (d *lxc) MigrateReceive(args instance.MigrateReceiveArgs) error {
 
 			d.logger.Debug("Done receiving final dump rsync")
 
+			// CRIU writes most images as the outer daemon's root user, but some namespace-owned
+			// images retain the source container's shifted IDs. Normalize only those shifted
+			// owners before migrate() applies the target container's idmap to the tree.
+			if len(srcIdmap.Entries) > 0 {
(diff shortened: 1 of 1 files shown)


internal/server/instance/drivers/util_test.go
```diff
@@ -34,6 +34,13 @@ func TestRoundToBlockSize(t *testing.T) {
 	require.Equal(t, int64(0), value)
 }
 
+func TestCRIUStateUnshiftSkipper(t *testing.T) {
```
stgraber: Again with the pointless test which just makes the origin function harder to read...

Please drop this test and please drop criuStateUnshiftSkipper and errCRIUStateOwnerNotShifted and just move that logic into the parent function. Either passing a function in-line to UnshiftPath or defining the function locally in MigrationReceive as we don't have a use for it elsewhere.

fivetime: Agreed. I removed the test, sentinel error, and global skipper, and inlined the ownership validation in the ShiftOwner callback.

stgraber:

fivetime:

PR #3832, opened 2026-08-13
Allow concurrent instance creation from one image

fivetime:
Instance creation currently takes an exclusive per-image lock for the full storage copy. Independent creations from the same already-present image are serialized even though they only read the image and cached image volume.

Add context-aware, writer-prioritized named read-write locks. Image consumers take shared locks, while image download/replacement/deletion and cached-volume deletion remain exclusive. The storage layer separately protects the cached image volume so expiry or regeneration cannot remove it while a copy is in progress.

The locking tests pass with the race detector. The locking, storage and incusd package tests and `go vet` pass on Linux.

### DIFF
--- a/cmd/incusd/daemon_images.go
+++ b/cmd/incusd/daemon_images.go
@@ -52,13 +52,22 @@ type imageDownloadArgs struct {
 	SourceProjectName string
 }
 
-// imageOperationLock acquires a lock for operating on an image and returns the unlock function.
+// imageOperationLock acquires an exclusive lock for creating, replacing, or deleting an image.
 func imageOperationLock(ctx context.Context, fingerprint string) (locking.UnlockFunc, error) {
 	l := logger.AddContext(logger.Ctx{"fingerprint": fingerprint})
 	l.Debug("Acquiring lock for image")
 	defer l.Debug("Lock acquired for image")
 
-	return locking.Lock(ctx, fmt.Sprintf("ImageOperation_%s", fingerprint))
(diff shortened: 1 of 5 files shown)


stgraber:
@fivetime do you have metrics proving that this actually helps?
Unless I missed something, a regular launch will still go through imageDownload, even if the image already exists locally. This will then grab the read/write lock and basically block there if another instance currently has the read lock.

This looks like a LOT of hand rolled locking code for effectively no benefit, in fact, it looks like you may be making things worse in the event where EnsureImage decides that it needs to replace the image that's currently present.

It looks like it would be possible for two users with different image needs to now hit EnsureImage concurrently. If the first gets a match, it continues, then the second isn't a match which causes the image to be deleted and re-created, while the first is still actively reading from it, getting you a potential partial/corrupted image or a failure.

fivetime:
Thanks, you're right to question this.

The measurements I have show the existing per-fingerprint serialization is a real bottleneck (an 8-create same-image run took 126.56s, while splitting the same 8 creates across two fingerprints took 78.55s), but I do not have a controlled before/after benchmark proving this patch helps.

More importantly, after tracing the full path again, the current patch is not safe. The read lock is acquired after `EnsureImage` returns. Another request can therefore replace the cached image between validation and acquisition of the read lock, so the first request may consume an image that no longer matches the conditions it validated. The read lock prevents deletion during the copy, but it does not make validation and consumption atomic.

I don't think adding more hand-rolled locking to this PR is the right response. I am withdrawing it. A future proposal would need to preserve the validation-to-consumption invariant, handle replacement explicitly, and include controlled before/after measurements and concurrency fault tests.

PR #3838, opened 2026-08-13
Extend image locking

stgraber:


cmd/incusd/images.go
```diff
@@ -2455,6 +2463,14 @@ func autoUpdateImage(ctx context.Context, s *state.State, op *operations.Operati
 		return nil, nil
 	}
 
+	// Lock this operation to ensure that concurrent image operations don't conflict.
+	unlock, err := imageOperationLock(ctx, fingerprint)
+	if err != nil {
+		return nil, err
```
bensmrs, line 2469: You added a log message above, but not here.

cmd/incusd/images.go
```diff
@@ -2704,7 +2720,13 @@ func pruneExpiredImages(ctx context.Context, s *state.State, op *operations.Oper
 // pruneExpiredImage removes an expired image's database records and, once unused by all projects,
 // its storage volumes and files.
 func pruneExpiredImage(ctx context.Context, s *state.State, op *operations.Operation, fingerprint string, dbImages []dbCluster.Image, projectsImageRemoteCacheExpiryDays map[string]int64) error {
-	var err error
+	// Lock this operation to ensure that concurrent image operations don't conflict.
+	unlock, err := imageOperationLock(ctx, fingerprint)
+	if err != nil {
+		return err
```
bensmrs, line 2726: Same.

bensmrs (changes requested):

cmd/incusd/images.go
stgraber: Above it's because it's not fatal (continue). Here it's fatal and so will be returned.

cmd/incusd/images.go
stgraber: Same, this one is being returned as an error so doesn't need the logging that we do when we just continue.

bensmrs (approved):

PR #3839, opened 2026-08-13
Improvements to `incus remote add`

stgraber:


cmd/incus/remote.go
```diff
@@ -403,7 +489,7 @@ func normalizeAddress(addr string) (string, string, error) {
 	return addr, rScheme, nil
 }
 
-func (c *cmdRemoteAdd) run(cmd *cobra.Command, args []string) error {
+func (c *cmdRemoteAdd) run(cmd *cobra.Command, args []string) (err error) {
```
bensmrs, line 492: So that’s one of the Go quirks I’m not comfortable with. Why is that needed?

cmd/incus/remote.go
```diff
@@ -1378,6 +1378,20 @@ func (c *cmdRemoteRename) run(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	// Rename the client certificate files.
+	if !conf.Remotes[remoteName].Global {
```
bensmrs, line 1382: I had completely forgotten about those…

bensmrs (approved):

cmd/incus/remote.go
stgraber: See the `defer func` which then reacts to `err`?

That's what having a named return lets you do. It basically defines the variable for you within the scope of the function and lets you access it safely from a defer.

We generally don't use them in Incus as it's a Go-ism we haven't found particularly useful in general, some projects use them everywhere to save themselves having to `:=` the variable and I believe you can also then omit the final return?

But in Incus, we only use that syntax in a couple of spots where we do want a defer which can reliably pick up the return value(s) from the parent function.

cmd/incus/remote.go
stgraber: Ah yeah. Not sure how many people actually use the /etc/incus/ stuff.

At least one given that otherwise we wouldn't have implemented it, but there probably are some organizations that pre-configure all their remotes that way on their bastions or whatever :)

PR #3840, opened 2026-08-13
Various bugfixes

stgraber:


bensmrs:
Haha, looks like someone shot himself in the foot with this :)

stgraber:
> Haha, looks like someone shot himself in the foot with this :)

Just a tiny bit :)

Pretty sure it's a case of our tests being wrong, we shouldn't mess with features.networks=true if we don't have OVS/OVN.

bensmrs (approved):
Shouldn’t we also add a startup patch to fix existing mistakes?

stgraber:
> Shouldn’t we also add a startup patch to fix existing mistakes?

I don't think it's strictly needed. A project configured that way would be completely unable to start an instance. And a project without any instance can have its feature set adjusted to turn off the feature.

bensmrs:
Fair enough

PR #3841, opened 2026-08-13
Fix LVM cluster deletion issue when deleting the initial pool

stgraber:


stgraber:
TLDR on this one is that sanlock uses an LV on a ONE VG for the lock state. If you decide to delete that VG, sanlock becomes very sad about any other VG on the system.

We need to detect this issue and effectively promote another VG to be the new global lock.

internal/server/storage/drivers/driver_lvm_utils.go
```diff
@@ -99,13 +99,20 @@ func (d *lvm) openLoopFile(source string) (string, error) {
 // meaning that the object was not found. Returns true if it is (exit status 5) false if not.
```
bensmrs: Nit: remove the mention in parentheses.

internal/server/storage/drivers/driver_lvm_utils.go
```diff
@@ -99,13 +99,20 @@ func (d *lvm) openLoopFile(source string) (string, error) {
...
+		return false
+	}
+
+	// LVM uses exit status 5 for all command failures, so check the error output to
+	// tell missing objects apart from other failures (e.g. lock manager errors).
+	var runError subprocess.RunError
+	if errors.As(err, &runError) {
+		stderr := runError.StdErr().String()
+
+		return strings.Contains(stderr, "not found") || strings.Contains(stderr, "Failed to find") || strings.Contains(stderr, "No physical volume label read")
```
bensmrs, line 113: As with encoders, double-check that localization doesn’t break your checks.

bensmrs (commented):
Not my specialty, that sanlock dance :)
Still, a few nits.

internal/server/storage/drivers/driver_lvm_utils.go
stgraber: I think we have a wrapper to force use of C as the locale. Though this is also in the daemon with normal systemd units not usually setting the locale variables for their services.

bensmrs (approved):

PR #3849, opened 2026-08-15
Improve cluster response times

stgraber:


internal/server/cluster/events.go
```diff
@@ -270,12 +270,27 @@ func EventsUpdateListeners(s *state.State, hbMembers map[int64]APIHeartbeatMembe
...
 			continue
 		}
 
 		if localEventMode != EventModeFullMesh && !RoleInSlice(db.ClusterRoleEventHub, hbMember.Roles) {
+			// No direct connection is made to this member in event-hub mode, so clear any
+			// stale unavailable flag left over from an offline period.
+			listenersLock.Lock()
+			delete(listenersUnavailable, hbMember.Address)
+			listenersLock.Unlock()
+
```
bensmrs, line 293: Why is that needed?

cmd/incusd/images.go
```diff
@@ -5081,16 +5086,38 @@ func imageSyncBetweenNodes(ctx context.Context, s *state.State, r *http.Request,
...
+				}
+			}
+
+			return nil
 		})
 		if err != nil {
 			return fmt.Errorf("Failed to get nodes for the image synchronization: %w", err)
 		}
 
 		if len(addresses) <= 0 {
```
bensmrs: That’s a weird check.

bensmrs (approved):

cmd/incusd/images.go
stgraber: Hmm, yeah, no idea what that's about, will fix to == 0 as len() can't really return a negative value. Speaking of, why is len() an int and not a uint...

internal/server/cluster/events.go
stgraber: Cleanup for when we're adding cluster roles and effectively switch topology from full mesh to hub mode.

PR #3859, opened 2026-08-17
incusd/instance/lxc: Keep OCI VOLUME paths persistent instead of tmpfs

jochumdev:
Paths declared as VOLUME in an OCI/Docker image config were being converted to ephemeral tmpfs mounts, silently discarding their contents on every instance restart. Our instances already have their own persistent storage, so skip those mounts instead and leave the path as a normal part of the instance rootfs.

Identified via the org.opencontainers.image.volumes annotation ([opencontainers/umoci#678](https://github.com/opencontainers/umoci/pull/678)), falling back to the source == "none" heuristic for older umoci versions that don't set it.

This has been in umoci since v0.0.0-rc2 (2016) so `VOLUMES` in Dockerfiles have been "tmpfs" since then.

### DIFF
--- a/internal/server/instance/drivers/driver_lxc.go
+++ b/internal/server/instance/drivers/driver_lxc.go
@@ -2588,6 +2588,14 @@ func (d *lxc) startCommon() (string, []func() error, error) {
 			lxcMounts = append(lxcMounts, filepath.Clean(fmt.Sprintf("/%s", fields[1])))
 		}
 
+		// OCI VOLUME paths, tracked through the org.opencontainers.image.volumes annotation (PR #678).
+		ociVolumes := make(map[string]bool)
+		for _, vol := range strings.Split(config.Annotations["org.opencontainers.image.volumes"], ",") {
+			if vol != "" {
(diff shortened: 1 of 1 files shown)


stgraber:
As far as I can tell, the option here isn't part of the OCI specification.
So we'd want to wait for it to be in an OCI spec and properly supported by umoci before doing anything with it.

Looking at the logic, the first part of the loop has:
```
// We only support simple tmpfs at this stage.
if len(mount.UIDMappings) > 0 || len(mount.GIDMappings) > 0 || mount.Type != "tmpfs" {
    continue
}
```

So we're specifically only mounting a `tmpfs` if the provided OCI `image.json` specifically tells us to mount a tmpfs. Given a legitimate tmpfs will always have a Source of "none", this change actually appears to regress every single legitimate tmpfs requests.

We also appear to be implementing exactly what the JSON file is requesting us to do, so it doesn't seem like a good idea to put a hack in this part of the code. Instead it would be better for umoci or possibly our own repacker to alter the JSON data to reflect that those paths should in fact NOT be tmpfs.

PR #3865, opened 2026-08-18
incus/server/device/nic/ovn: hot-reloadable ovn nic limits

DarkressX:


internal/server/device/nic_ovn.go
```diff
@@ -1160,7 +1160,13 @@ func (d *nicOVN) Update(oldDevices deviceConfig.Devices, isRunning bool) error {
 	}
 
 	// Apply any changes needed when assigned ACLs or external NAT addresses change.
```
stgraber: The comment could do with an update.

stgraber:
@DarkressX I'm assuming that you've tested this and confirmed that InstanceDevicePortStart indeed takes care of applying any such changes?

internal/server/device/nic_ovn.go
DarkressX: Good point

DarkressX:
> @DarkressX I'm assuming that you've tested this and confirmed that InstanceDevicePortStart indeed takes care of applying any such changes?

I did test this.
Changing the limits will remove the qos rule of the port and re add it with updated values.
In OVN this is represented by meters, which can also be hotplugged.

stgraber (approved):

PR #3869, opened 2026-08-20
Add incus start --boot-next=xxx and more NVRAM improvements

bensmrs:


cmd/incus/action.go
```diff
@@ -155,6 +161,7 @@ func (c *cmdAction) command(action string) *cobra.Command {
 		cli.AddBoolFlag(cmd.Flags(), &c.flagStateful, "stateful", i18n.G("Store the instance state"))
 	case "start":
 		cli.AddBoolFlag(cmd.Flags(), &c.flagStateless, "stateless", i18n.G("Ignore the instance state"))
+		cli.AddStringFlag(cmd.Flags(), &c.flagBootNext, "boot-next", "", "", i18n.G("Attempt to boot a specific boot option (set to `ask` for an interactive choice)"))
```
stgraber: There is some magic we've done with `--console` which lets you detect `--boot-next` being passed with no value which would let you get rid of `ask` but just having `--boot-next` for the menu or `--boot-next="foo"` for a specific option.

bensmrs: Renamed `--override-boot` as it makes more sense if the option is to be used without a parameter.

stgraber (approved):

PR #3871, opened 2026-08-20
cmd/incus: add --data-file flag for query command

breml:
@stgraber One thing to point out: since the file is passed as io.Reader to the client (RawQuery), the content type is set to `application/octet-stream` regardless of the actual payload in the file. I wonder, if additional flags (e.g. `--data-file-json` and `--data-file-binary`) are required here to support more accurate content type headers.

### DIFF
--- a/cmd/incus/query.go
+++ b/cmd/incus/query.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"os"
 	"slices"
 	"strings"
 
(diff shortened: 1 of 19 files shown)


stgraber:
@breml commit should be titled "incus/query: Add --data-file flag" to better line up with our usual pattern. You're also missing a "make i18n" run (committed as "i18n: Update translation templates").

For the Content-Type. I think we're likely fine with octet-stream for now. If that becomes a problem, we'll probably want a RawQueryHeaders or something which can pass headers through and then we can implement a "-H" like curl.

breml:
@stgraber PR updated

breml:
@stgraber for this to be useful in my use case, we need to add `-H` already now, since I need the possibility to pass `X-Incus-mode`.

stgraber:
> @stgraber for this to be useful in my use case, we need to add `-H` already now, since I need the possibility to pass `X-Incus-mode`.

Does the transfer fail without it? Otherwise, you can always chown/chmod through exec afterwards :)

But anyway, introducing support for `-H` by exposing a RawQueryHeaders or similar from the client package would be fine to introduce, feel free to send a follow-up PR if useful!

PR #3875, opened 2026-08-21
Allow sharing specific networks with restricted projects

stgraber:
This allows for sharing of specific networks from the `default` project with another project which has both `features.networks=true` (OVN) and `restricted=true` enabled.

In that scenario, setting `restricted.networks.access` allows read-only access of specific managed networks.

The reason for all this is to have projects which are otherwise isolated with their own network view (OVN) to still be passed a select set of administrator-managed networks.

A concrete example of this would be an environment where a tenant has access to a project with OVN enabled. They can create their own OVN networks but there's also a physical network (VLAN) which they need to be able to access.

That's exactly the case we're running into with the Linux Containers lab where the majority of our networks are OVN networks for Github runners or similar builders, but we also have a physical lab with physical machines sitting on a VLAN which we need to access from within the environment.


cmd/incusd/daemon.go
```diff
@@ -383,6 +383,11 @@ func allowPermission(objectType auth.ObjectType, entitlement auth.Entitlement, m
 				return project.ImageProjectFromRecord(p)
 			}
 
+			if objectType == auth.ObjectTypeNetwork && len(muxVars) > 0 {
```
bensmrs, line 386: For my culture, when does `len(muxVars) == 0` in that context?

cmd/incusd/patches.go
```diff
@@ -107,6 +107,7 @@ var patches = []patch{
 	{name: "authorization_client_routes_from_driver", stage: patchPreDaemonStorage, run: patchAuthorizationClientRoutesFromDriver},
 	{name: "network_ovn_l2proxy_arp_proxy", stage: patchPostDaemonStorage, run: patchGenericNetwork(patchNetworkOVNL2ProxyARPProxy)},
 	{name: "auth_openfga_instance_tcp", stage: patchPostNetworks, run: patchGenericAuthorization},
+	{name: "auth_openfga_shared_networks", stage: patchPostNetworks, run: patchGenericAuthorization},
```
bensmrs, line 110: Do we have to add a patch each time we change authorization models?

bensmrs (approved):

cmd/incusd/patches.go
bensmrs: And I’m curious what happens at the very first daemon startup; do the generic patches get run dozens of times?

stgraber: A completely fresh Incus daemons starts up with all patches already applied, so none of them run.

We do indeed have to put a call to patchGenericAuthorization every time we need a refresh of the OpenFGA model. We used to try to auto-detect the condition but that was very unreliable. Pushing a new model on every startup also isn't ideal as OpenFGA actually keeps older models around, so we'd quickly be using a fair bit of space for nothing.

bensmrs: Thanks, you answered my followup question ;)

cmd/incusd/daemon.go
stgraber: Just panic prevention to avoid a potential access of a non-existent index. Though currently all codepaths using a network object type do pass at least a name through.

bensmrs: Still, there are plenty of places in this function that don’t check it, so we may wanna be a tad more consistent on those.

PR #3876, opened 2026-08-21
Support for nic burst limits

### LINKED ISSUE #2379: Support for burst limit in nic and disk device configuration

I need help enabling burst I/O for VMs in my Incus cluster.

I'm managing an Incus VM cluster for developers who typically have bursty workloads - they perform short-duration, high-intensity operations like repository cloning, dependency downloads, and compilation tasks. These workloads would benefit significantly from burst I/O capabilities rather than continuous rate limiting.

Currently, I can only find continuous limiting options such as `limits.read`, `limits.write`, `limits.ingress`, and `limits.egress`. However, I haven't been able to locate any configuration options that would allow for burst I/O behavior. Is there a way to configure burst I/O limits for VMs in Incus, or are there alternative approaches to better accommodate these bursty workload patterns?

Soft limit option for vms

```bash
# for disk
qemu -drive file=hd0,throttling.bps-total=100,throttling.bps-total-max=2000,throttling.bps-total-max-length=60
# for network interface
tc class add dev tap0 parent 1:1 classid 1:10 htb rate 1mbit ceil 10mbit
```

presztak:
Closes: https://github.com/lxc/incus/issues/2379


presztak:
@stgraber, I'm not sure if we should aim for consistency across all config keys for disks and NICs. Right now we have `limits.egress.burst` and `limits.egress.burst.length` (same pattern for ingress). But for NICs, there's actually no parameter equivalent to length. For example, on the ingress side we have the `Ceil` and `Buffer` parameters (in ClassHTB), where Buffer is actually a size, not a length in seconds, more in line with the leaky bucket concept.

I'm wondering whether we should instead have config keys that map more directly (1:1) to what we already have for OVN and netlink. So for bridged, p2p, and routed NIC types, we'd have:

ingress:
- limits.ingress.burst (Ceil)
- limits.ingress.size (Buffer)

egress:
- limits.egress.burst (PeakRate)
- limits.egress.size (Burst)

and for OVN (we only have a single burst parameter, which is actually the size of the bucket):
- limits.ingress.burst
- limits.egress.burst

stgraber:
Maybe use `.bucket` instead of `.size` so it's slightly clearer what it's about?

PR #3878, opened 2026-08-21
Add NVRAM support in QEMU scriptlet

bensmrs:
This also adds tests and fixes bugs.


bensmrs:
~~Well, I can’t get interactions with the console to behave within the CI, so I’ll patch the test into something less funny.~~ I have solved that with longer sleeps.

PR #3883, opened 2026-08-23
incusd/instance/qmp: Set a write deadline on monitor commands

stgraber:


presztak (approved):

internal/server/instance/drivers/qmp/qmp.go
bensmrs, line 398: Is that still relevant?

stgraber: Yeah, what we've done is added a context to prevent hanging during a command write. But we can still hit a timeout waiting to get a response from it.

PR #3884, opened 2026-08-23
Add support for using librbd for all Ceph interactions

stgraber:
We've seen a number of limitations and issues with `krbd` over the years. This offers an alternative by having Incus re-use our existing `qemu-nbd` logic and /dev/nbdX as the backing device for all Ceph volumes.

This causes more userspace processes and has a few limitations (live resize) but also comes with better system stability/recoverability and allows for the use of all Ceph features.

### DIFF
--- a/internal/server/storage/drivers/driver_ceph_volumes.go
+++ b/internal/server/storage/drivers/driver_ceph_volumes.go
@@ -1219,10 +1219,28 @@ func (d *ceph) SetVolumeQuota(vol Volume, size string, allowUnsafeResize bool, o
 				return err
 			}
 		} else if sizeBytes > oldSizeBytes {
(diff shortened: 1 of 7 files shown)


internal/server/storage/drivers/driver_ceph_volumes.go
```diff
@@ -1253,6 +1271,19 @@ func (d *ceph) SetVolumeQuota(vol Volume, size string, allowUnsafeResize bool, o
...
+				err = d.rbdUnmapVolume(vol, true)
+				if err != nil {
+					return err
+				}
+
+				devPath, err = d.rbdMapVolume(vol)
+				if err != nil {
+					return err
+				}
+			}
```
presztak, line 1285: Nit: We could turn that into a helper function.

presztak (approved):

PR #3885, opened 2026-08-23
Allow image architectures filters to apply to both Incus architecture and the property

stgraber:


cmd/incus/image.go
```diff
@@ -1289,6 +1289,11 @@ func (c *cmdImageList) imageShouldShow(filters []string, state *api.Image) bool
 			if ok && fmt.Sprintf("%v", val) == value {
 				found = true
 			}
+
+			// Match architecture filters against the Incus architecture name too.
+			if !found && internalFilter.DotPrefixMatch(key, "architecture") && value == state.Architecture {
```
bensmrs, line 1294: `DotPrefixMatch` feels very overkill when checking non-dotted keys :)

bensmrs (approved):

cmd/incus/image.go
stgraber: Ah yeah, it's mostly for consistency, it's really the "prefix" part I was interested in having here ;)

PR #3892, opened 2026-08-24
Manipulate NVRAM through initial keys

### LINKED ISSUE #2627: Allow adding custom certificates to the OVMF DB

bensmrs:
The initial keys are deliberately not frozen after the first boot, as i) it makes sense to force rebuilding the NVRAM with new defaults and ii) you can’t block profile modification anyway, so new profile-defined defaults would still be applied.

Closes: #2627

### DIFF
--- a/cmd/incus/low_level.go
+++ b/cmd/incus/low_level.go
@@ -1131,24 +1131,6 @@ func (c *cmdLowLevelSecureBoot) command() *cobra.Command {
 	return cmd
 }
 
-// eslGUIDVar gets the GUID and variable name associated to the given ESL.
-func eslGUIDVar(esl string) (string, string) {
-	var guid, varName string
-	switch esl {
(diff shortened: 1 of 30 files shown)


stgraber:
@bensmrs minor static analysis issues

bensmrs:
Yup, ended up doing some minor refactoring to only perform the `Set` at a single place in the code.

stgraber (approved):

PR #3893, opened 2026-08-24
Various migration/evacuation fixes

stgraber:
This is a collection of fixes from a recent set of tests auto-evacuating a very busy lab environment. The fixes basically cover:
 - Harden the client reconnection as it occasionally still disconnected
 - Handle a race where a clean shutdown completes before the forced shutdown fully runs
 - Make sure no instance creation can occur while evacuating/evacuated
 - Drain all ongoing instance creations prior to evacuating a server
 - Improve migration received errors (QEMU)
 - Improve NUMA placement when nodes are short on memory
 - Cleanup error on missing project during bulk instance state operations
 - Don't cancel ongoing evacuations at the first sign of failure, let the ongoing ones complete, then fail
 - Correctly record instance location once QEMU has resumed on target

### DIFF
--- a/cmd/incusd/instance_post.go
+++ b/cmd/incusd/instance_post.go
@@ -1037,9 +1037,11 @@ func migrateInstance(ctx context.Context, s *state.State, inst instance.Instance
 			return fmt.Errorf("Instance move to destination failed on source: %w", err)
 		}
 
-		err = destOp.Wait()
-		if err != nil {
-			return fmt.Errorf("Instance move to destination failed: %w", err)
+		// A live migration on the same shared storage is handed over once the source succeeds,
+		// so the database record must follow the instance even if the destination reported an error.
+		destErr := destOp.Wait()
+		if destErr != nil && (!req.Live || !sourcePool.Driver().Info().Remote || req.Pool != "") {
+			return fmt.Errorf("Instance move to destination failed: %w", destErr)
 		}
 
 		// Update the database post-migration.
@@ -1160,6 +1162,10 @@ func migrateInstance(ctx context.Context, s *state.State, inst instance.Instance
 			if err != nil {
 				return fmt.Errorf("Failed starting instance on target member: %w", err)
 			}
+		}
+
+		if destErr != nil {
+			return fmt.Errorf("Instance moved to %q but destination reported an error: %w", targetMemberInfo.Name, destErr)
 		}
 	}
 
--- a/internal/server/instance/drivers/driver_common.go
+++ b/internal/server/instance/drivers/driver_common.go
@@ -28,6 +28,7 @@ import (
 	deviceConfig "github.com/lxc/incus/v7/internal/server/device/config"
 	"github.com/lxc/incus/v7/internal/server/device/nictype"
 	"github.com/lxc/incus/v7/internal/server/instance"
+	"github.com/lxc/incus/v7/internal/server/instance/drivers/qemudefault"
 	"github.com/lxc/incus/v7/internal/server/instance/instancetype"
 	"github.com/lxc/incus/v7/internal/server/instance/operationlock"
 	"github.com/lxc/incus/v7/internal/server/lifecycle"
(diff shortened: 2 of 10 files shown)


internal/server/instance/drivers/driver_qemu.go
```diff
@@ -1023,13 +1023,14 @@ func (d *qemu) restoreStateHandle(ctx context.Context, monitor *qmp.Monitor, f *
...
 	if err != nil {
-		if errors.Is(err, qmp.ErrMonitorDisconnect) && util.PathExists(d.LogFilePath()) {
-			qemuError, err := os.ReadFile(d.LogFilePath())
-			if err != nil {
-				return err
+		if errors.Is(err, qmp.ErrMonitorDisconnect) {
+			// Crash output goes to stderr (early log), -D only captures QEMU's internal logging.
+			qemuError, readErr := os.ReadFile(d.EarlyLogFilePath())
+			if readErr != nil || len(strings.TrimSpace(string(qemuError))) == 0 {
+				qemuError, _ = os.ReadFile(d.LogFilePath())
```
bensmrs, line 1030: You don’t want to combine both outputs? (I dunno if there is additional output than what’s on stderr)

internal/server/instance/drivers/driver_common.go
```diff
@@ -1657,19 +1657,37 @@ func (d *common) balanceNUMANodes() error {
...
 
+	numaNodesToUse := 1
 	if err == nil && limitsCPU > cpusPerNumaNode {
-		numaNodesToUse := int(math.Ceil(float64(limitsCPU) / float64(cpusPerNumaNode)))
+		numaNodesToUse = int(math.Ceil(float64(limitsCPU) / float64(cpusPerNumaNode)))
+	}
+
+	// Similarly, if `limits.memory` is greater than the amount of memory per NUMA node,
+	// use as many NUMA nodes as needed to fit it.
+	if conf["limits.memory"] != "" {
```
bensmrs: And if it’s not set, you consider that the default fits all modern CPUs?

internal/server/instance/drivers/driver_common.go
```diff
@@ -1657,19 +1657,37 @@ func (d *common) balanceNUMANodes() error {
...
+		if err == nil && len(nodes) > 0 {
+			memoryPerNumaNode := int64(memory.Total) / int64(len(nodes))
+			if memoryPerNumaNode > 0 && limitsMemory > memoryPerNumaNode {
+				numaNodesToUse = max(numaNodesToUse, int(math.Ceil(float64(limitsMemory)/float64(memoryPerNumaNode))))
+			}
 		}
+	}
+
+	// Cap at the number of available NUMA nodes.
+	numaNodesToUse = min(numaNodesToUse, len(nodes))
```
bensmrs, line 1690: Don’t you want to fail early if there are fewer nodes that what you computed?

cmd/incusd/instance_post.go
```diff
@@ -1037,9 +1037,11 @@ func migrateInstance(ctx context.Context, s *state.State, inst instance.Instance
 			return fmt.Errorf("Instance move to destination failed on source: %w", err)
 		}
 
-		err = destOp.Wait()
-		if err != nil {
-			return fmt.Errorf("Instance move to destination failed: %w", err)
+		// A live migration on the same shared storage is handed over once the source succeeds,
```
bensmrs, line 1040: Why on the same shared storage only?

bensmrs (commented):
A few discardable nits, and a question out of curiosity.

internal/server/instance/drivers/driver_qemu.go
stgraber: I'm yet to find a case where they both have output this early on. I usually have the opposite issue of neither of them having anything :)

internal/server/instance/drivers/driver_common.go
stgraber: Yeah, default being 1GiB, you don't really want to try to split that across nodes.

May be cleaner to use qemudefaults.MemSize and then actively decide not to split anything below 1GiB though.

internal/server/instance/drivers/driver_qemu.go
stgraber: Basically, it's pretty rare for the early logfile to contain anything, but if it does, it would contain errors that would basically prevent the normal logger from coming online.

internal/server/instance/drivers/driver_common.go
stgraber: We allow over-provisioning in Incus, may not be advisable when talking about memory, but it's a thing that's currently allowed and that I didn't want to regress.

bensmrs: Hmmm and what happens when migrating those? Isn’t it precisely the problem you hit?

stgraber: Picking a server with enough memory or failing if not is something the placement scriptlet can handle. The actual split over the NUMA nodes when in balanced mode isn't as that's handled within Incus.

In my case I ended up manually setting 4 NUMA nodes so Incus would basically be forced what the new logic does here.

bensmrs: Ok

cmd/incusd/instance_post.go
stgraber: When dealing with a local disk, a failure on the target would kick a reverter that will typically kill off QEMU and delete the copied disk completely. In that scenario you end up with an empty destination and a dead source. Moving on would cause us to delete the source which would be a bad idea.

Shared storage is different in that there isn't any data copied (obviously) and so no reverter that could go and wipe data. Instead if we get to the scenario where the source has completed but the target is returning an error, we're better off treating the instance as now being on the target.

internal/server/instance/drivers/driver_common.go
stgraber: updated the logic to use qemudefaults

cmd/incusd/instance_post.go
bensmrs: Ah, thanks, quite a complex train of thoughts!

bensmrs (approved):

Issue #1302, opened 2024-10-11
Add support for network zone DNS NOTIFY


gbilic:
Hi all,
It would be cool to have a resolution available as soon as possible after the creation of new instances via dns server notification.
Also, currently, AXFR requests provide the zone with a serial number equal to the current timestamp, which triggers a systematic update on the slave servers.

stgraber:
We may be able to have some events trigger a notify event for affected zones, though that may be a bit tricky to figure out in some cases. 

For the serial, we don't have a way to change that as Incus generates the zone on demand and so has no way to actually version it.

gwenya:
I've looked at the code for this a bit and I think for all static IPs we should be able to easily trigger notify events, since we just need to observe changes to the config values. 

For dynamic addresses, dnsmasq has a `--dhcp-script` option which causes it to run the specified script (presumably it could be any executable?) whenever there are changes to the dhcp leases. Could we use that?

stgraber:
For monitoring dnsmasq, I think using inotify would probably be cheaper than having commands get spawned all the time.

I think one bigger issue here is cluster handling as we'd want all servers to know when something affecting the DNS zone has changed on any of the other servers, so that may make things a fair bit more complex.

Then on top of that, I think we'd want to have some kind of limits as we definitely don't want an unprivileged user (user restricted to a single project) to be able to hammer things and cause all the Incus servers in the cluster to hammer the DNS servers with NOTIFY.

gwenya:
Oh yeah, I didn't even think about clustering. I haven't looked at the code around that at all yet.
Regarding the limits, I think a first implementation could just have it disabled by default and documented with an according warning, and more generally enabling it per-zone.

gbilic:
Unrelated to clustering. Another thought: a way of doing things that would serve the same purpose would be to use dynamic dns updates instead.
(This is in fact what I do at the moment when launching a new container via an ansible script invoked by cloud-init).

PR #3896, opened 2026-08-24
Implement DNS NOTIFY

stgraber:
#1302

### DIFF
--- a/cmd/incusd/daemon.go
+++ b/cmd/incusd/daemon.go
@@ -1528,6 +1528,12 @@ func (d *Daemon) init() error {
 		logger.Info("Started DNS server")
 	}
 
+	// Watch for DHCP lease and static host changes to send DNS NOTIFY messages.
+	err = networkZone.StartLeasesWatcher(d.shutdownCtx, d.State())
+	if err != nil {
+		logger.Warn("Failed to start network zones watcher", logger.Ctx{"err": err})
+	}
+
 	// Setup the networks.
 	if !d.serverClustered || !d.db.Cluster.LocalNodeIsEvacuated() {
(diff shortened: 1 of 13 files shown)


bensmrs (approved):
What’s the watcher for, exactly? To support changes made by external parties, or is it a way to simplify the internal wiring by only having the on-disk zones be the source of truth and avoid thinking too much about message passing between all the components?

stgraber:
> What’s the watcher for, exactly? To support changes made by external parties, or is it a way to simplify the internal wiring by only having the on-disk zones be the source of truth and avoid thinking too much about message passing between all the components?

It's because Incus itself doesn't keep track of DHCP leases, it does however go and read them when generating the DNS zone.

So if you want to know when a container has acquired a DHCP lease and should get a DNS record in its zone, you've got to watch the dnsmasq lease file.

bensmrs:
Any plans for a full Go implem someday?

stgraber:
> Any plans for a full Go implem someday?

A full in-go DHCP server? We kinda looked into it a while back but it was a bit unpleasant and folks like their raw.dnsmasq. So not currently planning to switch but it may happen if we get too annoyed with dnsmasq :)

jochumdev:
incus-compose will get one, can upstream it once battle tested https://github.com/lxc/incus-compose/pull/177

stgraber:
> incus-compose will get one, can upstream it once battle tested [lxc/incus-compose#177](https://github.com/lxc/incus-compose/pull/177)

That issue is for a DNS server, we're talking DHCP here.

jochumdev:
Yes, a replacement for dnsmasq needs one for DNS as well.

PR #3900, opened 2026-08-24
Validate PEM blobs

bensmrs:


bensmrs:
Actually I could be more clever with the regex, I’ll patch that

bensmrs:
We’re very inconsistent in how we handle CA certificates, so this PR needs a bit more care to fix that.

bensmrs:
I *think* we get the best of both worlds this way. Essentially, we didn’t reject CA bundles in the past, but we only processed the first one. I’m not saying it makes sense everywhere to handle bundles, but that’s another topic. A notable limitation is that our `CheckTrustState` function and `CertInfo` struct cannot be easily ported, as they contain a CRL and it’d be quite non-trivial to handle properly. I’m not sure if there’s an impact, but if there’s a problem, we’re probably already facing it anyway.

PR #3902, opened 2026-08-25
Bump baseline to Go 1.26 (and fix/modernize)

stgraber:


bensmrs:
Poor trixie :)

bensmrs (approved):
What does the last one do?
Did the things you fixed come up in the linter following the version bump? if not, can’t you add a few linting rules for what you fixed?

gibmat:
> Poor trixie :)

Should be easy enough to undo changes if necessary for cherry-picking back security fixes, otherwise we've got Go 1.26 in trixie-backports, so no issues there. :slightly_smiling_face:

stgraber:
> What does the last one do? Did the things you fixed come up in the linter following the version bump? if not, can’t you add a few linting rules for what you fixed?

Last one is a golangci-lint fix that came up once the rest got fixed and the new baseline of 1.26 was set in gomod.

> Did the things you fixed come up in the linter following the version bump? if not, can’t you add a few linting rules for what you fixed?

A bit of a mix of both. In general we should do a `go fix` and `modernize` run every time we bump our baseline as not all of those necessarily trip the linters. We also should switch to a much more strict golangci-lint configuration, but that will be a HUGE branch.

We run incus-os with an extremely strict golangci-lint configuration, effectively enabling every single linter, then turning off some of the problematic ones and putting in a few targeted exceptions on top of that. But that was feasible given it was a new project, for Incus, we'd be looking at 24981 issues with the same config ;)

PR #3904, opened 2026-08-25
Support setting qdisc for virtual NICs

### LINKED ISSUE #1963: [FR] Support setting qdisc for virtual NICs

It's impossible to auto set the qdisc for the vNIC created by incus for instances. Unless you are able to write a program to monitor the interface creation and do such for them.

Add a config option to network devices that allow you set the desired qdisc for the device.

For VMs, more work need to be done because qemu nic typically support multi queue and the master qdisc is mq. It should add another options that allow user decide whether they want to replace the slave qdisc or the master "mq".

presztak:
Closes: #1963

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -3443,3 +3443,9 @@ communicating with the `radosgw` endpoint.
 
 This replaces `cephobject.radosgw.endpoint_cert_file`, existing values are
 converted automatically.
+
+## `device_queue_disc`
+
+Adds new `queue.discipline` and `queue.discipline.attach` configuration keys
+to `nic` devices of type `bridged`, `p2p` and `routed`, controlling the queuing
+discipline used on the host side of the NIC and where it gets attached.
--- a/internal/server/device/device_utils_network.go
+++ b/internal/server/device/device_utils_network.go
@@ -26,6 +26,7 @@ import (
 	"github.com/lxc/incus/v7/internal/server/ip"
 	"github.com/lxc/incus/v7/internal/server/network"
 	"github.com/lxc/incus/v7/internal/server/state"
+	"github.com/lxc/incus/v7/shared/api"
 	"github.com/lxc/incus/v7/shared/logger"
 	"github.com/lxc/incus/v7/shared/revert"
 	"github.com/lxc/incus/v7/shared/units"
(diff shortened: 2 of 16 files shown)


internal/server/device/device_utils_network.go
```diff
@@ -680,6 +681,99 @@ func nicValidateBurstLimits(config deviceConfig.Device, burstRate bool) error {
 	return nil
 }
 
+// nicQdiscKinds are the queuing disciplines that may be set on a NIC device.
+var nicQdiscKinds = []string{"codel", "fq", "fq_codel", "pfifo", "pfifo_fast", "sfq"}
```
stgraber: Another case of a variable being defined only to be used in exactly one spot (the validator config). Please move those directly into the validator.

internal/server/device/nic.go
```diff
@@ -55,6 +55,8 @@ func nicValidationRules(requiredFields []string, optionalFields []string, instCo
 		"ipv6.host_address":                    validate.Optional(validate.IsNetworkAddressV6),
 		"ipv4.host_table":                      validate.Optional(validate.IsUint32),
 		"ipv6.host_table":                      validate.Optional(validate.IsUint32),
+		"queue.disc":                           validate.Optional(validate.IsOneOf(nicQdiscKinds...)),
+		"queue.disc.attach":                    validate.Optional(validate.IsOneOf("root", "queue")),
```
stgraber: Since we've already expanded the `q` to `queue`, I think we should also expand the `disc` to `discipline`.

doc/api-extensions.md
```diff
@@ -3443,3 +3443,22 @@ communicating with the `radosgw` endpoint.
...
+Without a rate limit it becomes the root queuing discipline of the host side
+interface.
+
+It also adds a `queue.disc.attach` configuration key to the same devices, which
+is only valid on virtual machines. The host side interface of a virtual machine
+is a multi-queue TAP device, so by default (`queue`) the queuing discipline is
+attached to each of its transmit queues, below a `mq` root. Setting it to `root`
+instead puts a single queuing discipline at the root of the interface, which
+serializes all of its transmit queues. It cannot be set to `queue` when an
+ingress rate limit is configured, as that limit has to own the interface root.
```
stgraber: This seems overly verbose to me. API extensions should really just focus on what was added, not the why or details on how it works.

If those details are relevant to the user, they should instead be in actual documentation.

stgraber:
Looking good, just a few minor changes.

internal/server/device/device_utils_network.go
presztak: Fixed.

internal/server/device/nic.go
presztak: Done.

doc/api-extensions.md
presztak: Fixed.

internal/server/device/qdisc_test.go
```diff
@@ -0,0 +1,119 @@
+package device
```
stgraber, line 1: We usually use `tests` commits only for system tests (tests/ folder).

The unit test here should probably go in the commit that introduced the logic instead (I thought that was the case yesterday).

stgraber:
@presztak just one commit slicing issue, otherwise good to go. If you rebase, also move the API extension to the beginning of the branch please.

stgraber:
Fixed it for you

presztak:
ok, thanks!

PR #3905, opened 2026-08-25
More Secure Boot commands

bensmrs:


bensmrs:
Now that we have PEM support in config keys with our custom `SIGNATURE` anchors, PEM feels like a convenient interchange format for ESLs, at the cost of losing the owner metadata, but it’s not a huge loss IMO.

PR #3907, opened 2026-08-26
shared/cmd: Modernize table look

bensmrs:


bensmrs:
```
(before)# incus low-level secureboot list desktop kek
+------+-----------------------+--------------+--------------------------------------+
| TYPE |    OWNER GUID NAME    | FINGERPRINT  |               SUBJECT                |
+------+-----------------------+--------------+--------------------------------------+
| x509 | MICROSOFT_VENDOR_GUID | a1117f516a32 | Microsoft Corporation KEK CA 2011    |
+------+-----------------------+--------------+--------------------------------------+
| x509 | OVMF_VENDOR_GUID      | 5fb05ed84c51 | Debian UEFI Secure Boot (PK/KEK key) |
+------+-----------------------+--------------+--------------------------------------+

(after)# incus low-level secureboot list desktop kek
┌──────┬─────────────────────┬─────────────┬────────────────────────────────────┐
│ TYPE │   OWNER GUID NAME   │ FINGERPRINT │              SUBJECT               │
├──────┼─────────────────────┼─────────────┼────────────────────────────────────┤
│x509  │MICROSOFT_VENDOR_GUID│a1117f516a32 │Microsoft Corporation KEK CA 2011   │
│      ├─────────────────────┼─────────────┼────────────────────────────────────┤
│      │OVMF_VENDOR_GUID     │5fb05ed84c51 │Debian UEFI Secure Boot (PK/KEK key)│
└──────┴─────────────────────┴─────────────┴────────────────────────────────────┘

(before)# incus low-level secureboot list desktop kek --format=compact
 TYPE     OWNER GUID NAME     FINGERPRINT                 SUBJECT
 x509  MICROSOFT_VENDOR_GUID  a1117f516a32  Microsoft Corporation KEK CA 2011
 x509  OVMF_VENDOR_GUID       5fb05ed84c51  Debian UEFI Secure Boot (PK/KEK key)

(after)# incus low-level secureboot list desktop kek --format=compact
TYPE│   OWNER GUID NAME   │FINGERPRINT │              SUBJECT
x509│MICROSOFT_VENDOR_GUID│a1117f516a32│Microsoft Corporation KEK CA 2011
x509│OVMF_VENDOR_GUID     │5fb05ed84c51│Debian UEFI Secure Boot (PK/KEK key)
```

bensmrs:
Lemme patch the doc real quick

bensmrs:
Alright, should be good to go.

stgraber:
@bensmrs hmm, I think the change to the main table is fine but I'm worried about the one to the compact table as I'd expect a bunch of people are parsing it with awk or similar, expecting fields to be separate by spaces/blank characters.

Now adding a column delimiter similar to the regular table view may cause some breakage and confusion.

bensmrs:
Well the thing is, the compact format doesn’t look that compact anymore when you look at the new table. And I’ve honestly always been disappointed by its legibility; the vertical bar allows both to make it more compact and easy to read. Keeping the current compact style on the premise that people deliberately (changing to this format is a manual action) chose to parse a format that’s not made for anyone other than humans is an argument that really doesn’t resonate with me. There can already be spaces in a whole bunch of places and we have messed up with what gets fed into the, again, human-suited views of tables in the past more than I can count to claim any form of stability with those (think added columns, changed default columns depending on the context, single options that became lists displayed on several lines in a single cell, etc.).

Any way those words may make you give in? :)

stgraber:
I never use the compact view so I don't personally care all that much, but I can definitely see folks having enjoyed its combination of being human readable yet easily parsable, this branch breaks the latter.

On the default output, your examples above seem to be dropping space previously put to the left and the right of the values. To me this makes the new output look significantly worse than the previous version.

bensmrs:
> I never use the compact view so I don't personally care all that much, but I can definitely see folks having enjoyed its combination of being human readable yet easily parsable, this branch breaks the latter.

Sorry, but no, it’s not easily parsable in the general case. Tools like `cut` don’t cut it (pun intended), and awk fails in the worst possible ways:
```

$ incus list s --format=compact
     NAME        STATE               IPV4                           IPV6                  TYPE     SNAPSHOTS  LOCATION
 sxxxxxxxxxx    STOPPED                                                                 CONTAINER  2          sirius4
 sxxxxxxxx      RUNNING  10.xx.x.24 (eth0)             2001:xxx:xxxx:xxxx:x::24 (eth0)  CONTAINER  0          sirius4
 syyyyyyy       RUNNING  172.18.0.1 (br-58d9a321428c)  2001:xxx:xxxx:xxxx:x::32 (eth0)  CONTAINER  0          sirius4
                         172.17.0.1 (docker0)
                         10.xx.x.32 (eth0)

$ incus list s --format=compact,noheader | awk '{print $1}'
sxxxxxxxxxx
sxxxxxxxx
syyyyyyy
172.17.0.1
10.xx.x.32
```

> On the default output, your examples above seem to be dropping space previously put to the left and the right of the values. To me this makes the new output look significantly worse than the previous version.

So this change is not to take separately from another one that’s not reflected outside of the terminal: separators are printed with the dimmed escape sequence, thus reducing the visual load a fair bit. Spaces were good to separate the cell content from the delimiters, but I felt after comparing {dimmed/regular}, {space in header/no space in header}, {space in body/no space in body} that the combination dimmed + space in header + no space in body was the one that reduced visual bloat the most. 
image
The problem is that in the end, every single choice will be considered opinionated by one party or the other :\

stgraber:
Alright, fine, I don't care for the compact stuff ;)

However for the dimmed space thing, it looks bad when you pasted it in the Github description and I have to read a LOT of output on Github so it's going to annoy me a fair bit. I'd much prefer we return to regular spaces. You can always add a dimmed-space column option for those who really want it :)

bensmrs:
> Alright, fine, I don't care for the compact stuff ;)

Well you definitely seemed to!

> However for the dimmed space thing, it looks bad when you pasted it in the Github description and I have to read a LOT of output on Github so it's going to annoy me a fair bit. I'd much prefer we return to regular spaces. You can always add a dimmed-space column option for those who really want it :)

Ok, that one is very fair, but I’m keeping the dimmed separators.

PR #3911, opened 2026-08-26
shared/uefi: Correctly handle padded VARS files

bensmrs:
This commit allows parsing AAVMF VARS files, which are padded to 64MiB. It removes the need for magic length values by properly interpreting them directly from the firmware volume.


bensmrs:
Please run the `nvram_vm` test on aarch64, @stgraber. I can’t do it right now…

PR #3915, opened 2026-08-27
Implement interactive sessions on Windows

stgraber:


cmd/incus-agent/os_windows.go
```diff
@@ -330,8 +330,18 @@ func osStartExecCommand(ctx context.Context, cmd *exec.Cmd, pty io.ReadWriteClos
 }
 
 func osHandleExecControl(control api.InstanceExecControl, s *execWs, pty io.ReadWriteCloser, proc execProcess, l logger.Logger) {
-	// Ignore control messages.
-	return
+	if control.Command != "signal" {
+		return
+	}
+
+	// Windows has no signals, treat them all as a request to terminate the process.
```
bensmrs: It does not have signals but it does have messages, and I remember Windows having several ways to terminate processes. Maybe that’s limited to windowed applications, though, so take it with a pinch of salt.

cmd/incus-agent/exec.go
bensmrs, line 1: It’s not clear to me why you did that.

bensmrs (commented):
LGTM, just 2 quick comments.

cmd/incus-agent/exec.go
stgraber: Windows doesn't use a regular exec.Cmd but instead its own thing out of SpawnProcess.

The interface and struct here allows us to have something we can implement everywhere which provides the few functions that we call from the shared logic (exec.go).

An alternative implementation would have pushed a bunch more logic into the os-specific files.

cmd/incus-agent/os_windows.go
stgraber: Just looked into that mess, looks like it's a bit limited when a process is spawned by a service vs interactively, but we should be able to at least inject ctrl+c through the console.

stgraber:
@bensmrs added a commit which implements ctrl+c through the console and ignores all signals which shouldn't cause process termination

bensmrs (approved):
Nice, thanks for the signal handling

PR #3920, opened 2026-08-28
Various bugfixes

stgraber:


stgraber:
Bit of a mix of fixes.

- The first is just a missing Close() which was causing us trouble unmounting VMs that have had their NVRAM modified.
- The second is a few changes to limit an issue where we have VMs incorrectly get reported as dead, tripping onStop and causing their PID files to be deleted when they're in fact still alive. The commits cover a few different ways this could have happened from hitting a timeout (bumping base to 2s), to overflowing the line parser (possible with large number of disks or large stats answers).
- The third fixes an issue when doing OIDC against Rauthy (which we're rolling out in IncusOS). This IdP does not tolerate empty values in its URLs and so would fail our requests. This adds some extra logic to strip any field with an empty value from the URL. As part of that, I picked up a race condition where we were changing the global http transport/roundtripper when one goroutine would be going through re-auth. Fixing the race is a bit messy but we now pick the correct transport based on who we're talking to.

bensmrs (approved):

PR #3923, opened 2026-08-29
Implement -a for incus file push/pull

stgraber:


bensmrs (approved):
Nice, glad you managed to remove those ugly `-1` constants.
Gentle reminder to test that Windows support isn’t now broken, as that’s something that happened in the past when reworking push logic.

PR #3937, opened 2026-09-01
QEMU/QMP fixes

stgraber:
We're still seeing some occasional timeouts on functioning systems, so let's take a slightly different approach to handling heavily loaded systems.

### DIFF
--- a/internal/server/instance/drivers/qmp/commands.go
+++ b/internal/server/instance/drivers/qmp/commands.go
@@ -1513,6 +1513,8 @@ func (m *Monitor) blockJobDismiss(jobID string) error {
 // its final state or is missing. Concluded jobs are dismissed and their error, if any, is
 // returned. The returned boolean is true when the job concluded rather than turned ready.
 func (m *Monitor) blockJobWait(jobID string, waitReady bool, exitOnNotFound bool) (bool, error) {
+	queryDeadline := time.Now().Add(2 * time.Minute)
+
 	for {
 		var resp struct {
 			Return []struct {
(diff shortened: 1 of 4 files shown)


internal/server/instance/drivers/qmp/commands.go
```diff
@@ -1525,9 +1527,17 @@ func (m *Monitor) blockJobWait(jobID string, waitReady bool, exitOnNotFound bool
...
+			// Tolerate timeouts for a while as QEMU's main loop may be busy with I/O.
+			if errors.Is(err, ErrMonitorTimeout) && time.Now().Before(queryDeadline) {
+				time.Sleep(time.Second)
+				continue
+			}
+
 			return false, err
 		}
 
+		queryDeadline = time.Now().Add(operationCommandTimeout)
```
bensmrs: I’m not a huge fan of using the same timeout name for very different realities. One handles a limitation of the tool we’re interacting with, while the other encompasses whole other classes of things that can fail (in addition to the aforementioned).

internal/server/instance/drivers/qmp/qmp.go
```diff
@@ -376,28 +388,64 @@ func (qmp *qemuMachineProtocol) runWithFile(command []byte, file *os.File, id ui
...
-		qmp.replies.Delete(id)
-		return nil, fmt.Errorf("%w: %q after %s", ErrMonitorTimeout, cmd, timeout)
+
+wait:
+	for {
+		select {
+		case res = <-repCh:
+			break wait
+		case <-warnCh:
+			warnCh = nil
```
bensmrs, line 438: Does that do anything special?

bensmrs (approved):
Plenty of interesting stuff we got there!
Just 2 comments.

internal/server/instance/drivers/qmp/commands.go
stgraber: Hmm, yeah, I'll pass the value directly on this one.

internal/server/instance/drivers/qmp/qmp.go
stgraber: Makes it so we trigger on the warning timeout, then set it to nil so we don't trigger again next time we go through the loop, effectively lets us eliminate that case from the select on second entry.

bensmrs: Ah right, I didn’t notice the loop.

stgraber:
@bensmrs made the change for that timeout

bensmrs (approved):

PR #3939, opened 2026-09-02
Set PKCE challenge and verifier values

stephendotcarter:
This PR adds the required `code_challenge` and `code_verifier` values for clients that are set to require PKCE.

The `zitadel/oidc` package does not have built-in support as its potentially not providing any additional security for the device flow, more details here:
- https://github.com/zitadel/oidc/issues/784

I've tested on:
- `Keycloak 26.6.1` with `Require PKCE` both enabled and disabled and `Method` set to `S256`.
- `Rauthy v0.36.2` with `PKCE` set to `S256`.

I assume we don't want to support the `plain` method type.

Also note, `provider.IsPKCE()` only identifies if PKCE is supported, not necessarily required, so it should be safe to send the values in that case.

### DIFF
--- a/client/incus_oidc.go
+++ b/client/incus_oidc.go
@@ -3,6 +3,7 @@ package incus
 import (
 	"context"
 	"crypto/rand"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"io"
@@ -63,6 +64,7 @@ type oidcTransport struct {
 	deviceAuthorizationEndpoint string
 	tokenEndpoint               string
 	audience                    string
+	codeVerifier                string
 }
 
 // Transport returns the underlying *http.Transport used for the Incus server.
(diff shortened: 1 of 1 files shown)


client/incus_oidc.go
```diff
@@ -333,7 +339,24 @@ func (o *oidcClient) authenticate(issuer string, clientID string, audience strin
 	o.oidcTransport.deviceAuthorizationEndpoint = provider.GetDeviceAuthorizationEndpoint()
 	o.oidcTransport.tokenEndpoint = provider.OAuthConfig().Endpoint.TokenURL
 
-	resp, err := rp.DeviceAuthorization(context.TODO(), strings.Split(scopes, ","), provider, nil)
+	var authFn httphelper.FormAuthorization = nil
```
stgraber: As reported by the static analysis, you don't need to `= nil` it.

stephendotcarter: Fixed.

stgraber (approved):

PR #3949, opened 2026-09-03
Various bugfixes

stgraber:


internal/server/config/map.go
```diff
@@ -252,15 +253,14 @@ func (m *Map) set(name string, value string, initial bool) (bool, error) {
...
-	current := m.GetRaw(name)
 	if key.Type == Bool {
 		value = normalizeBool(value)
 		current = normalizeBool(current)
 	}
 
 	// Compare the new value with the current one, and return now if they
 	// are equal.
-	if value == current {
+	if isSet && value == current {
```
bensmrs, line 263: We should be safe without `isSet`; if `!isSet`, then `current == ""`, but `value` is guaranteed to be different from `""`.

bensmrs (approved):
Woops, sorry about the pipes

internal/server/config/map.go
stgraber: Actually, we're not, removing this will cause test failures.

That's because for bools we run through normalizeBool, so you'd end up with "false" instead of empty.

bensmrs: Aaaah right

PR #3964, opened 2026-09-07
Add more storage property caching

stgraber:


internal/server/storage/drivers/driver_lvm_utils.go
```diff
@@ -922,26 +923,73 @@ func (d *lvm) thinPoolVolumeUsage(volDevPath string) (uint64, uint64, error) {
...
+	}
+
+	cache := getPropertyCache("lvm", 100*time.Millisecond, 15*time.Second)
+
+	size, ok := cache.get(volDevPath, "lv_size", fill)
+	if !ok {
+		return 0, 0, false, nil
+	}
+
+	dataPercent, _ := cache.lookup(volDevPath, "data_percent")
```
presztak: Shouldn't we validate `ok` value here?

stgraber: good catch, fixed!

stgraber:
@presztak should be good to go

presztak (approved):

PR #3969, opened 2026-09-08
incusd/api: Exclude node-local config from untargeted cluster ETag

stgraber:


stgraber:
Fixes an issue where Incus behind a load-balancer can get a different server for GET and PUT when no specific target is specified.

presztak (commented):
Correct me if I'm wrong, but with this change, when the target flag isn't specified, the ETag is calculated from the global server config. So when saving a local config key without `target` flag (which seems to be allowed right now?), two different users could do so concurrently without any conflict detection, since the ETag never reflects the local config state. Am I right?

stgraber:
Yeah, you're right and there's unfortunately not much way to detect/handle that.

As the vast majority of our configuration keys are global, it should be best to prioritize their correct handling. Local configuration keys in a cluster are often set through a `--target` as otherwise you can't be sure of where they're being set. And outside of a cluster, the relax logic doesn't hit at all.

presztak (approved):

PR #3976, opened 2026-09-11
internal/server/storage/drivers: Handle dotted Ceph client IDs

fivetime:
`EnsureClientPrefix("incus.pr")` treats any dot as an entity prefix and returns `incus.pr`, which Ceph rejects as an invalid entity name. Only an existing `client.` prefix should suppress prefixing, so dotted client IDs become `client.incus.pr` while fully qualified names remain unchanged.

Verified against a disposable Ceph 19.2.3 monitor with cephx enabled:

```sh
ceph auth get-or-create client.incus.pr mon 'allow r' -o /tmp/incus-pr.keyring
ceph --name incus.pr --keyring /tmp/incus-pr.keyring fsid
ceph --name client.incus.pr --keyring /tmp/incus-pr.keyring fsid
```

The unqualified name fails with `rados_initialize ... -22`; the qualified name succeeds. A different key for the same identity cannot authenticate.

Validation: `go test -v ./...` and `make static-analysis`. The added unit test was removed following review.


internal/server/storage/drivers/utils_ceph_test.go
```diff
@@ -0,0 +1,20 @@
+package drivers
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestEnsureClientPrefix(t *testing.T) {
```
stgraber, line 9: Please drop the test

stgraber (changes requested):

fivetime:
Dropped the added test as requested. The Ceph authentication reproduction remains in the description. The full unit suite and static analysis pass.

fivetime:
The new CI run failed in `System (tip, cluster, dir, ubuntu-24.04)` at `test/suites/clustering.sh:542`: after stopping node2, `incus launch testimage bar` succeeds but `incus info bar | grep -q "Location: node1"` fails. The job uses the `dir` backend, whereas this PR changes only Ceph client-name normalization, so this appears unrelated to the patch; I have not established the scheduling failure's root cause.

Could you rerun the failed job when convenient? My rerun request returned HTTP 403 (repository admin rights required).

Failed job: https://github.com/lxc/incus/actions/runs/34797130886/job/103832301317

PR #3977, opened 2026-09-11
internal/server/util: Recognize named unconfined AppArmor profiles

fivetime:
`/proc/self/attr/current` can contain a named profile in unconfined mode, such as `runc (unconfined)`. Comparing that string with just `unconfined` makes Incus incorrectly treat its daemon as already confined and disable its own AppArmor profile handling.

Normalize labels ending in ` (unconfined)` inside the existing `AppArmorProfile` function. Keep enforced, complain and mixed labels unchanged.

Runtime validation on Linux 7.0.0-28 with AppArmor parser 4.0.1 and runc 1.3.4 covered named unconfined and enforced profiles. Kernel stacking with `aa-exec` reports `(unconfined)` for two unconfined profiles and `(mixed)` when one profile is enforced; the latter remains confined after this change.

Validation: `go test -v ./...` and `make static-analysis`. The separate helper and added unit test were removed following review.


internal/server/util/apparmor_test.go
```diff
@@ -0,0 +1,24 @@
+package util
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestAppArmorProfileName(t *testing.T) {
```
stgraber, line 9: drop the test

internal/server/util/apparmor.go
```diff
@@ -9,8 +9,17 @@ import (
...
 	contents, err := os.ReadFile("/proc/self/attr/current")
 	if err == nil {
-		return strings.TrimSpace(string(contents))
+		return appArmorProfileName(string(contents))
 	}
 
 	return ""
 }
+
+func appArmorProfileName(profile string) string {
```
stgraber, line 18: move into the existing function

stgraber (approved):

stgraber (changes requested):
see inline comments

fivetime:
Moved the label handling into the existing `AppArmorProfile` function and dropped the added test. The full unit suite and static analysis pass.

fivetime:
The `System (stable, cluster, dir, ubuntu-24.04)` job failed at `test/suites/clustering.sh:542`: `incus launch testimage bar` succeeds after node2 is stopped, but the expected `Location: node1` assertion fails. This is the same assertion failure seen in the independent Ceph-only PR #3976 (tip/cluster/dir); it appears unrelated to this AppArmor parsing change, though the root cause is not yet established.

Could you rerun the failed job when convenient? The rerun API requires upstream repository admin rights, which my account does not have.

Failed job: https://github.com/lxc/incus/actions/runs/34797365780/job/103832961518

PR #3978, opened 2026-09-11
internal/server/instance/drivers: Publish migration errors before completion

fivetime:
The filesystem/state workers close their completion channels in a defer, before `errgroup` sees their returned error and cancels its context. A waiter can therefore observe completion and an uncancelled context after a failed transfer, start CRIU restore, or send a successful migration response.

Publish each transfer result before closing its completion channel. Check that result both before restore and before the final response; retain the context check for control-monitor errors. The regression test holds the worker between publishing completion and returning to `errgroup`, with two concurrent waiters, so the failure window is deterministic and needs no timing assumptions.

Validation: focused tests with the race detector, `go test -v ./...`, and `make static-analysis`.

### DIFF
--- a/internal/server/instance/drivers/driver_lxc.go
+++ b/internal/server/instance/drivers/driver_lxc.go
@@ -7001,9 +7001,9 @@ func (d *lxc) MigrateReceive(args instance.MigrateReceiveArgs) error {
 	}()
 
 	// Start filesystem transfer routine and initialize a channel that is closed when the routine finishes.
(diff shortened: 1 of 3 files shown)


stgraber (changes requested):
This is overly complicated for the needed change. I've sent a simpler version which handles both drivers here: https://github.com/lxc/incus/pull/3980

fivetime:
Thanks for simplifying this and covering both drivers in #3980!

PR #3979, opened 2026-09-11
internal/server/instance/drivers: Unmount residual Unix devices during cleanup

fivetime:
Failed restore cleanup can reach `removeUnixDevices` with device bind mounts still present. Removing their mountpoints returns `EBUSY`, leaving the entries behind.

Unmount them normally before removing the entries. Ignore only the expected not-mounted/already-gone errors; collect other unmount/removal errors and return them after attempting the remaining devices. An unmount failure keeps its mountpoint intact. This uses no lazy detach.

A separate private mount-namespace check exercised failed-restore cleanup, repeated cleanup, a real `EPERM` failure after dropping `CAP_SYS_ADMIN`, and a real `EBUSY` failure while the current directory held the bind mount busy. The busy mount and its contents remained accessible; cleanup succeeded after releasing the reference. The previous `MNT_DETACH` implementation failed this check by returning success for the busy mount.

Validation: the separate runtime check with the race detector, `go test -v ./...`, and `make static-analysis`. The added test file was removed from this PR following review. The runtime check injects cleanup state directly and does not claim a full CRIU migration test.


internal/server/instance/drivers/driver_lxc_devices_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sys/unix"
+
+	"github.com/lxc/incus/v7/shared/api"
+	"github.com/lxc/incus/v7/shared/logger"
+)
+
+func TestRemoveUnixDevicesAfterFailedRestore(t *testing.T) {
```
stgraber, line 19: drop the test

internal/server/instance/drivers/driver_lxc.go
```diff
@@ -8923,22 +8923,30 @@ func (d *lxc) removeUnixDevices() error {
...
 			continue
 		}
 
-		// Remove the entry
+		// Detach bind mounts left behind by a failed restore before removing their mountpoints.
 		devicePath := filepath.Join(d.DevicesPath(), f.Name())
-		err := os.Remove(devicePath)
-		if err != nil {
-			d.logger.Error("Failed removing unix device", logger.Ctx{"err": err, "path": devicePath})
+		err := unix.Unmount(devicePath, unix.MNT_DETACH)
```
stgraber, line 8937: We should never use detach as that can then hide a mounted filesystem leading to data corruption or system crashes when we then go to the storage layer to delete it.

stgraber (changes requested):

fivetime:
Changed to a normal unmount (flags 0), retaining the mountpoint and returning the error on `EBUSY` or `EPERM`. Dropped the added test from this PR.

Verified both failures in a private mount namespace, including that cleanup succeeds after releasing the busy reference. The previous lazy-detach version incorrectly returned success for that busy mount. The full unit suite and static analysis pass.

PR #3994, opened 2026-09-14
Fix handling of volatile keys during snapshot restore

stgraber:


test/suites/projects.sh
```diff
@@ -1028,6 +1028,28 @@ test_projects_restrictions() {
...
+    # Snapshots can be restored with restricted.containers.lowlevel set to
+    # 'block' even though their server-generated volatile keys differ.
+    incus project set p1 restricted.containers.lowlevel=block
+    incus project set p1 restricted.snapshots=allow
+    incus init testimage c1
+    incus snapshot create c1 snap0
+    incus start c1
+    incus stop c1 --force
+    incus snapshot restore c1 snap0
+    incus snapshot restore c1 snap0
```
presztak, line 1040: Do we need call that twice?

stgraber: It's to make sure that the snapshot restore logic which modifies the volatile keys doesn't prevent a later snapshot restore. It shouldn't really be an issue but figured we may as well be safe here.

presztak (approved):

PR #4017, opened 2026-09-17
Prevent getting blocked during daemon shutdown

stgraber:


cmd/incusd/instances.go
```diff
@@ -570,6 +573,33 @@ func instancesShutdown(instances []instance.Instance) {
 					}
 				}()
 
+				// A shutdown forced from the start still gets the regular timeouts.
+				forceCh := ctx.Done()
+				if ctx.Err() != nil {
```
presztak: Not sure about this. Shouldn't we check `ctx.Err() != nil` before loop and have the same behaviour for all instances. With this approach after forcing shutdown in the middle of this loop we will set `forceCh` to `nil` for some instances.

stgraber: Good catch, updated!

presztak (approved):

PR #4020, opened 2026-09-18
incusd/network/ovn: Match IPv6 load balancer health monitors

stgraber:


internal/server/network/ovn/ovn_nb_actions.go
```diff
@@ -4037,8 +4037,14 @@ func (o *NB) GetLoadBalancersByStatusUpdate(ctx context.Context, mon ovnSB.Servi
...
 		}
 
+		// IPv6 addresses are stored bracketed in the mappings but not in the service monitor.
 		for k, v := range lb.IPPortMappings {
-			if k == mon.IP && v == fmt.Sprintf("%s:%s", mon.LogicalPort, mon.SrcIP) {
+			if strings.Trim(k, "[]") != mon.IP {
+				continue
+			}
+
+			port, srcIP, ok := strings.Cut(v, ":")
```
presztak, line 4046: Just double checking but port is before IP in that case?

presztak (approved):

internal/server/network/ovn/ovn_nb_actions.go
stgraber: @presztak in this case "port" refers to an OVN logical switch port, but yeah, is shows up first

PR #4033, opened 2026-09-20
Add support for creating multiple internal subnets

### LINKED ISSUE #1970: Allow to create multiple internal subnets

Incus allows only one internal subnet to be created:
```
incus network create net1 --type=ovn network=UPLINK ipv4.address=192.168.0.1/24 ipv6.address=2001:db8:1::1/64 ipv4.nat=true ipv6.nat=false
incus launch images:ubuntu/22.04 vm1 --network net1
```
On the OVN side it will be translate to: 
```
router beea0e3f-1782-4f54-be2e-e3537832c922 (incus-net86-lr)
    port incus-net86-lr-lrp-ext
        mac: "00:16:3e:6a:39:c7"
        networks: ["100.64.1.10/24", "2001:db8::2/64"]
    port incus-net86-lr-lrp-int
        mac: "00:16:3e:6a:39:c7"
        networks: ["192.168.0.1/24", "2001:db8:1::1/64"]
    nat 5b4ddf5e-6b49-49b6-8f77-a2b1dab79345
        external ip: "100.64.1.10"
        logical ip: "192.168.0.0/24"
        type: "snat"
```
The Logical router port incus-net86-lr-lrp-int is attached to a Logical Switch (internal subnet).

We could have more internal subnets connected to the same Logical router.
An user would like to have multiple internal subnets for different puporses on the same Logical Router.

presztak:
Closes: #1970


presztak:
The first two commits allow us to specify the `parent` option, which is overall node specific but shouldn't be for the OVN network type. We could come up with a different name to avoid these changes, but I guess `parent` makes sense.

internal/server/network/driver_ovn.go
```diff
@@ -1332,14 +1351,112 @@ func (n *ovn) getNetworkPrefix() string {
...
+
+		children = append(children, ovnChild)
+	}
+
+	return children, nil
+}
+
+// routerFamily returns the networks sharing our logical router, starting with its owner.
+func (n *ovn) routerFamily() ([]*ovn, error) {
+	owner := n
```
stgraber, line 1543: I'd pick a different name for that as `family` in networking usually refers to IPv4/IPv6. That got me confused for a bit here before I realized that family hears refers to parent+children.

maybe relatedNetworks or something along those lines?

doc/reference/network_ovn.md
```diff
@@ -51,6 +51,32 @@ The extended format is `<interfaceName>/<parentInterfaceName>/<vlanId>`.
...
 ```
 
+(network-ovn-child)=
+## Child networks
+
+An OVN network can be created with a `parent` pointing at another OVN network in the same project.
+Instead of creating a logical router of its own, the child attaches its own logical switch and subnet to the logical router of its parent:
+
+    incus network create net1 --type=ovn network=UPLINK ipv4.address=192.168.0.1/24
+    incus network create net2 --type=ovn parent=net1 ipv4.address=192.168.1.1/24
```
stgraber: We should use 192.0.2.0/24 and 198.51.100.0/24 as they're meant for documentation use.

doc/reference/network_ovn.md
```diff
@@ -51,6 +51,32 @@ The extended format is `<interfaceName>/<parentInterfaceName>/<vlanId>`.
...
+    incus launch images:ubuntu/22.04 vm1 --network net2
```
stgraber: use debian/13, we've tried to drop Ubuntu from examples/docs

stgraber:
@presztak the new tests don't appear to be run, they'd need to be added to the test matrix

internal/server/network/driver_ovn.go
presztak: Fixed.

doc/reference/network_ovn.md
presztak: Fixed.

doc/reference/network_ovn.md
presztak: Fixed.

presztak:
@stgraber, tests added to the test matrix.

PR #4042, opened 2026-09-21
Implement cluster metrics

stgraber:


cmd/incusd/api_metrics.go
```diff
@@ -359,6 +367,105 @@ func getFilteredMetrics(s *state.State, r *http.Request, compress bool, metricSe
 	return response.SyncResponsePlain(true, compress, metricSet.String())
 }
 
+// clusterMemberMetrics adds info, status and role metrics for every cluster member.
+func clusterMemberMetrics(ctx context.Context, s *state.State, out *metrics.MetricSet) error {
+	leaderAddress, err := s.Cluster.LeaderAddress()
+	if err != nil {
+		return err
+	}
```
presztak, line 376: I wouldn't return an error here but would just assign an empty string to `leaderAddress`. Then, even if this fails, we can still answer the `incus_cluster_member_status{status="offline"} == 1` query. WDYT?

presztak (approved):

PR #4043, opened 2026-09-21
Have CLI parser fail when object name is passed in remote name only context

stgraber:


bensmrs:
Not a huge fan of the solution; I’ll try to come up with something that fixes it another way, while handling all the edge cases mentioned internally

PR #4049, opened 2026-09-21
Add support for security.tags and exposure to OpenFGA

stgraber:
This adds a new comma separate `security.tags` key on the instances and when using OpenFGA, maps it to a new `security_tag` object. This by itself doesn't do anything but what it does is expose those tags within the OpenFGA model, making it possible for external authorization provisioning tools to look at instances and know the tags that are applied to them.

Such tools typically exclusively work against the OpenFGA data with no direct access to Incus. Up until this point, all they could do was look at instance and project names and inject tuples based on that. With this extra option, they can now also consider security tags applied to the instances.

It's up to the external tool whether to trust those tags and whether to trust them globally or just for a particular project, Incus just exposes the value.


internal/server/auth/driver_openfga.go
```diff
@@ -583,36 +583,249 @@ func (f *FGA) AddInstance(ctx context.Context, projectName string, instanceName
 
 // DeleteInstance deletes an instance from the authorizer.
 func (f *FGA) DeleteInstance(ctx context.Context, projectName string, instanceName string) error {
+	// If offline, skip updating as a full sync will happen after connection.
+	f.onlineMu.Lock()
+	defer f.onlineMu.Unlock()
+	if !f.online {
+		return nil
+	}
```
presztak: I wonder if we should only acquire a lock when reading `f.online`. This would avoid blocking the whole logic.

stgraber: Added a new isOnline helper and made use of it throughout.

stgraber:
@presztak that should take care of the review feedback

presztak (approved):

PR #4053, opened 2026-09-22
incus/server/storage/driver/ceph: Flatten image in background during instance creation

### LINKED ISSUE #3500: Ceph: Allow flattening image after instance has been created

Incus in combination with ceph allows for the convenient option to use cloning of images rather than copying. This offers the advantage, that images don't have to be unpacked again for new instances, if a smaller or eual sized image already exists in ceph, making instance deployment very fast.
The downside of this is, that ceph is unable to report the proper size used by the cloned image, as some blocks of the image belong to the parent image it was cloned of. This effect worsens if the cloned image has snapshots.

Unfortunately getting the disk utilization via the agent is not an option for us, as it's potentially unreliable because the agent can be deactivated from within the instance.

A bool configuration key in the device section of the instance like `flatten` or `postFlatten` that causes the link between the cloned image and its parent to be severed after the clone in ceph has been created. In ceph this can be achieved via the `rbd flatten` command
This would offer the time advantage of instance deployment by utilizing cloning, and at the same time allow ceph to report correct disk utilization of the image.

If this is an acceptable feature to be added, I'd be happy to implement it :)

DarkressX:
Closes #3500 

EDIT: Maybe some explanation is required.
This PR ensures we _always_ first create a clone.
If the `ceph.rbd.clone_copy` flag in the ceph pool is **false**, we start a background go-routine that triggers image flattening.
Errors during that flattening process are only logged, not trigger a failure of the instance creation.


stgraber:
@DarkressX generally looks fine but I'm worried about one particular path.

If we're copying something that's not an image, we're creating a snapshot and protecting it, then we clone and finally we flatten. But I don't see us ever unprotecting and deleting that snapshot following the flattening, so I'm assuming we're effectively leaking a snapshot every time this happens.

DarkressX:
@stgraber  You are correct, although it's not worse than the current bahaviour when the flag is set to false.
Would it be ok if I extend the periodic `pruneLeftoverImages` to take care of stale Ceph images as well? We just have to be careful that multiple cluster members are not racing for a cleanup at the same time.

EDIT: Disregard. It seems like this is already been taken care of whenever the `autoUpdateImagesTask` runs. Including checking for dependant clones, unprotecting, deleteing the readonly snapshot and the image itself.

stgraber:
@DarkressX added some logic to automatically get rid of the snapshots

PR #4054, opened 2026-09-22
incusd/instance/qemu: Abort snapshot transfer on failed live migration

stgraber:
When the migration fails after the snapshot mirror job has started, the job's filter node and the NBD target device are left in place. This makes the block-commit merging the snapshot back fail with "Need a root block node", leaving the overlay attached and any later migration failing with a duplicate node name.


presztak (approved):

dennisoderwald:
@stgraber When can we expect the fix to be included in the LTS release?

stgraber:
Will be in 7.0.2 which we're expecting to tag in the next couple of weeks.

